PR #71, opened 2026-01-09
Dev paul

OPS-PIvers:
Added live mode to widgets (phase one), read only student app page (phase 2), and now I need to implement phase 3 that allows write mode for students on app that logically would make sense for students to participate on. Not all apps make sense to be read and write, so we need to determine which ones are or not as well


OPS-PIvers:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/OPS-PIvers/SPART_Board/pull/71#pullrequestreview-3645736939)

PR #1225, opened 2026-04-07
Fix: Derive background auth state during render to avoid useEffect anti-patterns

OPS-PIvers:
Found an instance of a hacky `useEffect` containing a `setTimeout(..., 0)` used to update local state without triggering synchronous strict-mode issues. 

Refactored `hooks/useBackgrounds.ts` to implement the "Deriving state during render" best practice from the React documentation ("You Might Not Need an Effect"). It now uses `useState` to track the previous `user` and updates state conditionally during the main render block, while effectively returning early from the actual data-fetching `useEffect` if `!user`.

This ensures we do not trigger lint warnings around calling setters directly in effects while adhering tightly to modern React patterns. Added tests locally (lint, format, type-check) to ensure no regressions.

---
*PR created automatically by Jules for task [12073582119890295114](https://jules.google.com/task/12073582119890295114) started by @OPS-PIvers*

### DIFF
--- a/hooks/useBackgrounds.ts
+++ b/hooks/useBackgrounds.ts
@@ -27,10 +27,8 @@ export const useBackgrounds = () => {
   const [prevUser, setPrevUser] = useState(user);
   if (user !== prevUser) {
     setPrevUser(user);
-    if (!user) {
-      setManagedBackgrounds([]);
-      setLoading(false);
-    }
+    setManagedBackgrounds([]);
+    setLoading(!!user);
   }
 
   // Refs to prevent race conditions when both queries update simultaneously
--- a/hooks/useBackgrounds.ts
+++ b/hooks/useBackgrounds.ts
@@ -22,7 +22,16 @@ export const useBackgrounds = () => {
   const [managedBackgrounds, setManagedBackgrounds] = useState<
     BackgroundPreset[]
   >([]);
-  const [loading, setLoading] = useState(true);
+  const [loading, setLoading] = useState(!!user);
+
+  const [prevUser, setPrevUser] = useState(user);
+  if (user !== prevUser) {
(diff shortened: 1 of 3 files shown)


OPS-PIvers:
@jules

[hooks/useBackgrounds.ts](https://github.com/OPS-PIvers/SPART_Board/pull/1225/files/4b59f647b4c6291c18e4bec8b2f637ffc0b03846#diff-98d646ca933a3cab43c665f3fcb35bed05d7fafbf8085195f4803721397609bd)
Comment on lines +28 to +34
  if (user !== prevUser) {
    setPrevUser(user);
    if (!user) {
      setManagedBackgrounds([]);
      setLoading(false);
    }
  }
@[gemini-code-assist](https://github.com/apps/gemini-code-assist)
gemini-code-assist bot
[22 minutes ago](https://github.com/OPS-PIvers/SPART_Board/pull/1225#discussion_r3047986294)
 high

The current implementation of the "Adjusting state while rendering" pattern has a logic bug regarding the loading state and potential stale data.

Loading State Bug: When a user logs in (transition from null to User), user !== prevUser is true, but the if (!user) block is skipped. Consequently, loading remains false (its value from the logged-out state) while the useEffect begins fetching data. This prevents the UI from showing a loading indicator during the initial fetch.
Stale Data: When switching between different users (e.g., UserA to UserB), managedBackgrounds is not cleared in the render block. This causes the UI to display the previous user's backgrounds until the new Firestore subscription returns data.
Updating the logic to reset both loading and managedBackgrounds whenever the user identity changes ensures a consistent and responsive UX.

  if (user !== prevUser) {
    setPrevUser(user);
    setManagedBackgrounds([]);
    setLoading(!!user);
  }

PR #1297, opened 2026-04-13
⚡ Optimize removeWidgets group dissolution logic

OPS-PIvers:
💡 **What:** The optimization implemented is refactoring the `removeWidgets` function in `DashboardContext.tsx`. Specifically, it replaces a nested loop that filtered the full widgets array multiple times with a single-pass approach using a `Map` to track group counts.

🎯 **Why:** Filtering the full widgets array for every affected group ID was highly inefficient ($O(N \times M)$). This caused noticeable lag when removing multiple widgets that were part of many different groups.

📊 **Measured Improvement:**
- **Baseline:** ~1007ms (average 100.7ms per call) for 10,000 widgets and 500 groups.
- **Optimized:** ~22ms (average 2.2ms per call) for the same data.
- **Speedup:** Approximately **45x faster**.

The optimization ensures the dashboard remains responsive even as the number of widgets and groups grows.

---
*PR created automatically by Jules for task [13611942930188240706](https://jules.google.com/task/13611942930188240706) started by @OPS-PIvers*

### DIFF
--- a/context/DashboardContext.tsx
+++ b/context/DashboardContext.tsx
@@ -2553,23 +2553,42 @@ export const DashboardProvider: React.FC<{ children: React.ReactNode }> = ({
           d.widgets.forEach((w) => {
             if (idSet.has(w.id) && w.groupId) affectedGroupIds.add(w.groupId);
           });
-          let widgets = d.widgets.filter((w) => !idSet.has(w.id));
-          // Auto-dissolve groups with <=1 remaining member
+          const widgets = d.widgets.filter((w) => !idSet.has(w.id));
+
+          if (affectedGroupIds.size === 0) return { ...d, widgets };
+
+          // Count remaining members for each affected group
+          const groupCounts = new Map<string, number>();
+          widgets.forEach((w) => {
+            if (w.groupId && affectedGroupIds.has(w.groupId)) {
+              groupCounts.set(w.groupId, (groupCounts.get(w.groupId) ?? 0) + 1);
+            }
+          });
+
+          // Identify groups with <= 1 remaining member
+          const groupsToDissolve = new Set<string>();
           for (const gid of affectedGroupIds) {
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review
**Date:** 2026-04-14

### Summary
Optimizes `removeWidgets` in `context/DashboardContext.tsx` — replaces nested group dissolution with an O(N+M) approach using a `Map` of group counts and a single-pass dissolution check. Good structural refactor.

### Code Quality
- The Map-keyed counting pattern is the right fix for what was likely the hottest path in bulk widget deletion.
- Dissolution logic (group removed when count drops to zero) should be covered by a test — the old nested logic was easy to reason about inline; the new code is denser and benefits from a regression test.

### Widget Standards
N/A — this is context/state logic, not widget rendering.

### Admin Config Standards
N/A.

### Firestore & Security
- Verify that the resulting Firestore write still batches correctly after the in-memory widget list is updated. If `saveCurrentDashboard()` or equivalent is called synchronously after, a single batch write is expected.
- No rules implications.

### Regression Risk
**Medium.** This touches core dashboard state. Test matrix to cover:
- Remove one widget, no group (baseline).
- Remove all widgets of a group in one call → group should dissolve.
- Remove a subset of a group's widgets → group should remain with reduced members.
- Remove widgets that belong to multiple overlapping groups (if that's supported).
- Undo/redo interaction, if the context exposes it.

### Missing Tests
- No test file shown in this PR for the new removal logic. Strongly recommend a unit test in `tests/hooks/` or `tests/components/` before merge.

### Overall
**Merge readiness: Approve with a tests follow-up.** Logic looks correct, but given this is a hot core-state path, a dedicated regression test suite would de-risk the change meaningfully.

---
_Automated review by Claude Code._

---

OPS-PIvers:
@jules

given this is a hot core-state path, a dedicated regression test suite would de-risk the change meaningfully.

PR #1317, opened 2026-04-15
feat(quiz): assignment archive with pause/resume and PLC sharing

OPS-PIvers:
## Summary

Replaces the single-session-per-teacher quiz model with a per-assignment archive. Multiple classes can be active concurrently, each with its own join code, and responses are preserved until the teacher explicitly deletes an assignment. Adds PLC sharing of full assignments (settings + sheet URL) and fixes the paste flow so `/share/quiz` and `/share/assignment` links open the Quiz widget without a page reload.

### Data model

- `QuizAssignment` is a new record under `/users/{uid}/quiz_assignments/{id}` that owns session settings (className, PLC fields, session toggles) and a back-reference to the quiz in the Library.
- `QuizSession` docs at `/quiz_sessions/{sessionId}` are now keyed by the assignment's UUID (1:1) rather than teacher UID. `teacherUid` moved to a field; student join-by-code is unaffected.
- New `/shared_assignments/{shareId}` collection mirrors `/shared_quizzes/*` for PLC imports.
- Assignment status: `active | paused | inactive`. Paused keeps the session alive but halts the timer and shows a full-screen placeholder in the student app. Inactive ends it while preserving responses.

### UI

- Quiz Manager gains a **Library / Archive** tab switcher. Per-quiz END/RESUME controls on library rows are gone; assignment lifecycle lives entirely in the Archive.
- `QuizAssignmentArchive` lists rows with color-coded status pill and per-row actions: Copy URL, Monitor, Results, Edit Settings, Share, Pause/Resume, Deactivate, Delete.
- `QuizAssignmentSettingsModal` edits settings only (quiz content stays tied to the Library). `sessionMode` is locked while active/paused.
- `QuizLiveMonitor` adds Pause/Resume; End is now "Make Inactive" with a confirm dialog explaining preservation.
- Student app renders `QuizPausedPlaceholder` on paused sessions and rejects joins on ended/inactive.
- Quiz Settings' "Force End Active Session" button is replaced with a "View Assignment Archive" link.

### Paste flow fix

- `Dock.tsx` no longer does `window.location.href` on share-link paste; it sets `pendingQuizShareId` / `pendingAssignmentShareId` in `DashboardContext`.
- `DashboardView.tsx` imports the shared quiz or assignment, then opens (or focuses) a Quiz widget and switches it to the matching tab. External `/share/*` URLs still work via the mount-time URL parser.
- `smartPaste.ts` gains `tryParseAssignmentImport`.

### Firestore rules

- `/quiz_sessions/{sessionId}` rules moved to sessionId keying; ownership checks use `resource.data.teacherUid`. Response writes add one `get()` lookup — ~50ms per answer submission, fine at class scale.
- New owner-only rules for `/users/{uid}/quiz_assignments/*` and author-create rules for `/shared_assignments/*`.

### Migration

No script. Live sessions are ephemeral; orphan `/quiz_sessions/{teacherUid}` docs are left in place but become unreachable. Release note should warn teachers to re-assign anything mid-session.

## Test plan

- [ ] Teacher A creates Quiz X, clicks Assign with className "Period 1" → Archive shows one Active row with a code.
- [ ] Teacher A clicks Assign again with "Period 2" → Archive shows two Active rows with different codes; students can join either.
- [ ] Teacher A pauses Period 1 → student on Period 1 sees paused placeholder; resume brings them back.
- [ ] Teacher A makes Period 1 inactive → student sees session-ended; Results from Archive still load.
- [ ] Teacher A deletes Period 1 → Archive row gone, responses purged.
- [ ] Teacher A shares Quiz X; Teacher B pastes link → Quiz widget opens to Library tab with Quiz X present. No reload.
- [ ] Teacher A shares Period 2 assignment (with PLC sheet URL); Teacher B pastes link → Quiz widget opens to Archive tab, row is Paused, settings show preserved sheet URL. Resume → copy code → student joins.
- [ ] Rules: student responses accepted on active/paused, rejected on inactive; teacher can only access their own assignments.
- [ ] `pnpm run validate` green; 1156 tests pass.

https://claude.ai/code/session_01YWopTYRKmMGRWHUdkT8GfU

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -322,6 +322,7 @@ The project uses `@/` as an alias for the **root directory** (not `src/`):
 /users/{userId}/notebooks/{notebookId}      # Notes
 /users/{userId}/pdfs/{pdfId}                # PDFs
 /users/{userId}/userProfile/{profileId}     # User profile / building selection
+/users/{userId}/quiz_assignments/{id}       # Teacher's quiz assignment archive
 
 # Admin collections (admin read/write, authenticated read)
 /admins/{email}                             # Admin users (existence = admin)
@@ -334,6 +335,7 @@ The project uses `@/` as an alias for the **root directory** (not `src/`):
 
 # Shared/global collections
 /shared_boards/{shareId}                    # Shared dashboards
+/shared_assignments/{shareId}               # Shared quiz assignments (PLC sharing)
 /global_weather/{document}                  # Cached weather data
 /global_music_stations/{document}           # Shared music stations
 /global_pdfs/{pdfId}                        # Shared PDF library
@@ -344,7 +346,7 @@ The project uses `@/` as an alias for the **root directory** (not `src/`):
 
 # Live session collections
 /sessions/{userId}/students/{studentId}     # Live session students
-/quiz_sessions/{teacherUid}/responses/...   # Quiz responses
+/quiz_sessions/{sessionId}/responses/...    # Quiz responses (one session doc per assignment)
 /video_activity_sessions/{sessionId}/...    # Video activity responses
 /guided_learning_sessions/{sessionId}/...   # GL session responses
 /mini_app_sessions/{sessionId}              # Mini-app assignment sessions
--- a/components/layout/DashboardView.tsx
+++ b/components/layout/DashboardView.tsx
@@ -12,6 +12,7 @@ import {
 import { useAuth } from '@/context/useAuth';
 import { useLiveSession } from '@/hooks/useLiveSession';
 import { useQuiz } from '@/hooks/useQuiz';
+import { useQuizAssignments } from '@/hooks/useQuizAssignments';
 import { useStorage, MAX_PDF_SIZE_BYTES } from '@/hooks/useStorage';
 import { Sidebar } from './sidebar/Sidebar';
 import { Dock } from './Dock';
@@ -144,6 +145,8 @@ export const DashboardView: React.FC = () => {
     setZoom,
     pendingQuizShareId,
     clearPendingQuizShare,
+    pendingAssignmentShareId,
+    clearPendingAssignmentShare,
     // Widget grouping
     groupWidgets,
     groupBuildMode,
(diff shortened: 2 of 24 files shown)


OPS-PIvers:
Addressed review feedback in 052ffdb:

**Copilot**
- `useQuizAssignments.ts` (`setStatus`): clear `autoProgressAt` on pause and deactivate so an in-flight auto-advance timer can't fire after the teacher pauses.
- `useQuizAssignments.ts` (`resumeAssignment`): resume never-started sessions (no `startedAt` / `currentQuestionIndex < 0`) back to `'waiting'` instead of `'active'`, so imported paused PLC assignments send students to the waiting room rather than a spinner.
- `useQuizSession.ts` (join lookup): filter by joinable status client-side and prefer the newest doc, so a stale ended session with a recycled code can't shadow a live one.
- `Widget.tsx` (`onDelete`): block library-quiz deletion while active or paused assignments reference it; warn on archived references so teachers don't silently orphan monitor/results views.

**Gemini**
- `Widget.tsx` (share path): switched to the `loadQuiz` wrapper so share failures get the same null-guard, error toast and error-state handling as the monitor/results paths.
- `QuizAssignmentSettingsModal.tsx`: dropped `|| undefined` so empty strings actually clear `className` / `teacherName` / `periodName` / `plcSheetUrl` in Firestore. Readers already use `a.className && (...)` so empty-string reads behave the same as the previous absent-field case.
- `QuizAssignmentArchive.tsx`: switched the four inline confirm-dialog buttons to cqmin-based scaling to match the rest of the archive row.

**Skipped (false positives)** — Gemini flagged `setLoadedQuizData(data)` as missing at `Widget.tsx:656`, `:759`, `:785`. In all three cases the handler calls `await loadQuiz(meta)` (not the raw `loadQuizData`), and the `loadQuiz` wrapper at line 97 already calls `setLoadedQuizData(data)` internally before returning. State is set in the same event handler as `updateWidget`, so React batches both updates and the monitor/results view sees `loadedQuizData` populated on its first render.

Validation: `pnpm run validate` clean, 1156/1156 tests pass.

---

OPS-PIvers:
Addressed latest round in 587cef1:

- **`firestore.rules`** (Copilot, line 55): hardened `/users/{uid}/quiz_assignments/{id}` writes so `teacherUid` must match `userId` and `id` must match `assignmentId` on create, and both are immutable on update. This closes the spoofing gap on fields that ownership checks elsewhere in the rules (session docs, shared_assignments) rely on.

**Skipped (intentional pattern):**

- **`useQuizSession.ts:165`** and **`useQuizAssignments.ts:147`** (Copilot): these implement the *"adjusting state while rendering"* pattern from [React's `useState` docs](https://react.dev/reference/react/useState#storing-information-from-previous-renders), which this codebase explicitly prefers over `useEffect` for resetting derived state on prop changes. From [`CLAUDE.md`](https://github.com/OPS-PIvers/SpartBoard/blob/main/CLAUDE.md):

  > `useEffect` is an escape hatch, not a default. … Reset state on prop change using a `key` prop or the *"adjusting state while rendering"* pattern (store the previous prop value in state, compare during render, call the setter immediately if they differ).

  The setter is called during render but *before* any return, React detects the in-progress render and restarts with the new state without committing the stale one — no warning, no loop, no extra commit. Moving this to `useEffect` would introduce the exact extra render pass the React docs recommend avoiding.

---

PR #1354, opened 2026-04-19
Refactor `useEffect` prop synchronization in `SidebarBackgrounds`

OPS-PIvers:
Removed anti-pattern use of `useEffect` for syncing props and derived context state into local state in `SidebarBackgrounds`. Replaced with the "adjusting state while rendering" pattern using `prevProps` inside `useState` as per React best practices.

### DIFF
--- a/components/layout/sidebar/SidebarBackgrounds.tsx
+++ b/components/layout/sidebar/SidebarBackgrounds.tsx
@@ -91,6 +91,11 @@ export const SidebarBackgrounds: React.FC<SidebarBackgroundsProps> = ({
 
   // Category detail panel — null = overview, string = detail view for that category
   const [activeCategory, setActiveCategory] = useState<string | null>(null);
+  const [prevIsVisible, setPrevIsVisible] = useState(isVisible);
+  if (isVisible !== prevIsVisible) {
+    setPrevIsVisible(isVisible);
+    if (!isVisible) setActiveCategory(null);
+  }
 
   // Custom color picker state
   const [customColor, setCustomColor] = useState('#3b82f6');
@@ -185,11 +190,6 @@ export const SidebarBackgrounds: React.FC<SidebarBackgroundsProps> = ({
     if (tab !== 'media') setSearchQuery('');
   };
 
-  // Reset detail panel when sidebar is hidden
-  useEffect(() => {
-    if (!isVisible) setActiveCategory(null);
-  }, [isVisible]);
-
   // Fetch past uploads from Google Drive when the "My Uploads" tab is opened
   useEffect(() => {
     if (designTab !== 'my-uploads' || !isInitialized || hasFetchedDrive) return;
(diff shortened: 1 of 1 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-04-20

### Summary
Replaces two `useEffect` prop/derived-state synchronizers in `SidebarBackgrounds.tsx` with the "adjusting state while rendering" pattern (prev-value stored in state, compared during render, setter called immediately on change). Directly implements the guidance in CLAUDE.md's "useEffect is an escape hatch" gotcha.

### Code Quality
- ✓ Both conversions are correct. The `isVisible` → `activeCategory` reset now happens synchronously in the render phase rather than after commit — no user-visible flash.
- ✓ The `activeCustomValue` conversion preserves the original early-return semantics via an explicit `else` branch (hex skips the gradient parse, gradient parse skips the hex setter).
- ✓ `useEffect` remains used elsewhere in the file (Google Drive fetch, other tabs) so the import is not orphaned.

### Widget Standards
- ✓ Not applicable.

### Admin Config Standards
- ✓ Not applicable.

### Firestore & Security
- ✓ No changes.

### Regression Risk
- ✓ Low. Behavior is equivalent — `setActiveCategory(null)` still fires when `isVisible` transitions to `false`, and the hex/gradient parser still runs on each distinct `activeCustomValue`. The new pattern triggers an immediate re-render after `setPrev*`, which React collapses into the same commit, so no extra paint is observable.
- Minor note (not a blocker): multiple setState calls during render inside the `activeCustomValue` branch will cause React to bail and re-render once. This is the documented pattern and is fine here, but worth flagging for future readers who may see three setters during render and assume it's a bug.

### Missing Tests
- No dedicated test for `SidebarBackgrounds`. A small render test that mounts with `isVisible=true`, selects a category, toggles `isVisible=false`, and asserts the detail panel collapses would pin the behavior — but this is a preexisting gap, not introduced by this PR.

### Overall
**Merge readiness:** Ready
All 6 CI checks pass. Clean implementation of the preferred React pattern; the change is exactly what CLAUDE.md recommends.

---

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-04-21

### Summary
Converts two `useEffect`-based prop/state synchronizations in `SidebarBackgrounds.tsx` to the "adjusting state while rendering" pattern — the preferred React pattern explicitly documented in CLAUDE.md's "useEffect is an escape hatch, not a default" section.

### Code Quality
- ✓ Correct implementation of the pattern: `const [prevX, setPrevX] = useState(x); if (x !== prevX) { setPrevX(x); /* reset state */ }` runs during render and avoids an extra render pass.
- ✓ The `activeCustomValue` refactor correctly restructures the nested logic: the early `return` is replaced with an `if/else`, preserving the original short-circuit semantics.
- ✓ The `isVisible` reset correctly only clears `activeCategory` when transitioning to hidden (no change when transitioning to visible).
- ✓ `useEffect` is still used elsewhere in the file for genuine external-system synchronization (Google Drive fetch) — correct use of the escape hatch.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No rules or auth changes.

### Regression Risk
- ⚠ Low. Behavior is preserved; the only observable change is eliminating one extra render pass per `isVisible` or `activeCustomValue` transition.
- ⚠ Watch for any callers that depended on the effect's `batched` vs. `render-phase` timing, though none are expected since the previous effects also did local state updates only.

### Missing Tests
- ⚠ No component test for `SidebarBackgrounds.tsx`. Not a gap introduced by this PR, but a focused test that mounts the component, toggles `isVisible`, and asserts `activeCategory` resets would prevent regressions.

### Overall
**Merge readiness:** Ready
Clean, idiomatic refactor that matches a preference explicitly called out in CLAUDE.md. No behavior change. Safe to merge.

---

---

PR #1359, opened 2026-04-20
feat(org): add organizationMemberCounters CF for building/org/domain user counts

OPS-PIvers:
## Summary

- Fixes the bug where reassigning a user's building didn't update the Buildings list "Users" column. The denormalized `users` field on `/organizations/{orgId}`, `/buildings/{id}`, and `/domains/{id}` was only maintained by the manual `scripts/recount-org-members.js` stopgap until now.
- Adds `organizationMemberCounters` — an `onDocumentWritten` trigger on `organizations/{orgId}/members/{emailLower}` that applies `FieldValue.increment()` deltas to all three counter sets atomically.
- Closes the Phase 4.1 backlog entry for "Counter-maintenance Cloud Function" in `docs/organization_wiring_implementation.md`.

## Design notes

- **Pure helper + trigger split.** `planMemberCounterDeltas(before, after)` returns `{ orgDelta, buildingDeltas, emailDomainDeltas }` with zero-value entries pruned. The trigger applies deltas via `FieldValue.increment()`. Same split pattern as `organizationMembersSync` + `computeAdminAction`.
- **Won't resurrect deleted buckets.** Building and domain increments use `update()` (not `set({merge:true})`), so a stale member doc pointing at a deleted building gets "No document to update" → logged + skipped rather than re-creating the building.
- **Never throws.** Firestore retries thrown triggers; on a counter trigger that would double-count. Mirrors the discipline already documented on `organizationMembersSync`.
- **Domain resolution.** Email domain → doc id lookup is tolerant of stored `@`-prefix vs. no-prefix (`useOrgDomains` writes with `@`, the recount script normalizes both; this CF matches the script).
- **Reconcile tool stays.** `scripts/recount-org-members.js` remains the ground-truth rebuild if a partial failure drifts counters. Both `planMemberCounterDeltas()` and the script's `tallyMembers()` agree on bucket identity.

## Tests

21/21 unit cases in `functions/src/organizationMemberCounters.test.ts`:

- create / delete / no-op update
- add one building / reassign building / remove all buildings / the reported "previously unassigned → assigned" bug
- email domain change moves count between buckets
- missing `buildingIds` / empty-string ids / duplicate ids / missing email
- `resolveDomainDocId` with `@`-prefix, no-prefix, no-match, empty bucket, non-string field, case-insensitive

Following the pattern of `organizationMembersSync.test.ts` — executable documentation until the Phase 4.1 functions-scoped test runner lands. Root `pnpm run type-check:all`, `lint`, and `format:check` all pass.

## Test plan

- [ ] Deploy to `spartboard` (`firebase deploy --only functions:organizationMemberCounters`)
- [ ] In the Organization panel → Buildings view, note current user count on a building (e.g. Community Education = 0)
- [ ] Edit a user with no building assignment and assign them to that building
- [ ] Refresh Buildings view → user count should increment by 1 within a few seconds
- [ ] Reassign the same user to a different building → old building -1, new building +1
- [ ] Remove the user entirely → org counter -1, last building -1, domain -1
- [ ] Cross-check totals with `node scripts/recount-org-members.js --dry-run`; numbers should match. If they don't, run the script for real — the CF will stay in sync from that snapshot forward.

https://claude.ai/code/session_01N3BEV1MrkRZjTSkKLcxhBy


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-04-21

### Summary
Adds `organizationMemberCounters` — an `onDocumentWritten` trigger on `organizations/{orgId}/members/{emailLower}` that maintains denormalized `users` counters on the org doc plus each referenced building and domain doc via `FieldValue.increment()` deltas. Fixes the reported bug where reassigning a user's building didn't update the Buildings "Users" column, closing the Phase 4.1 backlog entry. Pure helper (`planMemberCounterDeltas`) / trigger split with 21/21 unit test cases.

### Code Quality
- ✓ Helper + trigger split matches the project convention (`organizationMembersSync` + `computeAdminAction`).
- ✓ `update()` (not `set({merge:true})`) on building + domain docs — correctly ensures a stale reference to a deleted bucket fails with "No document to update" rather than resurrecting the doc.
- ✓ Domain resolution is case-insensitive and tolerant of stored `@`-prefix vs. no-prefix, matching `scripts/recount-org-members.js`.
- ✓ Never-throws discipline is explicit and justified with a clear rationale.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
- ✓ No rules changes required — the trigger runs under Admin SDK so it bypasses rules by design.
- ✓ Trigger scope (`organizations/{orgId}/members/{emailLower}`) correctly matches the existing member-write pattern.

### Regression Risk
- ⚠ **At-least-once delivery.** Copilot's unresolved thread is correct: even without throws, Firestore/Eventarc can deliver the same event twice, which would double-apply a `FieldValue.increment()` delta. The CF's current defense is "don't throw" (avoiding handler retries), not idempotency. The `recount-org-members.js` script remains the authoritative repair path. Worth landing with an explicit doc note that rare drift is still possible, and consider an event-id dedupe marker for stronger guarantees if drift becomes observable.
- ⚠ Cross-document updates (`updateOrgCounter`, `updateBuildingCounter`, `updateDomainCounter`) are separate `update()` calls, not a batch/transaction. Describing them as "atomic" in the PR body is slightly misleading — `FieldValue.increment()` is atomic per document, but cross-doc can partially succeed. This is mentioned in the Copilot comments; not a blocker but worth tightening the doc prose.

### Missing Tests
- ✓ 21 unit cases cover create/delete/no-op/reassign/remove-all/missing-fields/duplicate-ids/domain-change/resolveDomainDocId edge cases — thorough.
- ⚠ **CI does not run these tests.** Copilot correctly flagged that `vitest.config.ts` excludes `functions/**` from the default `pnpm test`. `organizationMemberCounters.test.ts` is currently "executable documentation" — protects against local regressions but not CI regressions. Phase 4.1 backlog item. Worth wiring `pnpm -C functions test` into `validate` before the next CF PR.

### Overall
Clean implementation of a high-value Phase 4.1 trigger. Recommend (a) accepting Copilot's doc-prose suggestions about at-least-once delivery + cross-doc-atomicity caveats, and (b) filing a follow-up to wire `functions/**` tests into CI. Functional risk is low — the recount script remains the ground-truth repair path.

---

---

PR #1375, opened 2026-04-21
fix(admin): scope analytics to org + sync buildings counter

OPS-PIvers:
## Summary

Fixes three misalignments between the Organization and Analytics pages of the Admin Settings console. All three stem from code paths that predate the multi-tenant `/organizations/{orgId}/…` data model.

### 1. Stale `buildings` counter on Organizations table
The denormalized `organizations/{orgId}.buildings` counter was seeded at `4` by `scripts/setup-organization.js` and never updated when buildings are added/removed via the Buildings tab.

- **New Cloud Function** `organizationBuildingCounters` (`functions/src/organizationBuildingCounters.ts`) — `onDocumentWritten` trigger on `/organizations/{orgId}/buildings/{buildingId}` that recounts the subcollection on every create/delete and writes the result to the org doc. Uses recount (not `increment`) so it self-heals existing drift from the seed value.
- Exported from `functions/src/index.ts`.
- `scripts/recount-org-members.js` extended to also set `buildings` as a one-shot reconcile.

Firestore rules already forbid client writes to the `buildings` field on the org doc, so the trigger (Admin SDK) is the only path that updates it.

### 2. Foreign-domain users showing up in Analytics
`adminAnalytics` was calling `admin.auth().listUsers()` globally with no org filter, so any user (e.g. `@ashwaubenonk12.org`) who had ever signed in was included.

- `adminAnalytics` now requires `orgId` in the request body (400 if missing).
- Caller must be a super admin **or** a member of the requested org (403 otherwise).
- Replaced the global `listUsers()` loop with a scan of `/organizations/{orgId}/members/`. Auth metadata resolved via `admin.auth().getUsers()` in chunks of 100 using each member's `uid`. Members without a `uid` (never signed in) are counted in `total` but report `monthly: 0, daily: 0`.
- Client (`AnalyticsManager.tsx`) now sends `orgId` from `useAuth()` in the request body and re-fetches when the active org changes.

### 3. "Unknown Building (…)" labels in Analytics
Analytics was bucketing by the user's self-selected `selectedBuildings` profile field instead of the admin-assigned `buildingIds` on the member record. When the two drifted, the building lookup in `useKnownBuildings()` failed.

- Backend now buckets by `member.buildingIds` — the authoritative admin-assigned list from Organization > Users. These IDs are guaranteed to match live building docs because they come from the same write path.

## Test plan
- [x] `pnpm run validate` passes locally (1393 frontend tests + 134 functions tests, lint clean, format clean, typecheck clean)
- [ ] In a dev build signed in as super admin for Orono: add/remove a building, confirm the Organizations table cell updates immediately
- [ ] Manually set `organizations/{orgId}.buildings` to a wrong value in Firestore console, add or remove a building, confirm the trigger writes the correct count
- [ ] Confirm `ashwaubenonk12.org` no longer appears in the Analytics domain filter or domains table for Orono
- [ ] Total user count in Analytics matches Organization > Users rows (minus members without `uid`)
- [ ] Call `adminAnalytics` with missing `orgId` → expect 400
- [ ] Call `adminAnalytics` with a non-member / non-admin token → expect 403
- [ ] Every non-"No Building Assigned" row in "Users by Building" shows a real building name

## Deploy notes
Cloud Function changes require `firebase deploy --only functions:adminAnalytics,functions:organizationBuildingCounters`. Front-end deploys via the normal dev/main pipeline.

## Out of scope (follow-up)
- Migrating existing `selectedBuildings` profile fields to match `buildingIds` (audit needed — `context/AuthContext.tsx`, `components/layout/sidebar/SidebarBuildings.tsx`)
- Super-admin "all orgs" aggregate analytics view

https://claude.ai/code/session_017o2qhrc64fbt6bdct7yRPa

### DIFF
--- a/components/admin/Analytics/AnalyticsManager.tsx
+++ b/components/admin/Analytics/AnalyticsManager.tsx
@@ -1375,6 +1375,11 @@ export const AnalyticsManager: React.FC = () => {
     if (!orgId) {
       // Org membership hasn't resolved yet; skip the request rather than
       // firing one that the backend would reject for missing org scope.
+      // Clear the initial loading flag so the UI shows an empty state instead
+      // of a permanent spinner for users who don't have an active org.
+      if (isMountedRef.current) {
+        setLoading(false);
+      }
       return;
     }
     const requestId = requestSequenceRef.current + 1;
--- a/components/admin/Analytics/AnalyticsManager.tsx
+++ b/components/admin/Analytics/AnalyticsManager.tsx
@@ -1377,6 +1377,11 @@ export const AnalyticsManager: React.FC = () => {
       // explicit message rather than a permanent spinner or a blank page. This
       // also handles the case where `orgId` flips back to null after a prior
       // successful load (e.g., the user is removed from the org).
+      //
+      // Bump the request sequence before clearing state so any in-flight
+      // response from a prior org fails its `requestId === requestSequenceRef`
+      // guard and cannot overwrite the cleared state with stale analytics.
+      requestSequenceRef.current += 1;
       if (isMountedRef.current) {
         setLoading(false);
         setData(null);
--- a/components/admin/Analytics/AnalyticsManager.tsx
+++ b/components/admin/Analytics/AnalyticsManager.tsx
@@ -35,6 +35,7 @@ import {
 } from 'lucide-react';
 import { Modal } from '@/components/common/Modal';
 import { useAdminBuildings } from '@/hooks/useAdminBuildings';
+import { useAuth } from '@/context/useAuth';
 import type { Building } from '@/config/buildings';
 import { TOOLS } from '@/config/tools';
 
@@ -1350,6 +1351,7 @@ const DataTable: React.FC<{
 
 export const AnalyticsManager: React.FC = () => {
   const KNOWN_BUILDINGS = useKnownBuildings();
+  const { orgId } = useAuth();
   const [loading, setLoading] = useState(true);
   const [data, setData] = useState<AnalyticsData | null>(null);
   const [error, setError] = useState<string | null>(null);
@@ -1370,6 +1372,11 @@ export const AnalyticsManager: React.FC = () => {
   }, []);
 
   const fetchAnalytics = useCallback(async () => {
+    if (!orgId) {
+      // Org membership hasn't resolved yet; skip the request rather than
+      // firing one that the backend would reject for missing org scope.
+      return;
+    }
     const requestId = requestSequenceRef.current + 1;
     requestSequenceRef.current = requestId;
 
@@ -1387,7 +1394,7 @@ export const AnalyticsManager: React.FC = () => {
           'Content-Type': 'application/json',
           Authorization: `Bearer ${token}`,
         },
-        body: JSON.stringify({}),
+        body: JSON.stringify({ orgId }),
       });
 
       if (!response.ok) {
@@ -1450,7 +1457,7 @@ export const AnalyticsManager: React.FC = () => {
         setLoading(false);
       }
     }
-  }, []);
+  }, [orgId]);
 
   useEffect(() => {
     void fetchAnalytics();
--- a/components/admin/Analytics/AnalyticsManager.tsx
+++ b/components/admin/Analytics/AnalyticsManager.tsx
@@ -1373,12 +1373,16 @@ export const AnalyticsManager: React.FC = () => {
 
   const fetchAnalytics = useCallback(async () => {
     if (!orgId) {
-      // Org membership hasn't resolved yet; skip the request rather than
-      // firing one that the backend would reject for missing org scope.
-      // Clear the initial loading flag so the UI shows an empty state instead
-      // of a permanent spinner for users who don't have an active org.
+      // No active org: clear any stale data from a previous org and surface an
+      // explicit message rather than a permanent spinner or a blank page. This
+      // also handles the case where `orgId` flips back to null after a prior
+      // successful load (e.g., the user is removed from the org).
       if (isMountedRef.current) {
         setLoading(false);
+        setData(null);
+        setError(
+          'No organization selected. Analytics require an active organization.'
+        );
       }
       return;
     }
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -2017,13 +2017,30 @@ export const adminAnalytics = onRequest(
 
     const db = admin.firestore();
 
-    // 2. Verify caller is an admin (super admin in /admins, or a member of
-    // the requested org — org-scoped admins should see their own analytics).
+    // 2. Verify caller is authorized for the requested org. Two paths:
+    //   - Super admin: exists in `/admins/{email}`. May view any org.
+    //   - Org admin: has a member doc at `/organizations/{orgId}/members/{email}`
+    //     whose `roleId` is in the admin-tier set. Mirrors the role gating in
+    //     `assertCallerIsOrgAdmin` (organizationInvites.ts) but also admits
+    //     building_admin, since reading analytics is a lesser privilege than
+    //     inviting members.
+    const ORG_ADMIN_ROLE_IDS = new Set([
+      'super_admin',
+      'domain_admin',
+      'building_admin',
+    ]);
     const [adminDoc, memberDoc] = await Promise.all([
       db.collection('admins').doc(email).get(),
       db.doc(`organizations/${orgId}/members/${email}`).get(),
     ]);
-    if (!adminDoc.exists && !memberDoc.exists) {
+    const memberData = memberDoc.exists
+      ? (memberDoc.data() as { roleId?: unknown })
+      : undefined;
+    const memberRoleId =
+      typeof memberData?.roleId === 'string' ? memberData.roleId.trim() : '';
+    const isSuperAdmin = adminDoc.exists;
+    const isOrgAdmin = memberDoc.exists && ORG_ADMIN_ROLE_IDS.has(memberRoleId);
+    if (!isSuperAdmin && !isOrgAdmin) {
       console.error(
         `[getAdminAnalytics] Unauthorized access: ${email} is not an admin of ${orgId}`
       );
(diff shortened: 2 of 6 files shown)


functions/src/index.ts
OPS-PIvers, line 1: **Authorization too permissive — any org member can access analytics**

The new auth check is:
```ts
if (!adminDoc.exists && !memberDoc.exists) { ... 403 ... }
```

This grants full analytics access to *every* member of the org, not just admins. Analytics data includes last sign-in times, per-building user counts, and domain breakdowns — data that regular teachers shouldn't see. The PR description says "Caller must be a super admin or a member of the requested org", but that scope is too broad.

Consider requiring the caller to be a super-admin **or** a domain admin (e.g. check `memberDoc.data()?.role` against an admin role id, or reuse an existing `isDomainAdmin` helper). A regular teacher who happens to know the function URL should get a 403 here.

---

components/admin/Analytics/AnalyticsManager.tsx
OPS-PIvers, line 1: **Infinite loading spinner when `orgId` is null**

`fetchAnalytics` early-returns when `!orgId`:
```ts
if (!orgId) {
  return; // loading stays true forever
}
```

Because `loading` initializes to `true` and `setLoading(false)` is only called later in the happy path / catch block, users without an org (or during the brief window before `orgId` resolves from Firestore) see a permanent spinner with no error state.

Fix: call `setLoading(false)` before the early return, or initialize `loading` to `false` and set it `true` only after the `!orgId` guard passes.

---

functions/src/organizationBuildingCounters.ts
OPS-PIvers, line 1: **No test coverage for this new Cloud Function**

`organizationMemberCounters.ts` ships with `organizationMemberCounters.test.ts`. This new function has no corresponding test file. A few focused tests would give confidence in the recount logic — specifically:
- create trigger → count increments to correct value
- delete trigger → count decrements
- update-only event (no create/delete) → no write issued (the early return on `!created && !deleted`)
- pre-existing drift is self-healed (start with a wrong `buildings` value, fire a create, confirm the stored value equals the real count)

---

OPS-PIvers (commented):
Three issues found — none are blockers on their own but the authorization one is worth fixing before this lands in production.

---

functions/src/organizationBuildingCounters.ts
```diff
@@ -0,0 +1,64 @@
...
+        buildingId,
+        count,
+        trigger: created ? 'create' : 'delete',
+      });
+    } catch (err) {
+      logger.error('organizationBuildingCounters: recount failed', {
+        orgId,
+        buildingId,
+        error: err instanceof Error ? err.message : String(err),
+      });
```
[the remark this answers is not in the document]
OPS-PIvers: Agreed — addressed in `060f206`. The `catch` block now rethrows (wrapping non-Error values) so Cloud Functions retries the invocation, which is the correct behavior for a self-healing recount trigger. See `functions/src/organizationBuildingCounters.ts:65-69`.

---

functions/src/index.ts
```diff
@@ -2000,88 +2001,116 @@ export const adminAnalytics = onRequest(
...
+      // Firebase Auth account regardless of org and caused foreign-domain
+      // users to show up in a different org's analytics.
+      //
+      // For each member we need auth metadata (lastSignInTime) to compute
+      // engagement. Members without a `uid` (invited but never signed in)
+      // still count toward totals but have zero engagement.
+      interface MemberLite {
+        email: string;
+        uid: string | null;
+        buildingIds: string[];
```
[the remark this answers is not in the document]
OPS-PIvers: Good catch — fixed in `97c14c1`. Engagement buckets (`totalEngagement`, `usersByDomain`, `usersByBuilding`, `usersByDomainAndBuilding`) now iterate the full `members` roster (`functions/src/index.ts:2242`), and auth metadata (`lastSignInMs`) is joined in only when a `uid` resolves. Members without a `uid` still count toward totals/domain/building buckets but report zero engagement, matching the contract in the code comment.

---

functions/src/index.ts
```diff
@@ -2000,88 +2001,116 @@ export const adminAnalytics = onRequest(
...
+    // globally, which leaked foreign-domain accounts into the calling admin's
+    // analytics view.
+    const rawBody = req.body as { orgId?: unknown } | undefined;
+    const orgId =
+      rawBody && typeof rawBody.orgId === 'string' ? rawBody.orgId.trim() : '';
+    if (!orgId) {
+      res
+        .status(400)
+        .json({ error: 'invalid-argument', message: 'orgId is required' });
+      return;
```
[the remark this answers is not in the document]
OPS-PIvers: Added in `060f206`. `functions/src/index.test.ts` now covers both the 400 path ("returns 400 when orgId is missing from the request body", line 882) and the 403 path ("returns 403 when caller is neither super admin nor an org admin", line 910).

---

functions/src/index.ts
OPS-PIvers: Fixed in `060f206`. The auth check now requires the caller to be either a super admin (exists in `/admins/{email}`) **or** have an org-scoped member doc whose `roleId` is in `ORG_ADMIN_ROLE_IDS = {'super_admin', 'domain_admin', 'building_admin'}`. A regular teacher member of the org will no longer pass this gate. See `functions/src/index.ts:2020-2049`.

---

components/admin/Analytics/AnalyticsManager.tsx
OPS-PIvers: Fixed in `060f206`. The `!orgId` branch now calls `setLoading(false)`, clears `data`, and sets an explicit error message before returning (`components/admin/Analytics/AnalyticsManager.tsx:1374-1388`), so the UI renders a "No organization selected" error state instead of a permanent spinner.

---

functions/src/organizationBuildingCounters.ts
OPS-PIvers: Added in `060f206` — see `functions/src/organizationBuildingCounters.test.ts`. It covers:
- create trigger → writes the recounted value
- delete trigger → writes the recounted value
- update-only event → no write issued (early return)
- pre-existing drift is self-healed (live count overrides stored value)
- recount failures rethrow so Cloud Functions retries

All five cases pass.

---

functions/src/index.ts
```diff
@@ -2000,88 +2001,136 @@ export const adminAnalytics = onRequest(
...
+            authUsersMap.set(u.uid, {
+              email: u.email ?? '',
+              lastSignInMs: lastSignIn,
+            });
+          }
+        } catch (err) {
+          console.warn('[getAdminAnalytics] auth().getUsers() chunk failed', {
+            orgId,
+            chunkSize: chunk.length,
+            error: err instanceof Error ? err.message : String(err),
```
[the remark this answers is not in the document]
OPS-PIvers: Valid concern — fixed in `97c14c1`. The code now builds a `memberUids` set (`functions/src/index.ts:2126-2129`) separately from `authUsersMap`, and downstream filters use `memberUids.has(uid)` for dashboard membership (line 2182) and AI usage (line 2438). `authUsersMap` is reserved for joining `lastSignInMs` only — a transient `auth().getUsers()` chunk failure can only affect last-sign-in, not user/dashboard/AI totals.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-04-22

### Summary
Three well-targeted fixes to `adminAnalytics` that undo pre-multi-tenant assumptions: (1) a new `organizationBuildingCounters` Firestore trigger that self-heals the denormalized `buildings` counter on the org doc, (2) strict org scoping on `adminAnalytics` with 400/403 gating and member-scoped iteration instead of `listUsers()` globally, (3) building labels driven by admin-assigned `buildingIds` instead of user-selected profile fields. Ships with focused unit tests for both the trigger and the 400/403 paths.

### Code Quality
✓ Clean. All reviewer comments (copilot + the author's own self-review) have been addressed in commits `060f206` and `97c14c1`. The `auth().getUsers()` chunk-failure path now scopes downstream filters by a separate `memberUids` set so auth lookup failures only affect last-sign-in (not totals). Role gating uses an explicit `ORG_ADMIN_ROLE_IDS` set that mirrors `assertCallerIsOrgAdmin` in organizationInvites.ts but also admits `building_admin` (correct — analytics is lesser privilege than invites). No `any` without cast, no console.log, no dead code in the final diff (`buildingsMap` from the outdated review was removed).

### Widget Standards
✓ Not applicable — no widget code touched.

### Admin Config Standards
✓ Not applicable in the new-widget-config sense. Changes to `AnalyticsManager.tsx` correctly consume `orgId` from `useAuth()` and re-fetch when it changes. The `!orgId` branch now clears data + sets an explicit error instead of rendering a blank page.

### Firestore & Security
✓ No new rules needed. The trigger runs on Admin SDK so existing rules forbidding client writes to `organizations/{orgId}.buildings` continue to apply. Authorization flow is now properly scoped:
- Super admin (`/admins/{email}`) may view any org.
- Org-scoped admins require both a member doc at `/organizations/{orgId}/members/{email}` *and* a `roleId` in the admin-tier set.
- Plain members (teachers) no longer pass the gate — this fixes the permissiveness flagged by the reviewer.

One minor security observation (non-blocking): the 400/403 responses return error codes (`invalid-argument`, `permission-denied`) but no `requestId` or correlation token. If you have a Cloud Logging alert policy on `[getAdminAnalytics]` messages, a missing correlation id makes post-hoc triage harder. Worth a follow-up, not a merge blocker.

### Regression Risk
**Low.** Specific checks:

- **`context/DashboardContext.tsx`** — not touched. `getAdminBuildingConfig` pipeline unaffected.
- **`components/widgets/WidgetRegistry.ts`** — not touched.
- **`types.ts`** — not touched.
- **`functions/src/index.ts`** — signature of `adminAnalytics` is extended with a required `orgId` in the request body. The client is updated in the same PR to always send it, so no client ↔ server mismatch. Old clients without the updated UI will hit the new 400 — acceptable since `AnalyticsManager` is the only caller.
- **`organizationBuildingCounters`** — new function, new deployable name. Requires `firebase deploy --only functions:organizationBuildingCounters` as called out in the PR description. If the deploy misses this function, existing behavior is preserved (stale counter) — no crash risk.
- **`scripts/recount-org-members.js`** — one-shot reconcile extended to also set `buildings`. Additive change, backwards compatible with prior runs.

### Missing Tests
✓ Adequate. `organizationBuildingCounters.test.ts` covers create/delete/update-only/drift-self-heal/rethrow paths. `index.test.ts` now covers both 400 (missing orgId) and 403 (non-admin caller). One gap worth considering as follow-up: a unit test for the new "iterate `members` for totals, join `authUsersMap` for engagement" flow — e.g. an invited-but-never-signed-in member counts toward totals/domain/building buckets with zero engagement. The existing mock returns `exists: false` for `getAll`, which sidesteps this path; a dedicated test would lock in the stated contract.

### Overall
**Merge readiness:** Ready
Focused fix, reviewer-round-tripped, good test coverage for new code paths. Follow the deploy note (`firebase deploy --only functions:adminAnalytics,functions:organizationBuildingCounters`) and verify the trigger is actually deployed before smoke-testing — otherwise the `buildings` counter self-heal will silently no-op.

---

---

PR #1390, opened 2026-04-22
fix(rules): split session reads into get+list to unbreak teacher assignment creation

OPS-PIvers:
## Summary

- PR #1389's hoisted `!isStudentRoleUser()` short-circuit was insufficient. Firestore's static analyzer rejects `list` predicates that touch `resource.data` even when an earlier OR branch is provably true at runtime.
- Verified in production by two independent accounts (super-admin + sample teacher) — both fail identically with `permission-denied` toasts when clicking Assign on a quiz from the library, and the most recent successful `quiz_session` doc in the database predates the SSO deploy by ~36 hours.
- Splits `allow read` into separate `allow get` + `allow list` rules on all five session collections. The `list` predicate now contains zero `resource.data` references; the `get` rule keeps the studentRole class gate intact.

## Behavior after fix

| Caller | Before | After |
|---|---|---|
| Teacher (`useQuizAssignments.allocateJoinCode`, etc.) | denied | passes |
| Anonymous PIN student (`joinQuizSession`, etc.) | denied | passes |
| ClassLink-via-Google studentRole user | denied | denied on `list`, gated on `get` (intentional — they reach sessions via assignments page, not by typing codes) |

## Collections updated

- `quiz_sessions`
- `video_activity_sessions`
- `guided_learning_sessions`
- `activity_wall_sessions`
- `mini_app_sessions` (same pattern via `passesStudentClassGateList` — likely silently broken too, just less production exercise)

## Test plan

- [ ] Wait for dev-paul preview deploy
- [ ] Sign in as super-admin → click Assign on a quiz from library → confirm assignment appears under In Progress
- [ ] Sign in as a sample teacher → repeat → confirm same
- [ ] PIN-flow student joins via code → confirm `joinQuizSession` still works
- [ ] If green, promote to main

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -577,26 +577,33 @@ service cloud.firestore {
     // owns the document and stores their uid on the `teacherUid` field.
     // Multiple concurrent sessions per teacher are allowed.
     match /quiz_sessions/{sessionId} {
-      // Students need both `get` (single-doc reads after joining) and `list`
-      // (collection query by join code via `where('code', '==', ...)`) access.
-      // Restricting enumeration fully would require a Cloud Function proxy;
-      // for now we accept that authenticated users can list session documents.
+      // Reads are split into `get` (single doc) and `list` (collection query)
+      // because Firestore's rules engine treats them differently:
       //
-      // studentRole (GIS-authenticated) users must additionally have the
-      // session's classId in their classIds custom claim. Non-student callers
-      // (teachers, admins, anonymous PIN students) are unaffected.
+      //   - `get` evaluates per-document at read time. The studentRole class
+      //     gate can safely reference `resource.data` because the engine has
+      //     a concrete document to evaluate against.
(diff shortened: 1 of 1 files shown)


OPS-PIvers:
@gemini-code-assist you're right — the first cut blocked studentRole users from listing all 5 session collections, which broke `MyAssignmentsPage` (it does `where('classId', 'in', myClassIds)` on quiz/VA/GL/AW and `array-contains-any` on mini-app).

Pushed a revision (29d7906d) that drops `!isStudentRoleUser()` from the `allow list` rule on all 5 collections. studentRole users can now list session metadata, but per-class gating is still enforced at `get` (and at write time on responses), so they still cannot **open** a session outside their `classIds` claim — only enumerate metadata (title, classId, code, teacherUid).

The privacy tradeoff (any authenticated user can enumerate session metadata) matches the pre-SSO behavior. Tightening enumeration further would require a Cloud Function proxy, which is out of scope for this hotfix.

PR #1432, opened 2026-04-27
fix(quiz): stop mounting teacher AuthProvider on the /quiz student route

OPS-PIvers:
## Summary

- The `/quiz` route was wrapping `QuizStudentApp` in `<AuthProvider>` (the teacher AuthContext), which unconditionally subscribes to `admin_settings/user_roles` and `admin_settings/app_settings` for any authenticated user. Both are admin-only per `firestore.rules`, so non-admin students hit *Missing or insufficient permissions* in the console on every quiz load — surfaced more visibly after #1431 gave SSO students an already-authenticated identity at mount time.
- `QuizStudentApp` self-handles Firebase auth (`signInAnonymously` when no current user; preserves the SSO student-custom-token user otherwise) and nothing in the quiz subtree consumes `useAuth()`, so the teacher provider was unnecessary. Drop it to match every other student-only route.
- Bonus: wraps `reportTabSwitch`'s `updateDoc` in a `try`/`catch` that logs `sessionId`, `responseKey`, `authUid`, `isAnonymous`, `baseCount`, `hasPinField`, `hasTabSwitchField` before re-throwing — so the next intermittent *Failed to report tab switch* denial gives us enough state to bisect without a speculative rule change.

## Test plan

- [x] `pnpm run type-check` clean
- [x] `pnpm run format:check` clean
- [x] `pnpm exec eslint App.tsx hooks/useQuizSession.ts` clean
- [x] `pnpm exec vitest run tests/hooks/useQuizSession.test.ts` — 43/43 pass
- [x] Local preview at `/quiz?code=…` — PIN form renders; no `Error loading user roles`, `Error loading app settings`, or `Failed to report tab switch` in the post-navigation console
- [ ] Dev preview as an SSO student opening a quiz from `/my-assignments` — confirm clean console, answers persist, tab-switch warning increments
- [ ] Dev preview as an anonymous student via direct `/quiz?code=…` — PIN flow still gates entry, sign-in-anonymously still fires, tab switch still works

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -305,15 +305,17 @@ const App: React.FC = () => {
     );
   }
 
-  // Quiz student route — requires real Firebase auth (org Google account)
+  // Quiz student route — QuizStudentApp self-handles Firebase auth
+  // (signInAnonymously when no current user; preserves the SSO student-
+  // custom-token user otherwise). No teacher AuthContext consumers in the
+  // quiz tree, so wrapping in <AuthProvider> would only mount admin-only
+  // Firestore listeners that fail with permission-denied for students.
   if (isQuizRoute) {
     return (
       <DialogProvider>
-        <AuthProvider>
-          <Suspense fallback={<FullPageLoader />}>
-            <QuizStudentApp />
-          </Suspense>
-        </AuthProvider>
+        <Suspense fallback={<FullPageLoader />}>
+          <QuizStudentApp />
+        </Suspense>
         <DialogContainer />
       </DialogProvider>
     );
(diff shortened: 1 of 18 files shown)


OPS-PIvers:
Closing — opened against the wrong base. The fix should land on `dev-paul` first (matching the workflow used by #1430 and #1431), then ride into `main` with the next `dev-paul → main` integration. Reopened as a focused PR against `dev-paul` from branch `claude/quiz-route-auth-provider`. The Gemini import-style nits flagged here are on files from #1431 and aren't in scope of my AuthProvider fix; they should be addressed in a separate cleanup if wanted.

PR #1532, opened 2026-05-06
🎨 TimeTool: inline control row & reclaim wasted space

OPS-PIvers:
## Summary

- Merges the ± adjust buttons into the same horizontal row as play/pause/reset (`[− step] [play/pause] [reset] [+ step]`), so all four controls read as one deliberate group instead of two competing stacks above and below the time.
- Standardizes all four control buttons to `min(56px, 14cqmin)` (cqmin per CLAUDE.md scaling rules), so play/pause and reset no longer look oversized in visual mode. Play/pause keeps `bg-brand-blue-primary` for primary-action emphasis.
- Reclaims wasted space: progress ring bumps from 90% → 96% of cqmin so it nearly reaches the widget edges, and the digital-mode time font cap grows from `min(55cqh, 25cqw)` → `min(60cqh, 30cqw)` (stopwatch `18cqw` → `22cqw`) so the time fills more of the widget face. Verified in the dev preview the time text grew from ~107px to 125px tall on a default 400×400 widget.

The `showAdjustControls` guard now wraps just the two `AdjustButton` siblings, so stopwatch and fresh-setup states still hide ± while keeping play/reset visible. No changes to `useHoldAccelerate`, `useTimeTool`, settings, or registry — behavior of all four buttons (state, hold-to-accelerate ramp, clamping `adjustStepSeconds` to 5–60) is preserved.

## Test plan

- [x] `pnpm run validate` — type-check, lint (zero warnings), format-check, all tests pass (1966 + 193)
- [x] Rewrote the `±adjust button placement` describe block to assert the new layout: single absolute wrapper anchored from `top:`, all four buttons share the wrapper in `[−, play/pause, reset, +]` DOM order, `top:110%` digital / `top:120%` visual, plus explicit coverage for stopwatch and fresh-setup states hiding ±
- [x] Browser preview verified: 4 buttons rendered in correct order at 56×56 px each, wrapper has `top: 110%; gap: min(12px, 3cqmin)`, time text grew from ~107px → 125.4px tall (computed font-size resolved from `min(60cqh, 30cqw)`)
- [x] Fresh-setup state (timer paused at original duration) — ± hidden until timer is started or off-default
- [ ] Manual sanity check on the preview deploy: switch a timer to Visual mode and confirm the ring nearly fills the widget face
- [ ] Manual sanity check: stopwatch mode renders only play/reset (no ±)
- [ ] Manual sanity check: press-and-hold ramp on `+` still accelerates as before

### DIFF
--- a/components/widgets/TimeTool/TimeToolWidget.test.tsx
+++ b/components/widgets/TimeTool/TimeToolWidget.test.tsx
@@ -640,6 +640,8 @@ describe('TimeToolWidget', () => {
       renderWidget(widget);
 
       const subBtn = screen.getByLabelText('Subtract time');
+      // Timer is running, so play/pause shows the Pause label.
+      const pauseBtn = screen.getByLabelText('Pause');
       const resetBtn = screen.getByLabelText('Reset');
       const addBtn = screen.getByLabelText('Add time');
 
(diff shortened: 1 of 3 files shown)


OPS-PIvers:
Addressed in 31476bfd:

- Play/pause button now has a dynamic `aria-label` that flips between `Play` and `Pause` based on `isRunning` (using i18n keys `widgets.timeTool.play` / `widgets.timeTool.pause` to match the existing pattern of the ± buttons in this widget).
- Added `select-none` to both play/pause and reset to prevent accidental text selection during rapid taps.
- Bonus consistency win: migrated the reset button's hard-coded `aria-label="Reset"` to `t('widgets.timeTool.reset')` so it's translatable too.

Verified in the dev preview:
- `aria-label` reads `Play` when paused, `Pause` when running (toggles correctly).
- Both buttons have `select-none` applied.
- All 43 TimeTool tests still pass; full `pnpm validate` is green (1966 + 193).

PR #1539, opened 2026-05-07
feat(video-activity): SSO bypass + drop name field; mirror Quiz response keying

OPS-PIvers:
## Summary

Bring Video Activity student-experience to parity with the Quiz widget.

- **SSO students** auto-join with no PIN, no period picker, no name input — identified by their `studentRole` custom-token UID, period resolved from `session.classPeriodByClassId`.
- **Anonymous students** enter PIN + period only (the name field is gone). Response docs are now keyed `pin-{period}-{pin}` so the same PIN in different periods stays distinct and attempt limits will survive a device wipe.
- **Foundation for the assignment-options follow-up (PR1b):** `BaseSessionOptions` is extracted from `QuizSessionOptions`, and a new `VideoActivitySessionOptions` shape carries the VA-specific knobs (`rewindOnIncorrectSeconds`, `pointPenaltyOnIncorrect`, `attemptLimit`, `scoreVisibility`). All additions are optional and not yet consumed by UI.

## Why this PR is split from the original PR1 plan

The approved plan called for one PR1 covering both student-experience parity AND the full assignment-options UI surface. After exploring the surface area, I split it: this PR1a is the student-experience half (security-critical, touches Firestore rules + response keying); PR1b will land the new `AssignmentSettingsToggleGroup` shared primitive, the VA `extraSlot` rewrite, and `useVideoActivityAssignments` extension. The type additions in this PR sit dormant until PR1b consumes them.

## Migration note

In-flight VA sessions at deploy time lose progress on first student visit (the new key won't match the old `studentUid`-keyed doc). Assignment + session docs survive. Pre-PR1 archived responses retain their `name` field; Results UI already uses `useAssignmentPseudonyms` for display, so the orphaned field is harmless.

After merge to `main`, `firebase deploy --only firestore:rules` ships the new response-doc rules.

## Test plan

- [x] `pnpm run type-check` clean
- [x] `pnpm run lint` — zero new errors/warnings (pre-existing `remotion/` errors and 2 pre-existing nullish-coalescing warnings in `Results.tsx:512` + `VideoActivityLiveMonitor.tsx:118` unchanged)
- [x] `pnpm run format:check` — only pre-existing untracked `remotion/` files flagged
- [x] `pnpm run test` — all 2046 unit tests pass (regression-free)
- [ ] Manual: SSO student opens an `/activity/:sessionId` link → auto-joins without prompts
- [ ] Manual: anonymous student → enters PIN, picks period from list, joins (no name prompt)
- [ ] Manual: same PIN in two periods → two distinct response docs
- [ ] Manual: existing Quiz flow unchanged (Quiz `joinQuizSession` and `QuizStudentApp` untouched)

## Follow-up — PR1b

- Build new `components/common/library/AssignmentSettingsToggleGroup.tsx` (shared primitive)
- Replace VA `AssignModal` extraSlot with the toggle group + VA-specific block (rewind, penalty, attempt limit, score visibility)
- Extend `useVideoActivityAssignments.createAssignment` to persist new options on assignment + session docs
- Add tests for new options round-trip

### DIFF
--- a/components/videoActivity/VideoActivityStudentApp.tsx
+++ b/components/videoActivity/VideoActivityStudentApp.tsx
@@ -33,23 +33,28 @@ import { QuestionOverlay } from './QuestionOverlay';
 
 /**
  * Resolve the SSO student's class period from the session's
- * `classPeriodByClassId` map. Falls back to the first declared period name
- * when the map is missing (best-effort — the session-create flow populates
- * the map for new assignments). Returns undefined when no period can be
- * resolved; callers default the response doc's `classPeriod` accordingly.
+ * `classPeriodByClassId` map. Mirrors the Quiz semantics: only resolves
+ * when the intersection of the student's claimed `classIds` with the
+ * session's targeted `classIds` is unambiguously a single class. When the
+ * intersection is empty (claim doesn't overlap the session) or multiple
+ * (the student is in 2+ targeted classes), returns `undefined` and lets
+ * the response doc carry no period — wrong-but-confident attribution to
+ * `periodNames[0]` is worse than no attribution because it silently
+ * corrupts results filtering and the export sheet's class-period column.
  */
 function resolveSsoClassPeriod(
   session: VideoActivitySession,
   classIdsClaim: string[]
 ): string | undefined {
   const map = session.classPeriodByClassId;
-  if (map) {
-    for (const classId of classIdsClaim) {
-      const period = map[classId];
-      if (period) return period;
-    }
-  }
-  return session.periodNames?.[0];
+  if (!map) return undefined;
+  const sessionClassIds = new Set(session.classIds ?? []);
+  const matches = classIdsClaim
+    .filter((id) => sessionClassIds.size === 0 || sessionClassIds.has(id))
+    .map((id) => map[id])
+    .filter((period): period is string => typeof period === 'string');
+  if (matches.length !== 1) return undefined;
+  return matches[0];
 }
 
 // ─── Root ──────────────────────────────────────────────────────────────────────
@@ -85,7 +90,7 @@ export const VideoActivityStudentApp: React.FC = () => {
             if (Array.isArray(claimedClassIds)) {
               setSsoClassIds(
                 claimedClassIds.filter(
-                  (id): id is string => typeof id === 'string'
+                  (id): id is string => typeof id === 'string' && id.length > 0
                 )
               );
             }
@@ -194,17 +199,30 @@ const JoinAndPlay: React.FC<JoinAndPlayProps> = ({
   const [lookingUp, setLookingUp] = useState(false);
 
   // SSO auto-join. Mirrors QuizStudentApp's pattern: a single ref guards
-  // against StrictMode double-invoke; on success the student is taken
-  // straight to the player. Local error state because `lookupSession` can
-  // throw before the hook's `joinSession` runs.
+  // against StrictMode double-invoke. The ref is reset whenever
+  // `joinStatus` flips to `'error'` so a transient Firestore hiccup
+  // doesn't strand the student on a permanent loader — the next render
+  // (e.g. a refresh, network recovery) re-arms the effect for another
+  // attempt. Local `ssoAutoJoinError` covers the lookup-failure leg
+  // (`lookupSession` swallows errors and returns `null`); the hook's own
+  // `error`/`joinStatus` carry post-lookup failures.
   const ssoAutoJoinStartedRef = useRef(false);
   const [ssoAutoJoinError, setSsoAutoJoinError] = useState<string | null>(null);
+
+  // Re-arm the started ref when the hook reports an error so a recoverable
+  // failure (e.g. transient network) can retry on the next effect pass.
+  // Done as adjust-state-during-render rather than an effect to avoid an
+  // extra render cycle (per CLAUDE.md "useEffect is an escape hatch").
+  if (joinStatus === 'error' && ssoAutoJoinStartedRef.current) {
+    ssoAutoJoinStartedRef.current = false;
+  }
 
   useEffect(() => {
     if (!isStudentRole || !sessionId) return;
     if (ssoAutoJoinStartedRef.current) return;
     if (joinStatus === 'joined' || joinStatus === 'loading') return;
     ssoAutoJoinStartedRef.current = true;
+    setSsoAutoJoinError(null);
     void (async () => {
       try {
         const sessionInfo = await lookupSession(sessionId);
@@ -223,14 +241,12 @@ const JoinAndPlay: React.FC<JoinAndPlayProps> = ({
         );
       }
     })();
-  }, [
-    isStudentRole,
-    sessionId,
-    ssoClassIds,
-    lookupSession,
-    joinSession,
-    joinStatus,
-  ]);
+    // `joinStatus` is intentionally omitted from deps. The ref guard above
+    // and the early-return on 'joined'/'loading' make a status-driven
+    // re-fire redundant; including it would just trigger a no-op re-render
+    // on every status transition.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isStudentRole, sessionId, ssoClassIds, lookupSession, joinSession]);
 
   // Track answered question IDs for anti-skip enforcement in VideoPlayer
   const answeredQuestionIds = React.useMemo(
@@ -249,15 +265,15 @@ const JoinAndPlay: React.FC<JoinAndPlayProps> = ({
     try {
       const sessionInfo = await lookupSession(sessionId);
       const periodNames = sessionInfo?.periodNames ?? [];
-      // Anon (PIN) joiners always disambiguate by period when the assignment
-      // declares any — `pin-{period}-{pin}` is what keys the response doc.
-      // Show the picker even when there's only one period to keep the doc
-      // key deterministic across re-joins.
-      if (periodNames.length > 0) {
+      // Anon (PIN) joiners disambiguate by period when the assignment
+      // declares more than one — `pin-{period}-{pin}` is what keys the
+      // response doc. With a single period there's nothing for the
+      // student to choose, so auto-pick to skip a friction tap.
+      if (periodNames.length > 1) {
         setPeriodStep(periodNames);
         return;
       }
-      await joinSession(sessionId, pin.trim(), undefined);
+      await joinSession(sessionId, pin.trim(), periodNames[0]);
     } finally {
       setLookingUp(false);
     }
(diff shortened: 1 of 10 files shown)


OPS-PIvers:
## CI status update

Pushed [977fe586](https://github.com/OPS-PIvers/SpartBoard/pull/1539/commits/977fe586) to address the **Code Quality Checks** failure. PR validation enforces `--max-warnings 0` and was failing on two pre-existing `prefer-nullish-coalescing` warnings in `Results.tsx:512` and `VideoActivityLiveMonitor.tsx:118`. Replaced both `||` fallthrough chains with explicit non-empty checks (same semantics, lint-clean).

### E2E `sharing.spec.ts` failure — **not from this PR**

The `Board Sharing › can share and import a board` test failure is unrelated to the changes here. The board-share flow was rewritten this morning in [e1767a55](https://github.com/OPS-PIvers/SpartBoard/commit/e1767a55) — the new `ShareLinkCreatorModal` opens a modal instead of the legacy "click → instant clipboard copy" path. The E2E test still expects the old behavior (waits for a `Link copied` toast that no longer appears immediately).

`tests/e2e/sharing.spec.ts` was last touched on 2026-04-07; the new modal landed 2026-05-07 09:29. `dev-paul` direct-push CI doesn't run E2E, so the breakage didn't surface until a PR ran. This PR is just the first one after the change.

The fix belongs in a separate PR that updates the test to drive the modal (pick mode → click "Create link" → expect the result URL). Happy to do that as a follow-up if you'd like; it's not in scope for VA student-experience parity.

OPS-PIvers:
## Review feedback addressed in [9bdc3458](https://github.com/OPS-PIvers/SpartBoard/pull/1539/commits/9bdc3458)

### Security (firestore.rules)
- **`callerOwnsResponse()` regex too loose** *(Copilot)* — replaced with `isStudentRoleUser()` ternary branching: SSO must equal `auth.uid`; anonymous must match `^pin-[a-z0-9_]+-[a-z0-9_]+$` (mirrors `encodeResponseKeySegment`). Closes the cross-auth path where any caller could grief-create arbitrary `pin-*` docs.
- **`completedAttempts` resettable** *(Gemini critical, Copilot)* — added monotonic-growth check on update (`request.resource.data.get('completedAttempts', 0) >= resource.data.get('completedAttempts', 0)`) and forbade forging non-zero values at create. Same pattern applied to `tabSwitchWarnings`.
- **`is list` guard** *(Gemini)* — added before `answers.size()` / `hasAll(...)` calls.
- **Read rule throws on missing pin-\* doc** *(Copilot)* — added `resource == null` branch so the anon `getDoc()` probe in `joinSession` doesn't deny on first visit. Mirrors Quiz response read rule.
- Switched `affectedKeys()` → `changedKeys()` to match Quiz pattern.

### Correctness
- **`completedAttempts` per-question typing concern** *(Gemini)* — root cause was a typing bug. Matched Quiz: `completedAttempts: number` (counter), not `number[]`. Quiz uses an integer `increment(1)` on each completion; the rule check is `>=`. The prior `number[]` shape was inconsistent.
- **`completedAttempts` not initialized in newResponse** *(Gemini high)* — initialized `completedAttempts: 0` and `tabSwitchWarnings: 0` at response-doc create so the rules' monotonic comparison has a concrete prior value (otherwise `resource.data.completedAttempts` would be missing and the comparison would throw).

### Logging consistency
- **`console.warn` / `console.error`** *(Copilot ×2)* — replaced both with `logError('VideoActivityStudentApp.authInit', err)` and `logError('VideoActivityStudentApp.ssoAutoJoin', err, { sessionId })`, plus the join-hook one (`useVideoActivitySessionStudent.joinSession`). Routes through the structured logger.

### Export
- **SSO rows exporting as "Student"** *(Copilot, low-conf flagged but valid)* — pass `byStudentUid` to `QuizDriveService.exportResultsToSheet` so SSO `studentRole` rows resolve to their ClassLink names.

### Verification
- `pnpm type-check` clean
- `pnpm lint` zero new violations
- All 2046 unit tests still pass

The unrelated E2E `sharing.spec.ts` failure is still expected (board-share modal flow change from [e1767a55](https://github.com/OPS-PIvers/SpartBoard/commit/e1767a55) wasn't paired with an E2E update); separate PR territory per [my earlier comment](https://github.com/OPS-PIvers/SpartBoard/pull/1539#issuecomment-4399474910).

OPS-PIvers:
## Subagent code review — all 19 actionable findings addressed in [a477e906](https://github.com/OPS-PIvers/SpartBoard/pull/1539/commits/a477e906)

Spun up a fresh code-reviewer subagent on the full PR diff and asked it to surface issues at every severity. It returned 22 findings (M3 withdrawn on closer reading; L3 deferred as future-fragile-but-works-today; M3 was the only true non-issue). The remaining 19 are fixed:

### Critical (2)
- **C1 — Live monitor blank for SSO students** ✅ Wired `useAssignmentPseudonyms` into `VideoActivityLiveMonitor`; added `pickDisplayLabel(response, byStudentUid)` that falls through (roster name → `response.name` → `response.pin`); the meta line hides `PIN …` when pin is absent.
- **C2 — SSO auto-join could strand students on permanent loader** ✅ Re-arm `ssoAutoJoinStartedRef` when the hook reports `joinStatus === 'error'` (adjust-state-during-render per CLAUDE.md). Transient Firestore failures now retry on the next render.

### Important (4)
- **I1 — Legacy-key fallback** ✅ `joinSession` now probes a legacy `auth.uid`-keyed doc when the deterministic key is empty; in-flight anon sessions resume against the existing response. SSO unaffected.
- **I2 — SSO `classPeriod` resolution** ✅ Tightened to mirror Quiz: only resolves when `claimedClassIds ∩ session.classIds` has exactly one match; dropped the `periodNames[0]` fallback that silently misattributed students in 2+ classes.
- **I3 — Doc-id-shape regex tests** ✅ Added two negative tests: (a) anon cannot create at arbitrary doc id; (b) SSO cannot create under `pin-*` shape.
- **I4 — Monotonic counter tests** ✅ Added two negative tests: cannot decrement `completedAttempts` or `tabSwitchWarnings`.

### Medium (5)
- **M1** auto-pick period when `periodNames.length === 1` (skip friction tap).
- **M2** stale `responseDocId` comment updated to reflect the new dual-keying scheme.
- **M5** remaining 3 `console.error` listener calls now route through `logError`.
- **M6** added defense-in-depth comment to `firestore.rules` explaining the redundant pin/name immutability checks.
- **M4** guard against the shared encoder collapsing to `'default'` for both segments — surfaces a user-facing error rather than silent collision.

### Low (4)
- **L1** swapped error-screen precedence to `error ?? ssoAutoJoinError` (matches Quiz).
- **L2** removed `joinStatus` from auto-join effect deps with a justification comment.
- **L4** added `id.length > 0` to the SSO `classIds` claim filter.
- **L5** added length cap `{1,64}` to the `pin-*` regex in `firestore.rules` (both Quiz and VA).

### Nitpicks (4)
- **N3** clarified the immutability comment that `studentUid` is checked separately.
- **N4** hedged `BaseSessionOptions` JSDoc to drop the speculative GuidedLearning reference.
- **N5** clarified `attemptLimit` JSDoc as per-attempt-of-activity (a counter), not per-question.
- **N6** documented that `VideoActivityResponse.isCorrect` semantics are deferred to PR1b's publish flow.

### Verification
- `pnpm type-check` clean
- `pnpm lint` clean (only pre-existing `remotion/` noise remains)
- All **2046 unit tests** pass
- 4 new rules tests added covering the security-critical doc-id and monotonic-counter contracts

The single deferred item is **L3** (SSO auto-join doesn't flip the hook's `joinStatus`/`error` directly — works today via the `ssoAutoJoinError` shadow but future-fragile). Calling that out as a follow-up rather than expanding scope here.

OPS-PIvers:
## Second-pass code review — all 5 findings addressed in [ff0eb8e5](https://github.com/OPS-PIvers/SpartBoard/pull/1539/commits/ff0eb8e5)

Spun up a second code-reviewer subagent and asked it to specifically audit the fixes from the first pass under the assumption that recently-changed code is the most likely place for new bugs to hide. It returned 5 actionable findings. Each one is a real catch.

### Critical — C1 fix was incomplete on multi-class topology
`VideoActivityLiveMonitor` and `Results` both wired the **single-class** `useAssignmentPseudonyms` hook. But `session.classId` is a transitional mirror of `classIds[0]` only — so for any assignment targeting 2+ classes, SSO students from `classIds[1+]` weren't resolved. Their rows fell back to `'—'` in the live monitor and to the generic `"Student"` label in the export sheet. This is the exact topology `classPeriodByClassId` exists to enable, so the regression was directly in the path this PR introduces.

✅ Switched both to `useAssignmentPseudonymsMulti` mirroring the [QuizLiveMonitor pattern](https://github.com/OPS-PIvers/SpartBoard/blob/main/components/widgets/QuizWidget/components/QuizLiveMonitor.tsx#L295-L307): `session.classIds ?? (session.classId ? [session.classId] : [])`.

### Important — C2 retry path was a no-op
Sharp catch from the reviewer. The render-time `ssoAutoJoinStartedRef` reset on `joinStatus === 'error'` was paired with deps that never changed (I removed `joinStatus` in the first pass, mistakenly thinking the ref guard made it redundant). After the ref was reset, **nothing triggered the effect to re-run** — recovery required a page refresh. The reviewer pointed out the retry behavior the commit message claimed wasn't actually delivered.

✅ Restored `joinStatus` to the effect deps and dropped the eslint-disable. Guard + reset + dep together now actually deliver the documented retry-on-error behavior.

### Important — M4 guard too narrow
`pin-default-default` only catches the both-segments-default case. The realistic shape is two students with all-special-character PINs in the **same period** collapsing to `pin-{period}-default` — much more reachable since PIN inputs are roster-controlled.

✅ Switched to `encodeResponseKeySegment(studentPin) === 'default'` so the per-student PIN segment is what's checked. Period defaulting alone is harmless because there's only one period bucket per session.

### Low — E2E silently passing on clipboard-mock failure
`sharing.spec.ts` would skip both the URL assertion and the import roundtrip when `clipboardText` was empty. So a modal that rendered an empty/non-`/share/` URL would still pass.

✅ Added `await expect(shareUrlInput).toHaveValue(/\/share\//)` directly on the input — works regardless of clipboard mock status. Falls back to reading the input value to keep the import roundtrip running.

### Low — `resolveSsoClassPeriod` over-rejects on duplicated periods
Rejected unambiguous cases when a student was enrolled in 2 targeted classes that happened to share a period name (e.g. teacher labelled both "Period 1") because `matches.length !== 1`.

✅ Dedupe via `Set` before the size check — only `Set.size !== 1` is genuinely ambiguous.

### Verification
- `pnpm type-check` clean
- `pnpm lint` zero new violations
- All **2046 unit tests** pass

The reviewer's confidence rating was 4/5 with a specific note that they didn't run the test suite or rules emulator themselves — the validation above closes that loop. No remaining findings; the residual notes (lifecycle audit on the I1 listener, period-dedup edge case) were either non-issues or now addressed.

PR #1541, opened 2026-05-07
fix(quiz): expose Question Randomization on create + reposition class picker on edit

OPS-PIvers:
## Summary

Fixes two discrepancies between the quiz **assignment-create** modal and the **edit-in-progress** modal, on top of the new shared `AssignmentSettingsToggleGroup` primitive landed in `dev-paul` (#1540).

### 1. Question Randomization missing on create
The shared toggle-group already renders a "Question Randomization" section, but the create flow (`QuizManager.tsx → AssignExtraSlot`) was passing `excludeSections={['randomization']}` and not threading the shuffle fields through. Result: teachers had no way to set them at assignment creation, and every new assignment used the hardcoded `Widget.tsx` defaults (`shuffleQuestions: false`, `shuffleAnswerOptions: false`).

This PR:
- Adds `shuffleQuestions` and `shuffleAnswerOptions` to `QuizAssignOptions`.
- `buildDefaultAssignOptions()` defaults `shuffleAnswerOptions` to **`true`** so newly created assignments match the edit-modal default and the legacy "always shuffle answer options" behavior; `shuffleQuestions` defaults to `false`.
- Passes both fields into `AssignmentSettingsToggleGroup.options`.
- Removes `excludeSections={['randomization']}` from `AssignExtraSlot`.
- Threads both fields into `sessionOptions` in `handleAssignConfirm()`. Downstream `useQuizAssignments.ts` already reads and persists them.

### 2. Class picker placement on edit
On create, the Class Periods picker sits at the top of `extraSlot` which is intuitive. On edit it was buried inside `plcSlot` next to PLC sharing — confusing and inconsistent. Moved the Class Periods picker out of `plcSlot` and to the top of `extraSlot` so the layout matches the create flow.

## Test plan
- [ ] Create a new quiz assignment → "Question Randomization" section appears with both toggles, defaults match edit modal (Shuffle Questions OFF, Shuffle Answer Options ON)
- [ ] Toggle Shuffle Questions ON, assign → confirm `sessionOptions.shuffleQuestions === true` on the resulting assignment doc
- [ ] Open edit modal on an in-progress assignment → "Class Periods" picker is now the first section under the mode selector (above Quiz Integrity)
- [ ] PLC toggle still works correctly with Class Periods no longer adjacent to it
- [ ] `pnpm run type-check`, `pnpm run lint`, `pnpm run format:check` all pass

https://claude.ai/code/session_01EP5QuGpM2wcyGtZ7J64DzB

### DIFF
--- a/components/widgets/QuizWidget/components/QuizAssignmentSettingsModal.tsx
+++ b/components/widgets/QuizWidget/components/QuizAssignmentSettingsModal.tsx
@@ -237,89 +237,7 @@ export const QuizAssignmentSettingsModal: React.FC<
       confirmLabel="Save"
       onAssign={handleAssign}
       extraSlot={
-        <AssignmentSettingsToggleGroup
-          modeLocked={modeLocked}
-          options={{
-            tabWarningsEnabled: options.tabWarningsEnabled,
-            showResultToStudent: options.showResultToStudent,
-            showCorrectAnswerToStudent: options.showCorrectAnswerToStudent,
-            showCorrectOnBoard: options.showCorrectOnBoard,
-            shuffleQuestions: options.shuffleQuestions,
-            shuffleAnswerOptions: options.shuffleAnswerOptions,
-          }}
-          onOptionsChange={(next) =>
-            // The primitive's internal `update` always emits the full options
-            // object with one field changed, so spreading `next` over `p`
-            // correctly merges without dropping other state. Avoids the
-            // `?? p.X` pattern that's easy to misread as a falsy bug.
-            setOptions((p) => ({ ...p, ...next }))
-          }
-          attemptLimit={options.attemptLimit}
-          onAttemptLimitChange={(v) =>
-            setOptions((p) => ({ ...p, attemptLimit: v }))
-          }
-          trailingSlot={
-            <CollapsibleSection label="Gamification">
-              <ToggleRow
-                compact
-                label="Speed Bonus Points"
-                checked={options.speedBonusEnabled}
-                onChange={(v) =>
-                  setOptions((p) => ({ ...p, speedBonusEnabled: v }))
-                }
-                hint="Up to 50% bonus for fast answers"
-              />
-              <ToggleRow
-                compact
-                label="Streak Bonuses"
-                checked={options.streakBonusEnabled}
-                onChange={(v) =>
-                  setOptions((p) => ({ ...p, streakBonusEnabled: v }))
-                }
-                hint="Multiplier for consecutive correct answers"
-              />
-              <ToggleRow
-                compact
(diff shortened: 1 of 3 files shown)


OPS-PIvers:
**Edit modal (`QuizAssignmentSettingsModal`) is not updated with shuffle settings — read/write gap**

The PR description says the shuffle toggles "only existed in the edit modal." Looking at the actual file, `QuizAssignmentSettingsModal.tsx` does **not** have these toggles — they didn't exist before this PR in either flow. More importantly, after this PR merges, the edit modal is still missing them.

Concretely, `QuizAssignmentSettingsModal.tsx` needs four updates that are not in this diff:

1. **`SettingsOptions` interface** (line 46–62): add `shuffleQuestions: boolean` and `shuffleAnswerOptions: boolean`.

2. **`initialOptionsFor`** (line 64–86): seed both from `a.sessionOptions`:
   ```ts
   shuffleQuestions: opts.shuffleQuestions ?? false,
   shuffleAnswerOptions: opts.shuffleAnswerOptions ?? true,
   ```

3. **`handleAssign` → `sessionOptions`** (line 152–161): include both fields in the patch so edits are persisted:
   ```ts
   shuffleQuestions: options.shuffleQuestions,
   shuffleAnswerOptions: options.shuffleAnswerOptions,
   ```

4. **`extraSlot` JSX** (around line 241, before `<SectionHeader label="Quiz Integrity" />`): add a `<CollapsibleSection label="Question Randomization">` block with the two `<ToggleRow>` controls, mirroring the new section in `AssignExtraSlot`.

Without these changes, a teacher who sets shuffle options at creation can never view or change them when editing the assignment. The setting silently persists at the creation-time value with no UI affordance.

---

OPS-PIvers:
**"Shuffle Questions" hint misleads teachers about mode scope**

The hint for the Shuffle Questions toggle in `AssignExtraSlot` reads:

```
"Each student gets a fresh question order on every attempt (self-paced mode)"
```

The parenthetical `(self-paced mode)` implies this feature only works — or is only relevant — in self-paced mode. But the toggle value flows unconditionally into `sessionOptions.shuffleQuestions` regardless of the selected mode. Teachers choosing teacher-paced or auto-progress who see this hint will likely leave the toggle off assuming it has no effect in their mode.

If question shuffling genuinely has no effect in teacher-paced / auto-progress (because the teacher advances the question and every student sees the same one), the toggle should either be disabled or hidden when those modes are selected. If it does apply in all modes, the hint should drop the mode qualifier:

```
"Each student receives questions in a different randomized order"
```

Either way, the current hint creates a false impression. The same issue applies to "Shuffle Answer Options" if it is also mode-dependent.

---

PR #1555, opened 2026-05-08
fix(quiz,va): close attempt-cap bypass — defense-in-depth + cross-launch ledger + PIN→SSO unification

OPS-PIvers:
## Why this PR

Closes a quiz attempt-cap bypass reported by Jen Ivers: a teacher set a quiz to 1 attempt, students logged in via the SSO student link and opened the quiz from `/my-assignments`, and at least one student was nevertheless able to complete the quiz more than once.

Diagnosis on the offending production session (`02b3bb3b…`, Week 3 quiz, `attemptLimit: 1`) showed five independently-broken enforcement layers. Each phase below closes one cluster:

| Hypothesis | Phase that closes it |
|---|---|
| A — `finalizeAllResponses` zeros the counter | 1 |
| B — Per-session limit, not per-quiz | 2 |
| C — Two-tab race in `completeQuiz` | 1 |
| D — No server-side cap | 1 |
| E — Mixed-auth duplicate (SSO + PIN both create response docs) | 3 |

**The bug we observed in production was almost certainly E** — the offending session has 18 SSO-keyed response docs alongside 5 PIN-keyed docs, in a session that was supposed to be SSO-only. A and C/D were latent for the same teachers; B is a semantic mismatch that comes up every time a teacher relaunches.

## ⚠️ Merge with **standard merge**, not squash

Three reviewable commits, intentionally separated. Each is its own coherent change with a focused commit message.

```
Phase 3  feat(quiz,va): PIN→SSO identity unification via pin_index
Phase 2  feat(quiz,va): cross-launch attempt ledger
Phase 1  fix(quiz,va):  defense-in-depth attempt-cap enforcement
```

When merging into `dev-paul`, **uncheck "squash"** so the three commits land separately. (Default for this repo is squash; this is the deliberate exception for this PR.)

## Phase 1 — Defense-in-depth within a session

- `finalizeAllResponses` now writes `completedAttempts: increment(1)` so a session-end sweep counts as a real completed attempt instead of leaving a stored 0 that the rejoin check happily steps over.
- Join cap check now uses `Math.max(completedAttempts ?? 0, 1)` for any `status === 'completed'` doc — a literal stored 0 can no longer slip through.
- `completeQuiz` and `completeActivity` are now `runTransaction` calls that read-then-write atomically and short-circuit on already-completed (new typed `AlreadySubmittedError` sentinel). Kills two-tab races and double-clicks.
- `firestore.rules` for `quiz_sessions/responses` and `video_activity_sessions/responses` now read the parent session's `attemptLimit` and deny any student-side update that pushes `completedAttempts` past it. Status / `completedAt` transitions from completed → completed are denied (same race protection at the rule layer).
- `QuizStudentApp` and `VideoActivityStudentApp` short-circuit to the submitted screen whenever `completedAttempts >= attemptLimit`, regardless of `status`. Defense in depth — kicks in only if some future bug leaks state past the hook checks.
- VA's join previously had **no attempt-cap enforcement at all**; this phase adds the same join cap + reset logic Quiz had.

## Phase 2 — Cross-launch attempt ledger

Per-session counters reset on every relaunch. New top-level cross-launch ledger keyed by `(content, student)` accumulates a student's completed attempts regardless of how many times the teacher relaunches.

- `/quiz_attempt_ledger/{quizId__studentUid}` and `/video_activity_attempt_ledger/{activityId__studentUid}`. Schema: `QuizAttemptLedger` / `VideoActivityAttemptLedger`. Stores `completedAttempts` (monotonic), `lastAttemptAt`, `lastSessionId`. Non-PII opaque ids only.
- `joinQuizSession` / VA `joinSession` read the ledger before admitting the student. Cap = `max(per-session counter, ledger counter, 1 if completed)`. New "no response yet but ledger says capped" branch handles the SSO student opening the new launch from `/my-assignments`.
- `completeQuiz` / `completeActivity` write the ledger increment **inside the same `runTransaction`** as the response status flip (reads precede writes per Firestore's rules) so concurrency cannot double-bump.
- `removeStudent` (Quiz) extends to a `writeBatch` that atomically deletes the response doc **and** the ledger entry, preserving the existing "free this student up to retake" UX. (VA has no `removeStudent` hook surface; ledger reset can ship later.)
- New rules for both ledger collections. Student create requires `completedAttempts == 1` exactly; student update requires `== old + 1` exactly (no jumping the counter); teacher reset requires `== 0`. Identity fields immutable.

**Identity caveat (resolved by Phase 3):** the ledger keys on `auth.uid`. SSO/`studentRole` users have a stable HMAC pseudonym, so the ledger enforces meaningfully. PIN joiners get a rotating anonymous uid — the Phase 3 bridge upgrades them to the SSO uid space, at which point this ledger covers them too with no further changes here.

## Phase 3 — PIN→SSO identity unification

Closes Hypothesis E. A student rostered into an SSO class can take a quiz once via `/my-assignments` (SSO doc keyed by HMAC pseudonym uid) AND once via `/quiz?code=…` with their PIN (PIN doc keyed `pin-{period}-{pin}`). Two different doc keys, two independent counters. **Production data on the offending session shows this happening live.**

Architecture is constrained by the existing "no PII in Firestore" policy: PINs and sourcedIds live in a Drive-backed roster file. The PIN-joining client cannot resolve identity locally. So:

- New non-PII Firestore sidecar at `/users/{teacherUid}/rosters/{rosterId}/pin_index/{indexKey}`. Maps `(period, pin) → HMAC pseudonym + classId`. Schema: `RosterPinIndexEntry` in `types.ts`. Holds only opaque hashes and ids.
- New `commitRosterPinIndexV1(rosterId, entries)` cloud function. Teacher-only. Reads the roster doc to recover `classlinkClassId`, computes pseudonyms with the same HMAC formula `studentLoginV1` uses, replaces the roster's pin_index in a single batch (writes desired entries + deletes the rest). Idempotent.
- New `pinLoginV1(kind, sessionId|code, pin, period?)` cloud function. Public. Resolves session → walks `rosterIds` → reads `pin_index/{indexKey}` → on hit mints a custom token whose uid matches what `studentLoginV1` would have minted for that student over SSO, plus `studentRole: true` and `classIds: [classId]`. On miss returns `matched: false` so the client falls through to the legacy anonymous PIN flow.
- `useRosters.ts` calls `commitRosterPinIndexV1` after every successful Drive write (create + update). Best-effort; failures log and continue.
- `useQuizSession.ts` and `useVideoActivitySession.ts` join flows: when joiner is anonymous AND has a PIN AND session has `rosterIds`, call `pinLoginV1` first and `signInWithCustomToken` with the result. After the swap, `studentUid` / `isAnonymous` are reassigned and the rest of the join (deterministic key, ledger lookup, SSO classId resolution) runs against the upgraded identity. The PIN-side and SSO-side response docs converge on the same key.
- `firestore.rules`: `pin_index` is admin-write-only (the cloud functions use admin SDK); read by the owning teacher and admins. Students never touch this collection directly.

**Migration:** new rosters get the index built on first save after deploy. Existing rosters get it built next time the teacher opens / saves the roster. Until then, PIN students of unindexed rosters fall back to the legacy anonymous flow — no regression, just no Phase 3 benefit on those rosters yet.

## Test plan

- [x] `pnpm run validate` passes locally (type-check, lint, format-check, 2166/2166 unit tests).
- [ ] CI passes (rules tests cannot run locally — Java 21+ required; CI has the right JDK).
- [ ] Manual on the dev preview URL:
  - [ ] Teacher launches a quiz with `attemptLimit: 1`, student completes via SSO → second open from `/my-assignments` shows the cap-reached UI.
  - [ ] Teacher ends and re-launches the same quiz (different `quiz_sessions` doc, same `quizId`); student opens new launch → cap-reached UI (Phase 2 ledger).
  - [ ] SSO student completes once; same student tries to join again via PIN at `/quiz?code=…` → cap-reached UI (Phase 3 bridge converges them on one response doc).
  - [ ] Two-tab test: SSO student opens quiz in tabs A and B, submits in A, attempts submit in B → second submission no-ops (Phase 1 transaction).
  - [ ] Teacher uses "Remove student" → student can take it once more, then is capped again (Phase 2 ledger reset).
  - [ ] Non-rostered PIN-only session still works (legacy fallback path).

### DIFF
--- a/components/quiz/QuizStudentApp.tsx
+++ b/components/quiz/QuizStudentApp.tsx
@@ -504,8 +504,26 @@ const QuizJoinFlow: React.FC<{ isStudentRole: boolean }> = ({
     return <WaitingRoom session={session} pin={pin} />;
   }
 
-  // Active quiz
-  if (session.status === 'active' && myResponse?.status === 'completed') {
+  // Active quiz — already-completed gate.
+  //
+  // Two paths reach the wait screen instead of the question flow:
+  //   1. The student's response is `status: 'completed'` (normal post-submit).
+  //   2. The student is at or past the session's attempt cap, regardless of
+  //      the response's current status. This is defense in depth: if any
+  //      bug ever resets a capped response back to `'joined'` (e.g. a stale
+  //      counter, a partial finalize, a future refactor that changes the
+  //      reset path), the UI still refuses to render the questions. The
+  //      `completeQuiz` transaction blocks a second submit too, but a
+  //      student briefly seeing a question they can't submit would be
+  //      confusing — short-circuiting at the UI is the cleaner UX.
+  const attemptLimit = session.attemptLimit ?? null;
+  const completedCount = myResponse?.completedAttempts ?? 0;
+  const atCap = attemptLimit !== null && completedCount >= attemptLimit;
+  if (
+    session.status === 'active' &&
+    myResponse &&
+    (myResponse.status === 'completed' || atCap)
+  ) {
     return (
       <QuizSubmittedWaitScreen
         session={session}
--- a/components/videoActivity/VideoActivityStudentApp.tsx
+++ b/components/videoActivity/VideoActivityStudentApp.tsx
@@ -512,8 +512,20 @@ const JoinAndPlay: React.FC<JoinAndPlayProps> = ({
   }
 
   // ── Completion screen ─────────────────────────────────────────────────────
-
-  if (videoEnded || myResponse?.completedAt) {
+  //
+  // Defense-in-depth gate: in addition to the existing video-ended /
+  // completedAt triggers, also short-circuit to the completion screen when
+  // the student is at or past the session's attempt cap, regardless of the
+  // response's `completedAt` state. The join hook resets `completedAt: null`
+  // when a student rejoins under the cap, and throws AttemptLimitReached
+  // when at/over it — so this branch is normally redundant. It exists so a
+  // future bug or stale snapshot that leaves a capped response in
+  // `completedAt: null` state can't leak the question UI back to the
+  // student.
+  const attemptLimit = session?.sessionOptions?.attemptLimit ?? null;
+  const completedCount = myResponse?.completedAttempts ?? 0;
+  const atCap = attemptLimit !== null && completedCount >= attemptLimit;
+  if (videoEnded || myResponse?.completedAt || atCap) {
     const answeredCount = myResponse?.answers.length ?? 0;
     const totalQuestions = session?.questions.length ?? 0;
     // Derive correctness via the shared grader so MA / FIB-variants /
--- a/firestore.rules
+++ b/firestore.rules
@@ -1634,7 +1634,17 @@ service cloud.firestore {
     // (atomic with the response delete) — see Phase 2 in the
     // attempt-cap fix branch.
     match /quiz_attempt_ledger/{ledgerId} {
+      // `resource == null` short-circuit lets the student-app join flow
+      // call getDoc() to probe whether a ledger entry exists before
+      // deciding to create it on first completion — without it, the
+      // rules engine throws a Null-value error on
+      // `resource.data.teacherUid` and the read is denied. Mirrors the
+      // response read rule pattern. The student's pseudonym uid is an
+      // opaque HMAC (computed in `studentLoginV1`), so probing another
+      // student's path is not realistically possible without already
+      // knowing that student's uid.
       allow read: if request.auth != null && (
+        resource == null ||
         isAdmin() ||
         request.auth.uid == resource.data.teacherUid ||
         request.auth.uid == resource.data.studentUid
@@ -1685,7 +1695,10 @@ service cloud.firestore {
     // Video Activity Attempt Ledger — same shape as the Quiz ledger.
     // Doc id: `${activityId}__${studentUid}`.
     match /video_activity_attempt_ledger/{ledgerId} {
+      // See `quiz_attempt_ledger` above for the `resource == null`
+      // rationale — same pattern, same threat model.
       allow read: if request.auth != null && (
+        resource == null ||
         isAdmin() ||
         request.auth.uid == resource.data.teacherUid ||
         request.auth.uid == resource.data.studentUid
--- a/firestore.rules
+++ b/firestore.rules
@@ -1604,6 +1604,104 @@ service cloud.firestore {
       }
     }
 
+    // Quiz Attempt Ledger — top-level, cross-launch attempt counter per
+    // (quiz, student). See `QuizAttemptLedger` in types.ts and
+    // `QUIZ_ATTEMPT_LEDGER_COLLECTION` in `useQuizSession.ts` for the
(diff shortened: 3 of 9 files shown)


OPS-PIvers:
All 11 review comments are already addressed in commit `322eaad6` ("fix(quiz,va): address PR #1555 reviewer comments"), pushed before this latest review pass. Both bots appear to be stateless and re-emitting the same feedback on each push. Verifying each in the current branch:

| # | Reviewer | Comment | Status in `322eaad6` |
|---|----------|---------|---------------------|
| 1 | gemini `types.ts:2453` | Add `orgId` to `RosterPinIndexEntry` | ✅ Added at `types.ts:2463` (`orgId: string` with full docstring at 2455-2462) |
| 2 | gemini `functions/src/index.ts:3727` | Extract `classlinkOrgId` from roster | ✅ At `functions/src/index.ts:3732`, with empty-string default |
| 3 | gemini `functions/src/index.ts:3770` | Include `orgId` in pin_index payload | ✅ Written on every `desired.set(...)` entry |
| 4 | gemini `functions/src/index.ts` | Parallelize `pinLoginV1` probes + use stored orgId | ✅ `Promise.all(rosterIds.map(...))`; `collectionGroup` lookup removed entirely |
| 5 | gemini `hooks/useQuizSession.ts` | Use `@/` path alias | ✅ `import { db, auth, functions } from '@/config/firebase';` (line 33) |
| 6 | gemini `hooks/useRosters.ts` | Use `@/` path alias | ✅ `import { db, functions, isAuthBypass } from '@/config/firebase';` (line 16) |
| 7 | Copilot `firestore.rules:1650` | Quiz ledger read needs `resource == null` guard | ✅ Added at `firestore.rules:1640`, with comment explaining the pattern |
| 8 | Copilot `firestore.rules:1704` | VA ledger same | ✅ Added at `firestore.rules:1701` |
| 9 | Copilot `hooks/useQuizSession.ts` | `AlreadySubmittedError` callers don't suppress | ✅ Replaced `throw new AlreadySubmittedError()` with early `return`; class removed entirely. Now matches `completeActivity`'s idempotent-no-op pattern |
| 10 | Copilot `types.ts` | Typo "Client clients" → "Clients" | ✅ Fixed at `types.ts:2442` |
| 11 | Copilot `functions/src/index.ts` | `pinLoginV1` `collectionGroup` hot path | ✅ Subsumed by #1-4 — the `collectionGroup` lookup was removed; orgId now comes from the index entry directly |

Branch is at:
```
322eaad6  fix(quiz,va): address PR #1555 reviewer comments      ← reviewer feedback
4ae51de7  fix(quiz,va): address PR #1555 review feedback        (earlier round)
e6ee2421  feat(quiz,va): PIN→SSO identity unification           (Phase 3)
8e255647  feat(quiz,va): cross-launch attempt ledger            (Phase 2)
5e8fca96  fix(quiz,va): defense-in-depth attempt-cap enforcement (Phase 1)
```

`pnpm run validate` clean (type-check + lint + format + 2196/2196 unit tests); `functions/` `tsc --noEmit` clean.

PR #1558, opened 2026-05-08
Implement YouTube search, Gemini recommendations, and timeline editor

OPS-PIvers:
This pull request introduces several new shared UI components and enhancements to existing components, focusing on improving the editor experience, modal overlays, and drag-and-drop interactions. The main changes include the addition of generic overlay and workspace components for editors, a reusable sortable list with drag-and-drop, and improvements to modal and library card usability.

**New shared editor components:**

* Added `AIGeneratorOverlay`, a reusable, editor-scoped overlay for AI-powered generators (e.g., quiz, video activity), providing a consistent modal UI for prompts and actions.
* Introduced `EditorWorkspace`, a two-pane modal workspace component built on `EditorModalShell`, enabling sticky context/detail panes and supporting overlays for editor-specific features.

**Drag-and-drop and list improvements:**

* Added `SortableList`, a generic vertical drag-reorder list component using `@dnd-kit/sortable`, allowing easy keyboard and pointer-based reordering for editor lists.

**Modal and overlay enhancements:**

* Enhanced `EditorModalShell` to support inline-editable titles, right-aligned header extras, and increased the default max width for a more flexible editor header and layout. [[1]](diffhunk://#diff-f2f79a2d2ffbd797decef254506d24c64342430ce612f77884756b3f4d7dc035R11-R24) [[2]](diffhunk://#diff-f2f79a2d2ffbd797decef254506d24c64342430ce612f77884756b3f4d7dc035R60-R62) [[3]](diffhunk://#diff-f2f79a2d2ffbd797decef254506d24c64342430ce612f77884756b3f4d7dc035L55-R72) [[4]](diffhunk://#diff-f2f79a2d2ffbd797decef254506d24c64342430ce612f77884756b3f4d7dc035L107-R154)
* Improved `DraggableWindow` to prevent modal overlay events (from portals) from bubbling to the underlying widget, avoiding focus and interaction bugs when modals are open. [[1]](diffhunk://#diff-0d33fd1f9e084a22794f65c762da0467ecb5d266454fa1359d74a6156b9c03dcR575-R599) [[2]](diffhunk://#diff-0d33fd1f9e084a22794f65c762da0467ecb5d266454fa1359d74a6156b9c03dcR1684-R1688)

**Library card action enhancements:**

* Updated `LibraryItemCard` to support a secondary primary action button, styled as a secondary CTA, and improved accessibility by adding `aria-label` to the primary action. [[1]](diffhunk://#diff-5007d16bfe77d73e965ea0736de066ba37b951dc9554d6c9392661a886f1556bR337) [[2]](diffhunk://#diff-5007d16bfe77d73e965ea0736de066ba37b951dc9554d6c9392661a886f1556bR351) [[3]](diffhunk://#diff-5007d16bfe77d73e965ea0736de066ba37b951dc9554d6c9392661a886f1556bL366-R371) [[4]](diffhunk://#diff-5007d16bfe77d73e965ea0736de066ba37b951dc9554d6c9392661a886f1556bR491-R529) [[5]](diffhunk://#diff-5007d16bfe77d73e965ea0736de066ba37b951dc9554d6c9392661a886f1556bR543)


OPS-PIvers:
**`syncedVideoActivityGroups.ts` — TOCTOU: share doc read outside the transaction**

`handleJoinSyncedVideoActivityGroup` fetches the share doc *before* the transaction begins, then uses the resolved `groupId` *inside* the transaction:

```ts
// Outside the transaction — snapshot is already stale by the time the tx runs
const shareSnap = await shareRef.get();
const groupId = shareData?.syncGroupId;

const groupRef = db.collection(SYNCED_COLLECTION).doc(groupId);
return db.runTransaction(async (tx) => {
  const groupSnap = await tx.get(groupRef); // stale groupId from outer read
  ...
```

If the share doc is deleted (or its `syncGroupId` rewritten) between the outer `get()` and the point where the transaction acquires its locks, the function uses the stale `groupId` to add the caller to a synced group they shouldn't be joining. Because Firestore transactions retry on contention, this window is real — not just theoretical.

The two sibling functions in this PR (`plcQuizSyncJoin.ts` and `plcAssignmentSyncJoin.ts`) both correctly read their "lookup" documents **inside** the transaction alongside the group doc:

```ts
// Sibling pattern — correct
return db.runTransaction(async (tx) => {
  const [plcSnap, plcQuizSnap] = await Promise.all([
    tx.get(plcRef),
    tx.get(plcQuizRef), // lookup doc inside tx
  ]);
  ...
```

`handleJoinSyncedVideoActivityGroup` should follow the same pattern:

```ts
export async function handleJoinSyncedVideoActivityGroup(
  db: admin.firestore.Firestore,
  uid: string,
  shareId: string
): Promise<JoinSyncedVideoActivityGroupResponse> {
  const shareRef = db.collection(SHARED_COLLECTION).doc(shareId);

  return db.runTransaction(async (tx) => {
    // Read share + group in same transaction — no TOCTOU window
    const shareSnap = await tx.get(shareRef);
    if (!shareSnap.exists) {
      throw new HttpsError('not-found', 'Shared video activity not found.');
    }
    const shareData = shareSnap.data() as { syncGroupId?: string } | undefined;
    const groupId = shareData?.syncGroupId;
    if (!groupId || typeof groupId !== 'string') {
      throw new HttpsError('failed-precondition', 'This share is not enabled for sync.');
    }

    const groupRef = db.collection(SYNCED_COLLECTION).doc(groupId);
    const groupSnap = await tx.get(groupRef);
    ...
  });
}
```

---

OPS-PIvers:
**`Timeline.tsx` — `window` pointer-event listeners leak if the component unmounts during a marker drag**

`handleMarkerPointerDown` registers `pointermove`, `pointerup`, and `pointercancel` listeners on `window`. They are only removed inside `onUpEvt`, which fires on pointer-up. If the user closes the editor modal while actively dragging a marker — before pointer-up fires — the listeners are never cleaned up:

```ts
const handleMarkerPointerDown = (e: React.PointerEvent<HTMLButtonElement>) => {
  const onMoveEvt = (ev: PointerEvent) => { ... setMarkerDragSec(...) };
  const onUpEvt = (ev: PointerEvent) => {
    window.removeEventListener('pointermove', onMoveEvt); // only path that cleans up
    window.removeEventListener('pointerup', onUpEvt);
    window.removeEventListener('pointercancel', onUpEvt);
    ...
    setMarkerDragSec(null);          // setState on unmounted component
    playerRef.current?.seekTo(...);  // stale ref
    onMarkerDrag?.(q.id, lastSec);   // stale prop closure
  };
  window.addEventListener('pointermove', onMoveEvt);
  window.addEventListener('pointerup', onUpEvt);
  window.addEventListener('pointercancel', onUpEvt);
};
```

When the component unmounts, the teardown `useEffect` cleans up the player and poll interval, but nothing removes these window listeners. After unmount, the next pointer-up fires `onUpEvt` which calls `setMarkerDragSec(null)` on a dead component and invokes stale `onMarkerDrag` / `onSelectQuestion` closures from the already-unmounted parent.

A clean fix is to track the active drag's cleanup function in a ref and call it from the component's unmount effect:

```ts
const activeDragCleanupRef = useRef<(() => void) | null>(null);

// In the teardown useEffect (the one that destroys the YT player):
useEffect(() => {
  ...
  return () => {
    cancelled = true;
    activeDragCleanupRef.current?.();  // cancel any in-flight drag
    activeDragCleanupRef.current = null;
    teardown();
  };
}, [videoId, ...]);

// In handleMarkerPointerDown, assign the cleanup:
activeDragCleanupRef.current = () => {
  window.removeEventListener('pointermove', onMoveEvt);
  window.removeEventListener('pointerup', onUpEvt);
  window.removeEventListener('pointercancel', onUpEvt);
};

// In onUpEvt, clear it after use:
activeDragCleanupRef.current = null;
window.removeEventListener(...);
```

---

OPS-PIvers:
**`useVideoActivityEditorState.ts` — `setState` calls inside a `setState` updater in `addQuestionAtTime`**

`addQuestionAtTime` calls `setTimestampInputs` and `setSelectedId` as side effects *inside* the `setQuestions` updater function:

```ts
const addQuestionAtTime = useCallback((seconds: number) => {
  setQuestions((prev) => {
    // ...compute fresh question...
    setTimestampInputs((tsPrev) => ({ ...tsPrev, [fresh.id]: secondsToMmSs(target) })); // ⚠️ side effect
    setSelectedId(fresh.id);  // ⚠️ side effect
    return sortByTimestamp([...prev, fresh]);
  });
}, []);
```

React's rules require updater functions to be pure — no side effects, including calling other state setters. In Strict Mode (which React 19 dev mode uses), updater functions are deliberately invoked twice to surface exactly this kind of violation. A double invocation here would call `setTimestampInputs` and `setSelectedId` twice per `addQuestionAtTime` call.

The reason the logic is inside the updater is to access the current `questions` array via `prev`. The fix is to take `questions` as a dependency of the callback and move all three setters to the call site:

```ts
const addQuestionAtTime = useCallback(
  (seconds: number) => {
    const used = new Set(questions.map((q) => q.timestamp));
    let target = Math.max(0, Math.floor(seconds));
    while (used.has(target)) target += 1;
    const fresh: VideoActivityQuestion = { ...blankQuestion(), timestamp: target };
    setQuestions((prev) => sortByTimestamp([...prev, fresh]));
    setTimestampInputs((prev) => ({ ...prev, [fresh.id]: secondsToMmSs(target) }));
    setSelectedId(fresh.id);
  },
  [questions]
);
```

The `questions` snapshot used to build `used` is the value at call time, which is correct — `addQuestionAtTime` is only called in response to a user gesture (clicking "Add at…"), so there's no stale-closure risk.

---

OPS-PIvers:
**`AIGeneratorOverlay.tsx` — `role="dialog"` without focus management; Escape handler is unreachable on open**

The overlay declares itself a modal dialog (`role="dialog" aria-modal="true"`) and attaches an Escape handler on the wrapper div:

```tsx
<div
  role="dialog"
  aria-modal="true"
  aria-label={title}
  className="absolute inset-0 z-20 ..."
  onKeyDown={(e) => { if (e.key === 'Escape') onClose(); }}
>
```

`onKeyDown` on a non-interactive `div` only fires when focus is already inside that element's subtree. When the overlay opens — typically triggered by a button press in the workspace below — focus stays on the triggering button, which is now visually obscured beneath the overlay. The user's first Escape press fires on the hidden button, not the overlay, so the handler is never reached.

This also breaks WCAG 2.1 SC 2.1.2 (No Keyboard Trap) and the ARIA authoring practices for modal dialogs, which require focus to move into the dialog when it opens.

Minimal fix — add `tabIndex={-1}` to the wrapper and call `.focus()` in a `useEffect` when `open` becomes `true`:

```tsx
const containerRef = useRef<HTMLDivElement>(null);

useEffect(() => {
  if (open) containerRef.current?.focus();
}, [open]);

<div
  ref={containerRef}
  tabIndex={-1}
  role="dialog"
  aria-modal="true"
  aria-label={title}
  className="absolute inset-0 z-20 ... focus:outline-none"
  onKeyDown={(e) => { if (e.key === 'Escape') onClose(); }}
>
```

Alternatively, add `autoFocus` to the close button so focus jumps there on render:

```tsx
<button autoFocus onClick={onClose} aria-label="Close generator">
  <X className="w-5 h-5" />
</button>
```

Either approach satisfies the focus-on-open requirement and makes the Escape handler reachable without a mouse.

---

PR #1663, opened 2026-05-18
fix(text-widget): drag-select crosses paragraphs + list buttons work on inline-only content

OPS-PIvers:
## Summary

Two long-standing bugs in the TextWidget editor:

### 1. Drag-select clamped to a single paragraph

Chrome's contenteditable drag-selection algorithm anchors to the clicked text node and refuses to extend across block boundaries when the mousedown target is text. (Clicks on the editor's padding worked fine because the anchor falls on the container — exactly the user's reported workaround.)

The prior workaround in `installDragSelectEnhancer` called `setBaseAndExtent` from a bubble-phase mousemove listener, but Chromium re-clamps the selection on each pointer tick as part of mousedown's default action, so the override gets undone.

**The fix:** call `e.preventDefault()` on the initial mousedown, suppressing Chromium's default drag-select entirely, and drive focus, caret placement, and extension manually. Interactive descendants (links, form controls) are excluded so their click defaults still fire.

### 2. Bulleted/numbered list buttons did nothing on inline-only editors

`toggleList` collects blocks from `editor.children`, which excludes text nodes. For an editor whose content is a single typed line (no Enter pressed yet) — i.e. one bare text node at the top level — there are zero collectable blocks and the list command silently no-ops.

The toolbar previously called `needsBlockNormalization` first, but that helper deliberately skips inline-only content to avoid stray line-boxes for cases like `<b>headline</b>`.

**The fix:** add a stronger `ensureTopLevelBlocks` helper that always wraps loose top-level content into a block, and route the toolbar's list button and the `Ctrl+Shift+7/8` shortcut through it. WrittenResponseEditor gets the same fix since it has the same toggleList flow.

## Files

- `utils/contentEditableDragSelect.ts` — rewrite enhancer with `preventDefault` + manual focus/caret management
- `utils/contentEditableBlocks.ts` — add `ensureTopLevelBlocks` helper (shares wrap logic with `normalizeEditorBlocks`)
- `components/widgets/TextWidget/FormattingToolbar.tsx` — route lists through `ensureTopLevelBlocks`
- `components/widgets/TextWidget/Widget.tsx` — same fix for the Ctrl+Shift+7/8 keyboard shortcut
- `components/quiz/WrittenResponseEditor.tsx` — same fix for the `<p>`-tag editor

## Test plan

- [x] Unit tests for `ensureTopLevelBlocks` (7 new cases in `utils/contentEditableBlocks.test.ts`)
- [x] Unit tests for inline-only list toggle in `FormattingToolbar.test.tsx` (2 new cases)
- [x] All 2762 existing unit tests still pass
- [x] `pnpm run lint`, `pnpm run type-check`, `pnpm run format:check` all clean
- [x] Verified drag-select crosses paragraphs in a standalone reproduction page with Playwright (clicking on bare text node, dragging across multiple `<div>` blocks)
- [x] Verified inline-only `<b>Bold text</b>` content gets wrapped in `<ol><li>...</li></ol>` when numbered-list button is clicked
- [ ] Reporter to verify in production: click on any paragraph text, drag up or down across multiple paragraphs — selection should follow the pointer
- [ ] Reporter to verify in production: highlight inline-only text, click bullet/numbered list buttons — list should be created

https://claude.ai/code/session_01ADgNurgXTXgMNsqSj5xTJk

---
_Generated by [Claude Code](https://claude.ai/code/session_01ADgNurgXTXgMNsqSj5xTJk)_

### DIFF
--- a/components/quiz/WrittenResponseEditor.tsx
+++ b/components/quiz/WrittenResponseEditor.tsx
@@ -261,12 +261,8 @@ const WrittenResponseEditorInner: React.FC<
   const handleListToggle = (listTag: 'ul' | 'ol') => {
     if (disabled || !editorRef.current) return;
     editorRef.current.focus();
-    // Always wrap loose top-level content in `<p>` blocks first. The
-    // weaker `needsBlockNormalization` skips inline-only content (to
-    // avoid stray line-boxes for `<b>hi</b>`-style snippets), but
-    // `toggleList` collects from `editor.children` and silently no-ops
-    // when no element children exist. ensureTopLevelBlocks is a no-op
-    // when content already has blocks.
+    // `toggleList` collects from `editor.children` (excludes text
+    // nodes), so any loose top-level text must be wrapped first.
     ensureTopLevelBlocks(editorRef.current, { wrapTag: 'p' });
     toggleList(editorRef.current, listTag, 'p');
     handleInput();
--- a/components/quiz/WrittenResponseEditor.tsx
+++ b/components/quiz/WrittenResponseEditor.tsx
@@ -26,6 +26,7 @@ import {
 } from 'lucide-react';
 import { sanitizeQuizResponse } from '@/utils/security';
 import {
+  ensureTopLevelBlocks,
   needsBlockNormalization,
   normalizeEditorBlocks,
 } from '@/utils/contentEditableBlocks';
@@ -260,13 +261,13 @@ const WrittenResponseEditorInner: React.FC<
   const handleListToggle = (listTag: 'ul' | 'ol') => {
     if (disabled || !editorRef.current) return;
     editorRef.current.focus();
-    // Run normalization first so any in-progress mixed structure
-    // (Chrome's bare-first-line + <div>-wrapped-rest pattern after
-    // typing Enter) becomes uniform `<p>` blocks. Then the list
-    // toggle operates on a clean tree.
-    if (needsBlockNormalization(editorRef.current, { wrapTag: 'p' })) {
-      normalizeEditorBlocks(editorRef.current, { wrapTag: 'p' });
-    }
+    // Always wrap loose top-level content in `<p>` blocks first. The
+    // weaker `needsBlockNormalization` skips inline-only content (to
+    // avoid stray line-boxes for `<b>hi</b>`-style snippets), but
+    // `toggleList` collects from `editor.children` and silently no-ops
+    // when no element children exist. ensureTopLevelBlocks is a no-op
+    // when content already has blocks.
+    ensureTopLevelBlocks(editorRef.current, { wrapTag: 'p' });
     toggleList(editorRef.current, listTag, 'p');
     handleInput();
   };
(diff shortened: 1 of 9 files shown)


OPS-PIvers:
**`contentEditableDragSelect.ts` — `e.preventDefault()` is not gated to Chromium**

The PR description and updated module comment both frame this as a Chromium-specific bug workaround, but the `e.preventDefault()` call in `onMouseDown` fires unconditionally in all browsers.

**Desktop (Firefox / Safari):** Drag-selection across block boundaries in `contenteditable` works natively in those engines, so the `preventDefault` replaces correct built-in behavior with the custom `setBaseAndExtent` path unnecessarily. Any edge case in `caretPositionFromPoint`'s polyfill path or in the manual range placement can now degrade selection in browsers that never had the clamp bug.

**Mobile / touch devices:** On iOS and Android, pointer interaction synthesizes `mousedown` events as part of the tap event chain. Calling `e.preventDefault()` there can suppress:
- The virtual keyboard appearing after a tap-to-focus on iOS Safari (programmatic `.focus()` does not always trigger the keyboard the same way a native-initiated focus does)
- Native long-press selection handles
- Tap-to-place-caret on some Android WebView environments

The manual `editor.focus({ preventScroll: true })` call restores focus, but it does not reliably replicate all the browser affordances that come with a natively-initiated focus-from-tap.

**Suggested fix — gate the `preventDefault` + manual caret placement to Chromium:**

```ts
// Only Chromium re-clamps setBaseAndExtent on each pointer tick;
// Firefox/Safari don't have this bug, and preventDefault on mousedown
// can affect touch-selection affordances in mobile browsers.
const isChromium =
  typeof navigator !== 'undefined' &&
  /Chrome\//.test(navigator.userAgent) &&
  !/Edg\/|OPR\//.test(navigator.userAgent);

if (isChromium) {
  e.preventDefault();
  if (document.activeElement !== editor) {
    editor.focus({ preventScroll: true });
  }
  // ... manual range placement (try/catch block) ...
}
anchor = pos;
isDragging = true;
```

In non-Chromium browsers the `isDragging`/`anchor` state machine still runs, and the `setBaseAndExtent` calls from `onMouseMove` extend the selection correctly — they just don't need to fight against the browser re-clamping them on the next tick, because that bug doesn't exist there.

---

OPS-PIvers:
Good catch — fixed in bb73c9a.

Implemented a `needsPreventDefaultWorkaround()` gate that runs the `preventDefault` + manual focus + manual caret placement only when:
1. UA contains `Chrome/` (covers all Blink forks — Chrome, Edge, Opera, Brave, Arc, Vivaldi — all share the clamp bug), AND
2. `(pointer: fine)` matches (skips touch-primary devices to preserve virtual-keyboard / long-press selection handles).

I broadened the regex past your `!/Edg\/|OPR\//` exclusion because modern Edge/Opera are Chromium-based and exhibit the same bug; happy to narrow it back if you've verified those forks have patched it.

The `anchor` / `isDragging` state machine still runs everywhere so `setBaseAndExtent` on mousemove keeps extending the selection in non-Chromium / touch — it just doesn't fight a non-existent re-clamp there.

Added 4 unit tests covering Chromium desktop, Firefox desktop, Safari desktop, and Chromium mobile (`pointer: coarse`).

---

PR #1702, opened 2026-05-26
Implement Phase 2 features for whiteboard: shapes, text, images, selection, multi-page, undo, export

OPS-PIvers:
This pull request introduces several improvements and new features, primarily focused on enhancing the Drawing Widget's usability and accessibility, as well as improving the SettingsPanel's positioning logic to better handle zoom and pan. Additionally, it updates context and provider mock implementations to support new annotation and migration features.

**Drawing Widget Enhancements:**

* Added a new `PageStrip` component to the Drawing Widget, providing a horizontally scrollable strip of page chips with bbox-only thumbnails, page navigation, add/delete/reorder actions, and an accessible kebab menu for each page. The popup menu is portalled to avoid clipping and is robust to scrolling and resizing.
* Updated the Drawing Widget's settings to support and display new background options (blank, grid, lines, dots), leveraging new background style utilities and migration helpers.

**Settings Panel Improvements:**

* Refactored the `SettingsPanel` to use the actual widget element's bounding rect for positioning, ensuring correct alignment when the dashboard is zoomed or panned. Now uses `useLayoutEffect` and listens to zoom changes for reactive updates. [[1]](diffhunk://#diff-555409ff9b8e992d1bd194b8af2cc4a832d0c088724c1112f7995321fb597316L1-R1) [[2]](diffhunk://#diff-555409ff9b8e992d1bd194b8af2cc4a832d0c088724c1112f7995321fb597316R10) [[3]](diffhunk://#diff-555409ff9b8e992d1bd194b8af2cc4a832d0c088724c1112f7995321fb597316R42-R46) [[4]](diffhunk://#diff-555409ff9b8e992d1bd194b8af2cc4a832d0c088724c1112f7995321fb597316L50-R86) [[5]](diffhunk://#diff-555409ff9b8e992d1bd194b8af2cc4a832d0c088724c1112f7995321fb597316L77-R118)

**Context and Provider Updates:**

* Expanded the mock dashboard context (`StudentContexts.tsx` and `SubsDashboardProvider.tsx`) to include new drawing-related state and no-op handlers, such as `drawingWidgetsMigrating`, annotation object CRUD operations, redo/undo support, and new annotation state fields (`activeTool`, `shapeFill`). [[1]](diffhunk://#diff-36bfa135ee0e09c1538ace333379c01d9be3e020e59a3abf94b448b03d03a59cR327) [[2]](diffhunk://#diff-36bfa135ee0e09c1538ace333379c01d9be3e020e59a3abf94b448b03d03a59cR390-R391) [[3]](diffhunk://#diff-36bfa135ee0e09c1538ace333379c01d9be3e020e59a3abf94b448b03d03a59cR405-R417) [[4]](diffhunk://#diff-f78afcf3feda2c077b8e8b1d817db5b99f098a66e32cf62ac015b9d31a8a1ef1R71-R72) [[5]](diffhunk://#diff-f78afcf3feda2c077b8e8b1d817db5b99f098a66e32cf62ac015b9d31a8a1ef1R223) [[6]](diffhunk://#diff-f78afcf3feda2c077b8e8b1d817db5b99f098a66e32cf62ac015b9d31a8a1ef1R326-R330)

**Documentation and Reference Updates:**

* Updated the widget code pattern reference table to point to the correct file for the "Building-defaults consumption" pattern, clarifying an exception to the standard file naming convention.


OPS-PIvers:
Added changelog entry `2026.05.26.2` — **14 details, 3 overview themes** (Draw widget upgrade + improvement + fixes). Review the copy before merge.

Headline: the full Draw widget Phase 2 — shapes, text, images, select/move/resize/rotate, multi-page, undo/redo, PNG/PDF export, background templates, plus the per-object cloud sync that keeps big boards fast. Followed by today's fixes: settings-panel zoom positioning, Specialist Schedule isPast guard, DE/ES/FR Boards Manager color picker labels, Instructional Routines step badge theme color, and the proportional-layout migration fix.

Commit: 8d69dddbc9b7f9ba6dc36107d94210cd719c4f9b

---

components/widgets/DrawingWidget/useCommandStack.ts
```diff
@@ -0,0 +1,208 @@
...
+  // keeping derived state in sync without an effect. The newer
+  // `react-hooks/refs` lint rule flags this on principle, but the pattern is
+  // intentional and the assignments are idempotent (same input → same ref
+  // value) so the StrictMode double-invoke is safe.
+  const objectsRef = useRef<readonly DrawableObject[]>(objects);
+  // eslint-disable-next-line react-hooks/refs
+  objectsRef.current = objects;
+  const stacksRef = useRef<Record<string, PerPageStack>>(stacks);
+  // eslint-disable-next-line react-hooks/refs
+  stacksRef.current = stacks;
```
OPS-PIvers, line 90: The `// eslint-disable-next-line react-hooks/refs` suppression comments on lines 88 and 91 are for a rule that CLAUDE.md explicitly states is **not active** in this repository's ESLint configuration. The render-body ref assignments here are correct and intentional — they don't need suppression at all. The same unnecessary comment appears in `useSelection.ts` line 151 for `selectedObjectRef.current = selectedObject`.

Remove all three suppression comments. The assignments stand as-is; the suppress lines only add noise and mislead future readers into thinking there's an active lint rule being bypassed.

---

components/widgets/DrawingWidget/Widget.tsx
```diff
@@ -27,55 +111,782 @@ export const DrawingWidget: React.FC<{
...
+    onTransformCommit: handleTransformCommit,
+    onRemoveObject: handleRemoveObject,
+    canvasRef,
+  });
+
+  // Clear selection whenever the user switches off the Select tool. Without
+  // this, selection chrome lingers after the user clicks Pen/Rect/etc.
+  useEffect(() => {
+    if (activeTool !== 'select') clearSelection();
+  }, [activeTool, clearSelection]);
```
OPS-PIvers: This `useEffect` is used to react to an internal state change (`activeTool`), which CLAUDE.md explicitly flags as an anti-pattern: "Do not use [useEffect] to … reset state on prop changes, or chain state updates — these all cause extra render passes."

`activeTool` is set by `setActiveTool()` (called from `handleToolClick`) in this same component. `clearSelection()` should simply be called there directly:

```typescript
const setActiveTool = (tool: ShapeTool) => {
  if (tool !== 'select') clearSelection();
  updateWidget(widget.id, {
    config: { ...config, activeTool: tool } as DrawingConfig,
  });
};
```

The `useEffect` + extra render pass is unnecessary; the event handler already knows the new tool at the moment it fires.

---

components/widgets/DrawingWidget/PageStrip.tsx
```diff
@@ -0,0 +1,459 @@
...
+
+/**
+ * Minimal bbox-only thumbnail. Each object's bounding box renders as a thin
+ * outlined rect, projected from the implicit canvas size (largest extent of
+ * any object) into the chip's dimensions. Deliberately a sketch — full
+ * per-page renders cost too much to draw N pages on every keystroke.
+ */
+const PageThumbnail: React.FC<{ page: DrawingPage }> = ({ page }) => {
+  if (page.objects.length === 0) return null;
+  const bboxes = page.objects.map((obj) => getBoundingBox(obj));
```
OPS-PIvers: `PageThumbnail` reads `page.objects` from the denormalized dashboard-doc cache, but `migrateDrawingToSubcollection` strips that array to `[]` after migration (the objects move to the Firestore subcollection). After any user's first migration run, every page will have `page.objects.length === 0`, so `PageThumbnail` will always return `null` and every thumbnail chip will be permanently blank.

The denormalized cache intentionally does NOT carry objects post-migration (that's the whole point of the subcollection move), so the thumbnail can't cheaply get the live objects here. Options:
1. Accept blank thumbnails post-migration and drop the `null` early-return so at least the chip background is visible.
2. Pass `subcollectionMigrated` down from `Widget.tsx` and render a generic "has content" indicator (e.g. a faint filled rect spanning the chip) when migrated and the object count is unknown.
3. Pass the live subcollection `objects` array as an additional prop and use it when the config flag is set.

---

components/widgets/DrawingWidget/Widget.tsx
```diff
@@ -27,55 +111,782 @@ export const DrawingWidget: React.FC<{
...
+            void subRemoveObject(id);
+          }
+        }
+        // Added or updated: walk next.
+        for (const [id, obj] of nextById) {
+          const prev = prevById.get(id);
+          if (!prev) {
+            void subAddObject(obj);
+          } else if (prev !== obj) {
+            void subUpdateObject(obj);
```
OPS-PIvers: All four subcollection write calls (`subClearObjects`, `subRemoveObject`, `subAddObject`, `subUpdateObject`) are fire-and-forget with no error handling. The comment explains the intent ("Firestore's optimistic listener will surface the new state through the subscription anyway"), but that only holds for the *success* path. On a write failure (network error, permission denial, quota exceeded), the local command-stack state has already advanced but the server hasn't — the next snapshot will silently revert the user's stroke/deletion back to the pre-write state with no feedback.

At minimum, failures should surface an error toast so the user knows their change didn't persist. A simple `.catch` on each write would suffice:

```typescript
subAddObject(obj).catch(() => addToast('Drawing change could not be saved.', 'error'));
```

Alternatively, `writeObjects` could accept an `onError` callback wired to `addToast` in the Widget, keeping the hook decoupled from UI.

---

OPS-PIvers (commented):
Reviewed PR #1702 against the 7 existing unresolved comments — the 4 new findings below are not covered by any of those threads.

---

components/widgets/DrawingWidget/useDrawingObjectsDoc.ts
```diff
@@ -0,0 +1,385 @@
...
+      const cm = activeSubsByContext.get(subKey.contextKey);
+      if (!cm) return;
+      const e = cm.get(subKey.pageId);
+      if (!e) return;
+      e.subscribers.delete(subscriber);
+      e.refs -= 1;
+      // Eviction is deferred to the next subscribe — keeping the listener
+      // around for the LRU window costs only a handful of bytes and lets
+      // back-navigation reuse the cached data.
+    };
```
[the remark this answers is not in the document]
OPS-PIvers: Valid concern — fixed in #1703 (sourced from `dev-paul` because the proxy here blocks direct pushes back to `dev-paul`). The cleanup callback now schedules a microtask after decrementing the refcount: if total refs across the context have dropped to zero (full widget unmount, not just a page switch), it tears down every cached listener and removes the context entry. The microtask delay is exactly the trick you suggested so a back-to-back unmount→remount during a page switch sees the new sub increment a ref before we evaluate the total.

---

utils/migrateDrawingToSubcollection.ts
```diff
@@ -0,0 +1,200 @@
...
+      ops.push({
+        kind: 'object',
+        pageId: page.id,
+        objectId: obj.id,
+        payload: obj,
+      });
+    }
+  }
+
+  for (let i = 0; i < ops.length; i += FIRESTORE_BATCH_OP_LIMIT) {
```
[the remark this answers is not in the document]
OPS-PIvers: Valid — fixed in #1703. The write queue is now split into `objectOps` and `pageOps` and concatenated children-first so any partial-batch failure leaves the parent page-meta docs unwritten. Existing tests still pass (they assert per-batch op counts, which the reorder preserves) and the idempotent retry path is unchanged.

---

components/widgets/DrawingWidget/Widget.tsx
```diff
@@ -44,38 +198,508 @@ export const DrawingWidget: React.FC<{
...
+    pageKey: activePage.id,
+    objects,
+    onObjectsChange: writeObjects,
+  });
+
+  const pageNav = useDrawingPages({
+    config: { ...config, pages, currentPage },
+    updateConfig,
+    onPageRemoved: commandStack.forgetPage,
+  });
```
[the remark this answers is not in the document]
OPS-PIvers: Valid — fixed in #1703. Added a new helper `utils/deleteDrawingPageSubcollection.ts` that lists the page's `/objects` subcollection, batch-deletes every doc in chunks of 450 (same pattern as `clear()` in `useDrawingObjectsDoc`), then deletes the parent page-meta doc last. `useDrawingPages.onPageRemoved` now routes through a new `handlePageRemoved` wrapper that calls `commandStack.forgetPage` (local stack) AND fires the subcollection cleanup when `subcollectionMigrated` is true. Errors surface through a toast (`'Page deleted, but cleanup failed in the background.'`) so the user knows if the orphan cleanup failed even though the local delete went through.

---

components/widgets/DrawingWidget/TextEditorOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+  // immaterial here — the consumers (blur/keydown handlers) fire from
+  // event loops AFTER the render commits.
+  const onCommitRef = useRef(onCommit);
+  const onCancelRef = useRef(onCancel);
+  const objectRef = useRef(object);
+  useEffect(() => {
+    onCommitRef.current = onCommit;
+    onCancelRef.current = onCancel;
+    objectRef.current = object;
+  });
```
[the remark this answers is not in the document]
OPS-PIvers: Partially valid — fixed in #1703. The ref-sync `useEffect` is gone; refs are now assigned during render (`onCommitRef.current = onCommit; onCancelRef.current = onCancel; objectRef.current = object;`) so blur/keydown handlers always see the latest closure without a one-render lag.

One correction on the suggestion: the `react-hooks/refs` rule **IS** active in this repo via `eslint-plugin-react-hooks@7.0.1`'s `recommended` config (CLAUDE.md is stale on that point — verified by running `pnpm run lint`), so each render-body assignment still carries a `// eslint-disable-next-line react-hooks/refs` suppression with an explanatory comment.

---

components/common/SettingsPanel.tsx
```diff
@@ -74,21 +94,28 @@ export const SettingsPanel: React.FC<SettingsPanelProps> = ({
...
+    setPosition({ top, left: Math.max(PANEL_MARGIN, (vw - pw) / 2) });
+  };
+
+  // Recompute on mount and whenever the widget position/size, viewport, or
+  // canvas zoom changes. getBoundingClientRect is not reactive, so any input
+  // that can alter the widget's screen rect must be in this dep list.
+  useLayoutEffect(() => {
+    updatePosition();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [widget.x, widget.y, widget.w, widget.maximized, viewport, zoom]);
```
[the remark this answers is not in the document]
OPS-PIvers: Valid — fixed in #1703. Lifted the function out of the layout effect into a `useCallback` with its real captured deps (`widget.x/y/w/maximized`, `widgetRef`). The layout effect now lists `updatePosition` + `viewport` + `zoom` and has no `exhaustive-deps` suppression.

Note: there's still one suppression on the line that calls `updatePosition()` — `react-hooks/set-state-in-effect`. That rule fires because the callback's whole job is `setPosition()` from inside a layout effect. The DOM-measurement → setState pattern is React's recommended approach for "measure post-layout, paint with the correct position" (you can't compute the rect during render — the DOM doesn't exist yet on first commit), so this one is a justified suppression rather than a code smell. Trade-off documented in the inline comment.

---

components/widgets/SpecialistSchedule/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  nowMinutes: number
+): boolean => {
+  if (isActive) return false;
+  // Intentional `||` (not `??`): an empty-string endTime must also fall back
+  // to startTime, matching how endTime is treated elsewhere in the widget.
+  // parseTime would return -1 for '' anyway, but doing the fallback here
+  // preserves the valid startTime path.
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+  const effectiveMinutes = parseTime(endTime || startTime);
+  if (effectiveMinutes < 0) return false; // unparseable — do not flag as past
```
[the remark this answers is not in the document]
OPS-PIvers: Valid — fixed in #1703. The suppression is gone; an explicit guard now treats `undefined` and `''` as absent before falling through to `startTime`:

```ts
const effective = endTime !== undefined && endTime !== '' ? endTime : startTime;
const effectiveMinutes = parseTime(effective);
```

Behavior matches before (empty-string endTime falls through to startTime, parseTime still returns -1 for the unparseable path) without disabling the lint rule.

---

components/widgets/DrawingWidget/TextEditorOverlay.tsx
```diff
@@ -0,0 +1,192 @@
...
+  // editor) doesn't double-fire onCommit and clobber an upstream state reset.
+  const finalizedRef = useRef(false);
+
+  // Translate object coords (canvas-internal px) to CSS px on the page.
+  const scaleX = canvasSize.width > 0 ? canvasRect.width / canvasSize.width : 1;
+  const scaleY =
+    canvasSize.height > 0 ? canvasRect.height / canvasSize.height : 1;
+  // Use page-level coords so the overlay sits over the canvas no matter
+  // where the parent container is scrolled. canvasRect already accounts for
+  // page scroll via getBoundingClientRect + scroll offsets at call site.
```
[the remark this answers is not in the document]
OPS-PIvers: Valid — fixed in #1703. Both the `canvasRect` JSDoc and the inline comment near the position calculation now say the rect is viewport-relative (raw `getBoundingClientRect()` output) and explicitly warn against adding scroll offsets, since the overlay is `position: fixed`. Behavior wasn't wrong; the comment was just misleading future readers.

---

components/widgets/DrawingWidget/useCommandStack.ts
OPS-PIvers: Going to leave the suppressions in place — verified by running `pnpm run lint` that the `react-hooks/refs` rule IS active in this repo. It comes in via `reactHooksPlugin.configs.recommended.rules` (eslint-plugin-react-hooks v7.0.1), and `Object.keys(p.configs.recommended.rules)` confirms `react-hooks/refs` is registered. Removing the suppressions produces hard lint errors (`Error: Cannot access refs during render`) on every render-body ref assignment.

The CLAUDE.md note about the rule being inactive looks stale. I cleaned up the surrounding comments in #1703 so they describe what's actually happening (intentional render-body assignment, suppressing the overly-strict rule because the pattern is idempotent), but kept the `// eslint-disable-next-line` comments themselves.

---

components/widgets/DrawingWidget/Widget.tsx
OPS-PIvers: Valid — fixed in #1703. The `useEffect` is gone; `setActiveTool` now clears selection synchronously in the handler:

```ts
const setActiveTool = (tool: ShapeTool) => {
  if (tool !== 'select') clearSelection();
  updateWidget(widget.id, {
    config: { ...config, activeTool: tool } as DrawingConfig,
  });
};
```

Saves the extra render pass, matches the CLAUDE.md "useEffect is an escape hatch" guidance, and `handleToolClick` already routes through `setActiveTool` so every tool-switch entry point is covered.

---

components/widgets/DrawingWidget/PageStrip.tsx
OPS-PIvers: Valid — fixed in #1703 with a blend of options 2 and 3. `PageThumbnail` now takes two extra props:

- `liveObjects?: DrawableObject[]` — passed only for the **active** page (the Widget has the live subcollection slice in scope and threads it in), so the active page's thumbnail draws an accurate bbox sketch even post-migration.
- `subcollectionMigrated?: boolean` — when true and the denormalized `objects` is empty, render a neutral filled rect (`absolute inset-1 rounded-sm bg-white/10`) instead of returning `null`. The chip no longer looks empty for pages that probably have content we just haven't loaded.

Pre-migration empty pages still return `null` (they really are empty). The cheaper alternative — N extra Firestore reads on popover open — wasn't justified for a sketch indicator.

---

components/widgets/DrawingWidget/Widget.tsx
OPS-PIvers: Valid — fixed in #1703. Added an `onSubWriteError` helper that calls `logError('DrawingWidget.subcollectionWrite', err, { widgetId })` and `addToast('Drawing change could not be saved.', 'error')`. All four call sites now thread through it:

```ts
subClearObjects().catch(onSubWriteError);
subRemoveObject(id).catch(onSubWriteError);
subAddObject(obj).catch(onSubWriteError);
subUpdateObject(obj).catch(onSubWriteError);
```

Hook stays decoupled from UI (no `addToast` import inside `useDrawingObjectsDoc`), the Widget owns the toast surface, and the local command-stack state still advances optimistically — the toast just makes the upcoming revert visible instead of silent.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-05-27

### Summary
Large Phase-2 Drawing Widget PR: new `PageStrip` component (page navigation, kebab menu via portal), background style options (blank/grid/lines/dots), `SettingsPanel` rect-relative positioning with zoom/pan support, plus expanded mock context providers for annotation CRUD/undo-redo. Several follow-up fixes (subscription leak cleanup, batch-write order, page delete subcollection cleanup, etc.) tracked and replied to in #1703.

### Code Quality
Generally clean. Substantial set of prior review comments from gemini-code-assist + copilot have been addressed by the author in #1703 with detailed replies. Two threads remain open with author replies — both reference #1703 fixes (subcollection write order, ESLint suppression on intentional empty-string fallback) and look resolvable once the human reviewer agrees.

### Widget Standards
✓ DrawingWidget's existing scaling/coordinate model (transform-based, `skipScaling: false`) is preserved. PageStrip is a settings/back-face UI primitive that doesn't need container-query scaling.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
- Subscription leak cleanup, fire-and-forget write error handling, batch write ordering, and page-delete subcollection cleanup all addressed per the author's replies referencing #1703. Worth re-verifying that the microtask-based unmount detection in `useDrawingObjectsDoc.ts` correctly handles the active-page→empty→active-page-switch case in practice (vs. just at the unmount).
- Make sure `firestore.rules` covers the `/objects` subcollection writes correctly. Not visible in this diff — flag for confirmation.

### Regression Risk
Medium-high given scope. The PR touches:
- `useDrawingObjectsDoc.ts` — multi-page subscription cache with refcount and microtask cleanup. The microtask trick is subtle; reviewers should walk through page-switch and full-unmount scenarios deliberately.
- `migrateDrawingToSubcollection.ts` — write order matters for pseudo-atomicity guarantee; verify the children-first ordering is preserved across all code paths.
- `PageThumbnail` — post-migration the denormalized `page.objects` is `[]`, so the thumbnail can only sketch the active page's content. Acceptable trade-off but worth documenting in code.
- `SettingsPanel` positioning — now uses `getBoundingClientRect()` + `useLayoutEffect`; verify behavior on rapid zoom transitions and when widget is unmounted mid-measurement.

### Missing Tests
- No tests in the diff for the new PageStrip kebab menu interactions (portal positioning, click-outside, keyboard navigation).
- Page-delete cleanup pseudo-atomicity isn't testable easily, but the `migrateDrawingToSubcollection` write order should be — verify existing tests cover the new children-first ordering.
- The `SettingsPanel` zoom-aware positioning could benefit from a regression test.

### Overall
Substantively ready, but the size and the dependency on follow-up fixes in #1703 mean the merge plan should be: land #1703 first (or fold it in here), then re-confirm the two still-open threads are addressed, then merge. Recommend a final human review of the multi-page subscription refcounting and the new Firestore subcollection cleanup paths before merging into main.

---

---

PR #1718, opened 2026-05-27
SmartNotebook editor authoring + randomizer/stations sizing fixes

OPS-PIvers:
## Summary

Substantial set of authoring improvements to the SmartNotebook widget plus a regression fix to the randomizer and a sweep on Stations.

### SmartNotebook editor (most of the diff)

- **Object→page links**: jump-to-link FAB (arrow) next to the manage-link FAB on linked objects, Ctrl/Cmd+click follows a link directly, "Go to page N" button in the LinkTargetPicker, and the picker no longer dismisses itself on open.
- **Copy/paste**:
  - In-app cross-page copy with link-preserving paste (Cmd+C / Cmd+V) — pasted clones get fresh ids and their hotspot links auto-recreated for the new page.
  - OS clipboard paste — image (data-URL embed) or text → SVG `<image>` / `<text>` objects at page center.
  - In-app clipboard wins over OS clipboard on Cmd+V so a stale screenshot can't stomp a notebook copy.
  - Cmd+D in-page duplicate now also carries links forward.
- **New authoring tools**: text drop, rectangle, ellipse, line, arrow (each as a real `<g data-edit-id>` that joins the existing select/move/resize/rotate/delete/link pipeline).
- **Page background picker**: solid color (8-swatch palette + custom), pattern overlay (lines/grid/dots), or uploaded image — embedded as a data URL so the page SVG stays self-contained.
- **Layer ordering**: Cmd+] / Cmd+[ / Cmd+Shift+] / Cmd+Shift+[ plus a 4-button on-screen group in the toolbar.
- **Toolbar refactor**: collapsed 9 inline tools + 5 colors + 3 width buttons into 6 top-level tools with per-tool popovers; shapes live behind a single Shapes button; opaque slate popover; continuous-range size sliders for pen/highlighter/shapes/text/eraser (4–80 px / 12–96 pt depending on tool); white added to the palette.
- **Undo/redo**: Cmd+Z / Cmd+Shift+Z / Cmd+Y, per-page stacks (cap 50), captures the as-loaded page as the first undo target.
- **Architectural cleanup**: PageEditor is now strictly write-only after mount — host autosave can churn freely without ever clobbering live DOM state. Killed a class of stale-state and mid-drag bugs.
- **Text-tool polish**: blur-then-pointerdown no longer drops a second text; clicking an existing text in text-tool mode opens it for editing; focus deferred via rAF so initial typing lands reliably.
- **Eraser**: now removes anything under the cursor (was ink-only), with the new size slider driving a multi-sample hit-radius.

### Randomizer + Stations

- Restored `clamp(min, ideal, max)` floors that PR #1626's `min()` sweep had erased — chips, buttons, icons no longer collapse on small widgets.
- Result-text font now sizes to the displayed string, with three regimes (spinning / settled / placeholder) so short winners like "1." fill the widget instead of being pinned to the longest roster name. Wider Lexend-bold char ratio (1.13) baked in.
- Same `min() → clamp()` sweep applied to StationsWidget's Shuffle / Rotate / Reset controls.

### Misc

- `.claude/launch.json` vite-dev → `autoPort: true` so worktree previews don't collide with the main dev server on :3000.

## Merge note

⚠️ **Use a regular merge commit (NOT squash)** — per the dev-paul → main convention.

## Test plan

- [ ] SmartNotebook: open a notebook, drop new text/shapes/images via paste, verify edits persist after autosave round-trip
- [ ] SmartNotebook: Cmd+Z undoes the last edit; Cmd+Shift+Z redoes
- [ ] SmartNotebook: Delete key on a selected object removes it on the first press (not the second)
- [ ] SmartNotebook: Cmd+C on a linked object → navigate → Cmd+V preserves the link
- [ ] SmartNotebook: paste an OS screenshot with an empty in-app clipboard → image lands on the page
- [ ] RandomWidget: short result strings ("1.") fill the widget; long roster names still fit
- [ ] StationsWidget: Shuffle / Rotate buttons keep their floor sizes on small widgets


OPS-PIvers:
Added changelog entry — 20 details, 4 overview themes (Notebook editor features, Notebook editor improvements, Specialist Schedule improvement, fixes). Review the copy before merge.

Commit: https://github.com/OPS-PIvers/SpartBoard/commit/be83a9a393e6f47e64bababeb40b0a2c773a4832

---

OPS-PIvers:
## Code Review — PR #1718

No existing comments on this PR. Posting findings after a full diff review.

---

### 🟡 MEDIUM — Container query scaling violation in toolbar icons

**File:** `components/widgets/SmartNotebook/components/PageEditorOverlay.tsx`

The toolbar button containers correctly use `cqmin` sizing (`width: 'min(36px, 8cqmin)'`), but the icons inside them use fixed-pixel Tailwind classes:

```tsx
// renderToolButton
<Icon className="h-4 w-4" />   // 16px fixed

// shape sub-tool buttons inside the shapes popover
<Icon className="h-5 w-5" />   // 20px fixed
```

`SmartNotebook` has `skipScaling: true`, so per CLAUDE.md all front-face content (including icons) must use `cqmin` units. On a large widget the button containers scale up but the icons stay 16/20px and appear tiny inside them. These should use `style={{ width: 'min(16px, 4cqmin)', height: 'min(16px, 4cqmin)' }}` (or similar) instead of Tailwind size classes.

---

### 🟡 MEDIUM — In-app clipboard permanently shadows OS paste

**File:** `components/widgets/SmartNotebook/components/PageEditor.tsx`

`pageEditorClipboard` is module-scoped and is never cleared once populated:

```ts
let pageEditorClipboard: ClipboardEntry[] = [];
// ...
if (pageEditorClipboard.length > 0) {
  e.preventDefault();
  pasteFromClipboard();
  return;
}
```

Once a teacher presses Cmd+C on any notebook object, `pageEditorClipboard` stays populated for the entire page session. Any subsequent document-level paste event (that isn't a native `<input>`/`<textarea>`/`contenteditable`) will be intercepted and paste the stale notebook objects instead of whatever the teacher just copied from the OS. Concrete scenario: teacher copies an SVG arrow → switches tasks → copies a screenshot from another app → returns to the notebook and tries to paste the screenshot → gets the old arrow instead.

The guard that exits early for native inputs only covers `INPUT`, `TEXTAREA`, and `contenteditable` targets, not arbitrary UI. The clipboard should be cleared (set to `[]`) when the editor is unfocused/blurred, or the paste handler should only intercept when the notebook editor is the active focus target.

---

### 🔵 LOW — SVG pattern ID collision when two notebooks are open

**File:** `components/widgets/SmartNotebook/components/PageEditorOverlay.tsx` — `PatternPreview` component

```tsx
const id = `bg-preview-${pattern}`;  // → 'bg-preview-lines', etc.
```

SVG `<pattern>` IDs are document-scoped (not scoped per `<svg>` element). If two SmartNotebook widgets are both open in edit mode with the background picker visible, they both inject `<pattern id="bg-preview-lines">` into the document. The browser uses the first matching ID, so one picker's swatches render the wrong pattern. Fix: append a stable unique suffix — `useId()` from React 18 works well here.

---

### 🔵 LOW — `key` prop on `renderToolButton` return value is a no-op

**File:** `components/widgets/SmartNotebook/components/PageEditorOverlay.tsx`

```tsx
const renderToolButton = (keyId: ToolButtonKey, ...) => (
  <button
    key={keyId}   // ← key on a non-list element; React ignores this
    ...
  >
```

`key` only has meaning when React reconciles a list (`.map()` etc.). `renderToolButton` is called as a plain function and its return is placed into JSX outside any array context, so React silently ignores the `key`. It reads as intentional and could mislead future editors. The prop should be removed.

---

### 🔵 LOW — Variable declaration inserted between import statements

**File:** `components/widgets/random/RandomWidget.tsx`

```ts
import { RandomFlash, RANDOM_FLASH_PLACEHOLDER } from './RandomFlash';

const PLACEHOLDER_LENGTH = RANDOM_FLASH_PLACEHOLDER.length;  // ← between imports

import { RandomGroups } from './RandomGroups';
```

Hoisting makes this valid JS/TS, but it violates the conventional import-block style and will trip up anyone skimming the import section. The constant should be moved to after all imports.

---

### 🔵 LOW — `BackgroundPicker` dialog declares `aria-modal` without a focus trap

**File:** `components/widgets/SmartNotebook/components/PageEditorOverlay.tsx`

```tsx
<div
  role="dialog"
  aria-modal="true"
  aria-label="Page background"
  ...
>
```

`aria-modal="true"` tells screen readers to treat content behind the dialog as inert, but no JavaScript focus-trap is in place. A keyboard user can Tab past the dialog boundary into the visually obscured background content. Per WCAG 2.4.3 and ARIA APG, modal dialogs must confine keyboard focus. The existing `useClickOutside` handles click-outside dismissal but there is no equivalent for keyboard navigation.

---

components/admin/WorkSymbolsConfigurationModal.tsx
```diff
@@ -147,10 +156,10 @@ export const WorkSymbolsConfigurationModal: React.FC<
...
   // --- Save / Close ---
-  const handleSave = () => {
+  const handleSave = async () => {
     setIsSaving(true);
     try {
-      onSave({
+      await onSave({
         config: globalConfig as unknown as Record<string, unknown>,
       });
       uploadedThisSessionRef.current.clear();
```
OPS-PIvers: **`handleSave` toasts "saved" but nothing is persisted.**

`handleSave` now `await onSave(...)` then toasts `"Work Symbols configuration saved"` and calls `uploadedThisSessionRef.current.clear()`. But the caller in `FeaturePermissionsManager.tsx:926` wires `onSave={(updates) => updatePermission('work-symbols', updates)}` — and `updatePermission` (line 200-209) returns `void` after only mutating local Map state. The real Firestore write happens in `savePermission` behind the OUTER "Save Changes" button.

Failure scenario: admin uploads 3 large symbol PNGs, clicks Save inside the modal, sees the success toast, closes the modal, and reloads/navigates before clicking the outer "Save Changes". Result: (a) the 3 Storage objects are orphaned because `uploadedThisSessionRef.clear()` defeated `handleClose`'s deleteFile cleanup, and (b) the permission record never received the new symbols. Pre-PR `handleSave` was synchronous and didn't toast "saved", so this misleading UI is new.

Either `onSave` should actually persist (and the outer Save button becomes idempotent), or the modal should not toast success / clear `uploadedThisSessionRef` until the outer save is confirmed.

OPS-PIvers:
**`components/widgets/SmartNotebook/components/PageEditorOverlay.tsx:399`**

**Notebook-switch while both notebooks are at page 0 leaves stale DOM in the editor — next edit corrupts the new notebook.**

`<PageEditor key={currentPage} svg={svgToEdit} … />` keys only on `currentPage`. PageEditor.tsx is now "write-only after mount" — its mount effect (line ~794-805, empty deps with `eslint-disable react-hooks/exhaustive-deps`) reads `svg` exactly once and intentionally ignores subsequent changes. `Widget.tsx:144-160` and `handleSelect:350-353` reset `currentPage` to 0 on notebook switch — but if it's already 0, the value doesn't change and the React key doesn't change either, so PageEditor does not remount.

Repro: open notebook A on page 0, make an edit, switch to notebook B (also page 0). The PageEditor's DOM still shows A's edited content; the host believes the user is editing B. The user's next pointerup serializes A's content into `editedSvgsRef.current` keyed under B's `currentPage` — autosave then writes A's content into B's page 0.

Fix: include the notebook identity in the key, e.g. `key={`${activeNotebookId}:${currentPage}`}`.

components/widgets/SmartNotebook/Widget.tsx
```diff
@@ -505,6 +505,26 @@ export const SmartNotebookWidget: React.FC<{
...
 
+  // Persist many new object→page links in one Firestore write — used after
+  // paste / duplicate of linked objects so a clipboard with five hotspots
+  // doesn't fan out to five sequential round-trips. New ids are assumed
+  // to be unique (caller generates them), so we just append.
+  const handleSaveObjectLinksBatch = async (
+    links: NotebookObjectLink[]
+  ): Promise<void> => {
+    if (!user || !activeNotebook || links.length === 0) return;
+    const next = [...(activeNotebook.objectLinks ?? []), ...links];
```
OPS-PIvers: **Read-modify-write race on `objectLinks` — concurrent paste/duplicate batches lose links.**

`handleSaveObjectLinksBatch` reads `activeNotebook.objectLinks ?? []` from the closure-captured React-state snapshot and writes the appended array with a plain `updateDoc({ objectLinks: next })` — no `arrayUnion`, no transaction. Same pattern in `handleSaveObjectLink` (line ~489) and `handleRemoveObjectLink` (line ~528).

`activeNotebook` is a `useMemo` over `notebooks` (Widget.tsx:116-119), populated only by `onSnapshot`. Two batch saves dispatched before Firestore round-trips back will both read the same baseline; the second `updateDoc` clobbers the first batch's new links.

Failure scenario: teacher pastes 3 linked objects (batch A writes baseline + [X1,X2,X3]) and within ~300ms pastes 2 more (batch B reads the *same* stale `activeNotebook.objectLinks` because the snapshot hasn't echoed yet and writes baseline + [Y1,Y2]). X1-X3 silently disappear from Firestore — the SVG still contains those objectIds, but their `NotebookObjectLink` rows are gone, so on reload the hotspots are dead.

Suggested fix: switch the batch and single-add paths to `arrayUnion(...newLinks)`; for remove, do a transactional read-modify-write or filter inside a runTransaction.

components/widgets/SmartNotebook/components/PageEditor.tsx
```diff
@@ -595,23 +1292,119 @@ export const PageEditor: React.FC<PageEditorProps> = ({
...
+        }
+        const text = data.getData('text/plain');
+        if (text) {
+          e.preventDefault();
+          pasteTextString(text);
+          return;
+        }
+      }
+    };
+    window.addEventListener('paste', onPaste);
```
OPS-PIvers: **`window.addEventListener('paste', onPaste)` is bubble-phase — Dock's identical bubble listener (Dock.tsx:517) is registered first and ALSO fires.**

Dock's paste handler (registered at app mount, runs first in bubble order) never calls `e.preventDefault()` / `e.stopPropagation()` — its `e.defaultPrevented` guard catches the SmartNotebook case only if PageEditor marks the event *before* Dock runs, which it can't from bubble phase. Result: Cmd+V of an image inside the notebook (a) opens Dock's board-level image-paste modal AND (b) inserts the same image into the notebook page.

Text-paste with a URL/sharelink also dual-fires: Dock detects the widget type and queues it while PageEditor drops the same string as an SVG `<text>`.

Fix: register PageEditor's listener with capture phase (`addEventListener('paste', onPaste, true)`) and call `e.stopImmediatePropagation()` (or at least `preventDefault()` early so Dock's `defaultPrevented` guard catches it). Matches the pattern already used at `StickerBookWidget.tsx:346`, `AnnotationOverlay.tsx:312`, `StickerLibraryModal.tsx:149`.

components/widgets/SmartNotebook/components/PageEditor.tsx
```diff
@@ -554,16 +857,410 @@ export const PageEditor: React.FC<PageEditorProps> = ({
...
+    };
+  }, [imperativeApiRef, reorderSelected, setBackground]);
 
   useEffect(() => {
     const onKey = (e: KeyboardEvent) => {
       if (editing) return;
+      // Undo / Redo land in their own branch so they fire whether or not
+      // anything is selected — the prior emit might've been a paste with
+      // selection still active, or a tool-mode action with no selection.
+      if (
```
OPS-PIvers, line 1270: **Window-capture keydown listener intercepts Cmd/Ctrl+Z, +Y, +C, +[, +] from any input on the page — there's no `e.target instanceof HTMLInputElement|HTMLTextAreaElement|HTMLElement(isContentEditable)` check.**

The only guard is `if (editing) return` (line ~1314), which only covers the editor's *own* inline textarea. Any other text field on the page (sidebar dashboard-name input, settings panels, other widgets) loses native undo/redo/copy to the notebook stack.

Failure scenarios:
1. Teacher clicks into the dashboard-rename input, types a typo, presses Cmd+Z to undo → capture-phase handler swallows the event and replays a notebook SVG instead. The typed character stays.
2. Teacher has selected an SVG object, then clicks into a roster input and selects text + Cmd+C → the OS clipboard is never updated; the notebook's `pageEditorClipboard` is overwritten with the SVG instead.

Fix: gate every modifier-key branch on `if (e.target && isEditableTarget(e.target)) return;` (matching what the paste handler at line ~1225 already does — only that one has the guard).

OPS-PIvers:
**`components/admin/WorkSymbolsConfigurationModal.tsx:136`**

**Three mutators still use the stale-closure pattern that `handleFiles` was explicitly fixed to avoid — concurrent upload completing between user click and re-render silently drops the new symbols.**

`handleFiles` (line ~108) was correctly converted to `setSymbols((prev) => [...prev, ...newSymbols])`. But `removeSymbol` (line 136), `updateSymbolTitle` (line ~140), `toggleBuilding` (line ~146) all still call `setSymbols(globalConfig.symbols.filter/map(...))` derived from closure state. The inline "switch from all to explicit" handler (line ~328-334) has the same shape.

Failure scenario: admin starts a 10-PNG upload (multi-second), then edits a previously-saved symbol's title while uploads run. The first uploads have already committed A..E via the functional updater; `updateSymbolTitle`'s closure captured `globalConfig.symbols` before A..E landed. `setSymbols(globalConfig.symbols.map(...))` overwrites with the pre-upload array, dropping A..E without warning.

Fix: convert all four to functional updaters, identical to the `handleFiles` fix.

OPS-PIvers:
**`components/widgets/SmartNotebook/components/PageEditor.tsx:78`**

**Module-scoped `let pageEditorClipboard: ClipboardEntry[] = []` leaks across PageEditor instances, across notebook switches, and across user sessions in the same tab.**

Two SmartNotebook widgets on the same dashboard share one clipboard. Two users on a shared classroom laptop share it (no sign-out clears it — only a page reload). The in-app-wins precedence over OS clipboard makes the leak observable: Teacher A's copied object will paste into Teacher B's notebook next time Teacher B presses Cmd+V, even after sign-out/sign-in within the SPA.

Combined with the unclamped `onFollowLink`, a pasted object can also carry a stale `targetPage` that doesn't exist in the destination notebook, producing a dead hotspot.

Fix: scope the clipboard per-widget — store it on a ref inside the PageEditor (or on a per-notebook key in module state keyed by `activeNotebookId`), and clear it on auth-state-change.

components/widgets/SmartNotebook/components/PageEditor.tsx
```diff
@@ -390,39 +530,129 @@ export const PageEditor: React.FC<PageEditorProps> = ({
...
+    if (!svgRef.current) return;
+    const serialized = exportEditedSvg(svgRef.current);
+    // Snapshot the baseline that was in effect BEFORE this edit so the
+    // user can Cmd+Z back to it. The very first emit uses the initial
+    // prop string (page-as-loaded) since lastEmittedSvgRef hasn't been
+    // populated by a previous emit yet. Identical strings are skipped
+    // so a no-op pointerup (e.g. drag below the threshold) doesn't fill
+    // the stack with duplicates.
+    const previousBaseline = lastEmittedSvgRef.current ?? initialSvgRef.current;
+    if (previousBaseline !== null && previousBaseline !== serialized) {
```
OPS-PIvers, line 599: **Page-background images are embedded as inline data URLs into the page SVG, then every subsequent `emitChange` snapshots the entire SVG (including the data URL) onto `pastSvgStackRef` — the 50-deep undo stack can balloon into hundreds of MB.**

`setBackground({ kind: 'image', dataUrl })` inlines `<image href="data:image/...;base64,...">` into the page SVG (so the page stays self-contained). The undo machinery treats every emitChange as "previous baseline" and pushes the whole serialized SVG.

Failure scenario: teacher uploads a 2 MB photo as the page background, then makes 50 small edits (each pushes onto the past stack). The stack holds ~50 × ~2.7 MB (base64 inflation) ≈ 135 MB of duplicated background data. On Chromebooks / classroom-grade hardware this is enough to crash the tab.

Fix options: store backgrounds out-of-band (Firestore field or Storage URL) and reference them by ID inside the SVG; or, when serializing for the undo stack, strip the background image to a placeholder and re-attach on undo.

components/widgets/SmartNotebook/components/PageEditor.tsx
```diff
@@ -390,39 +530,129 @@ export const PageEditor: React.FC<PageEditorProps> = ({
...
+    setEditing(null);
+    onChangeRef.current?.(next);
+  }, []);
+
+  // Rebuild the editor DOM from a serialized SVG string. Owns all overlay
+  // creation (selection group, corner handles, rotate handle, link FABs,
+  // marquee). Pulled out of the mount effect so undo / redo can re-init
+  // the DOM directly without needing the host to push a new svg prop
+  // through (the autosave round-trip skip would block that path).
+  const rebuildEditorDom = useCallback((svgString: string) => {
```
OPS-PIvers, line 661: **`rebuildEditorDom` (called from undo and redo) wipes `container.innerHTML` and nulls `handlesRef`/`linkHandleRef`/`marqueeRef`, but does NOT clear `dragRef`, `strokeRef`, or `shapeRef`. If undo/redo fires mid-gesture, those refs point at detached DOM nodes.**

Failure scenario: user starts a shape drag on a SMART board; a paired keyboard / MobileRemoteView / on-screen-keyboard combo sends Cmd+Z. `undo()` pops the past stack, `rebuildEditorDom` wipes the container. `shapeRef.current.wrapper` is now detached. Next pointermove calls `.setAttribute` on the orphan — user sees the gesture freeze with no error. On pointerup the commit branch runs against the detached node, `emitChange` serializes the post-undo SVG only, and the gesture is silently lost.

Fix: in `rebuildEditorDom`, also null `dragRef.current = null; strokeRef.current = null; shapeRef.current = null;` (and consider releasing any outstanding pointer capture).

components/widgets/SmartNotebook/components/PageEditor.tsx
```diff
@@ -554,16 +857,410 @@ export const PageEditor: React.FC<PageEditorProps> = ({
...
+      clone.setAttribute('data-edit-id', newId);
+      // Reset edit-bookkeeping so the pasted object starts fresh: its
+      // original transform becomes the new baseline, with no leftover
+      // edit-matrix from prior moves on the source page.
+      clone.setAttribute(
+        ORIG_TRANSFORM_ATTR,
+        clone.getAttribute('transform') ?? ''
+      );
+      clone.removeAttribute(EDIT_MATRIX_ATTR);
+      fg.appendChild(clone);
```
OPS-PIvers, line 1049: **`pasteFromClipboard` preserves the source object's absolute transform but never applies a paste offset — same-page Cmd+C → Cmd+V drops the clone exactly under the source.**

Lines 1019-1031 do `clone.setAttribute(ORIG_TRANSFORM_ATTR, clone.getAttribute('transform') ?? '')` and `removeAttribute(EDIT_MATRIX_ATTR)`, then `fg.appendChild(clone)`. Compare `duplicateSelected` (line ~858) which DOES apply `applyEditMatrix(clone, new DOMMatrix().translateSelf(20, 20))`.

Failure scenario: teacher copies a shape at (200,100) on page 1, Cmd+V on the same page → the clone renders at (200,100), perfectly stacked under the source. Selection is now on the (invisible) clone, but the user sees no change and concludes paste is broken. Cmd+D (duplicate) works fine because it offsets.

Fix: apply the same +20,+20 offset in `pasteFromClipboard` when source and destination page are the same (the cross-page case can keep the original coordinates).

components/widgets/SmartNotebook/components/pageEditorTypes.ts
```diff
@@ -4,14 +4,40 @@
...
+// Eight-color palette including white (intentionally added so teachers can
+// strike through dark imported SMART backgrounds) and a near-black for ink
+// readability. Ordered to roughly match the rainbow + neutrals at the ends.
 export const PEN_COLORS = [
-  '#e11d48',
-  '#2563eb',
-  '#16a34a',
-  '#111827',
-  '#f59e0b',
+  '#111827', // near-black
```
OPS-PIvers, line 33: **`PEN_COLORS[0]` changed from red `#e11d48` to near-black `#111827` — the default pen is now invisible on the dark SMART Notebook imported backgrounds the comment cites as the reason white was added.**

`PageEditor.tsx:292` and `PageEditorOverlay.tsx:126` both consume `PEN_COLORS[0]` as the initial pen color. `penColor` is plain `useState`, never persisted, so every fresh editor mount starts at the new default.

Failure scenario: teacher imports a SMART `.notebook` with a dark slide background, opens the editor, picks up the pen and draws — strokes are near-invisible until they manually open the color popover. Before this PR the default was red, which is high-contrast against both light and dark backgrounds.

Fix: either put a high-contrast color (red or white) at index 0 of `PEN_COLORS`, or initialize `penColor` to a specific contrast-safe hex independent of array order.

components/widgets/SmartNotebook/components/PageEditor.tsx
```diff
@@ -742,6 +1568,156 @@ export const PageEditor: React.FC<PageEditorProps> = ({
...
+      wrapper.setAttribute(ORIG_TRANSFORM_ATTR, '');
+      const textEl = document.createElementNS(SVG_NS, 'text');
+      textEl.setAttribute('x', String(p.x));
+      textEl.setAttribute('y', String(p.y));
+      textEl.setAttribute('font-family', 'sans-serif');
+      textEl.setAttribute('font-size', String(fontSize));
+      textEl.setAttribute('fill', penColorRef.current);
+      // Default placeholder so the empty object is visible until the user
+      // types; replaced by their input when the editor commits.
+      textEl.textContent = 'Text';
```
OPS-PIvers: **Text-tool drops a placeholder `<text>` with `textContent = 'Text'`, immediately calls `emitChange()`, then opens the inline editor. If the user cancels (Escape / click-away with no typing), the literal word "Text" is persisted on the page — there's no removal of the empty node.**

The Escape handler (line ~2180) only does `setEditing(null)`. The blur path (line ~2178 → `applyTextEdit` at line ~1474) writes `editing.value` (still `'Text'`) back to the object via `writeTextLines(obj, 'Text')` and calls `emitChange()` again. No empty/unchanged check, no removal of the wrapper `<g>`.

Failure scenario: teacher with the text tool active misclicks on the canvas, presses Escape — a stray "Text" label is left behind. Repeated accidental clicks litter the page.

Fix: in `applyTextEdit` (or the Escape handler), if `editing.value` is empty OR equal to the initial `'Text'` placeholder AND the user never typed (track an `dirty` flag), remove the wrapper node from the SVG and pop the placeholder snapshot off the undo stack.

components/widgets/SmartNotebook/components/PageEditorOverlay.tsx
```diff
@@ -346,49 +401,95 @@ export const PageEditorOverlay: React.FC<PageEditorOverlayProps> = ({
...
                 penColor={penColor}
                 penWidth={penWidth}
+                textSize={textSize}
+                eraserSize={eraserSize}
+                linkedObjectTargets={linkedObjectTargets}
+                imperativeApiRef={editorApiRef}
+                onSelectionChange={(sel) => setHasSelection(sel.length > 0)}
                 onChange={onEditChange}
                 onRequestLink={setLinkRequest}
+                onFollowLink={onPageChange}
```
OPS-PIvers: **`onFollowLink={onPageChange}` wires straight into `navigateToPage` with no bounds check — a stale link to a deleted page can drive `currentPage` out of `pageUrls` range.**

`Widget.tsx:469-473` `navigateToPage` does not clamp. `handleDeletePage` (Widget.tsx:585-615) clamps `currentPage` on deletion but never prunes `objectLinks` entries whose `targetPage >= newPageCount`. Compare `Viewer.tsx:260-266` which clamps explicitly: `Math.max(0, Math.min(activeNotebook.pageUrls.length - 1, targetPage))`.

Failure scenario: notebook has 6 pages with a hotspot on page 1 pointing to page 5. Teacher deletes page 5. Teacher returns to page 1, Cmd-clicks the orphaned hotspot. `onFollowLink(5)` → `navigateToPage(5)` → `setCurrentPage(5)`. The render path reads `activeNotebook.pageUrls[currentPage]` (undefined) and falls through to the non-edit/empty branch.

Fix: clamp `targetPage` in `navigateToPage` (or in the `onFollowLink` wrapper, matching the Viewer pattern). Bonus: prune `objectLinks` in `handleDeletePage` so the dead hotspots disappear from the manage-links UI too.

OPS-PIvers:
**`components/widgets/random/RandomWidget.tsx:1385`**

**`maxWordLength` (used to size the result text WHILE spinning) splits on `/\s+/` which matches NBSP. `displayedWordLength` (used AFTER settle) splits on `/[ \t\n\r\f\v]+/` which deliberately does NOT match NBSP — with an explicit comment that "CSS white-space: normal does NOT wrap at NBSP". The same browser rule applies during spinning, so `maxWordLength` under-estimates the rendered token length for NBSP-joined names.**

Failure scenario: a roster contains `"Mary Smith"` (10-char unbreakable rendered token — common with CSV imports / ClassLink). While spinning, `maxWordLength = max(['Mary','Smith']) = 5`, so the font is sized for a 5-char word (~15cqw). The browser renders the NBSP-joined name as one 10-char token and it overflows / clips on the right edge. When the spin settles, `displayedWordLength` correctly returns 10 and the font snaps smaller — visible "pop" on every spin that involves an NBSP-containing name.

Fix: use the same ASCII-only split for both formulas (apply `splitOnAsciiSpace` to the word-length scan, not just the settled display).

components/widgets/random/RandomWidget.tsx
```diff
@@ -1383,17 +1387,71 @@ export const RandomWidget: React.FC<{ widget: WidgetData }> = ({ widget }) => {
     [students]
```
OPS-PIvers, line 1387: **Empty-roster placeholder font size regressed from `min(26cqw, 20cqh)` to `min(~13cqw, 60cqh)` — the new "Ready?" prompt is roughly half the previous horizontal scale on a fresh widget.**

The previous code had an explicit `if (maxWordLength === 0) return 'min(26cqw, 20cqh)'` fallback. The new code routes the placeholder through `PLACEHOLDER_LENGTH = 6` → `cqwValue = Math.round(75/6) = 13` → `min(13cqw, 60cqh)`. On a typical wide-but-short widget shape the 13cqw is the binding constraint, so "Ready?" now renders visibly tiny.

Failure scenario: user adds a fresh Random widget without picking a class (or roster is empty). The big splashy "Ready?" is now half the size it used to be.

Fix: bring back the empty-roster branch's larger size, OR raise the placeholder cqw target so a 6-char string still fills the widget meaningfully.

components/widgets/SmartNotebook/components/PageEditor.tsx
```diff
@@ -554,16 +857,410 @@ export const PageEditor: React.FC<PageEditorProps> = ({
...
+    const parser = new DOMParser();
+    const newIds: string[] = [];
+    const newLinks: ClonedLinkInfo[] = [];
+    for (const entry of pageEditorClipboard) {
+      const doc = parser.parseFromString(
+        `<svg xmlns="${SVG_NS}">${entry.svg}</svg>`,
+        'image/svg+xml'
+      );
+      const sourceEl = doc.documentElement.firstElementChild;
+      if (!sourceEl) continue;
```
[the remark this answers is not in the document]
OPS-PIvers: Fixed in 5ff93db2 — added `if (doc.querySelector('parsererror')) continue;` before the firstElementChild read so a malformed clipboard entry is skipped instead of appending the parsererror node into the foreground.

components/widgets/SmartNotebook/components/PageEditor.tsx
```diff
@@ -554,16 +857,410 @@ export const PageEditor: React.FC<PageEditorProps> = ({
...
+        .map((c) => c as SVGGraphicsElement);
+      if (orderedSelected.length === 0) return;
+      if (direction === 'front') {
+        for (const obj of orderedSelected) fg.appendChild(obj);
+      } else if (direction === 'back') {
+        // Reverse so the first-selected ends up frontmost among the moved
+        // group, matching how "send to back" works in design tools.
+        for (let i = orderedSelected.length - 1; i >= 0; i--) {
+          fg.insertBefore(orderedSelected[i], fg.firstChild);
+        }
```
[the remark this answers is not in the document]
OPS-PIvers: Fixed in 5ff93db2 — guarded `fg.insertBefore(obj, fg.firstChild)` with `if (obj !== fg.firstChild)` so Safari/older browsers don't throw DOMException when an already-frontmost-of-back-group node is sent to back.

components/admin/WorkSymbolsConfigurationModal.tsx
OPS-PIvers: Fixed in 8fcf9267 — removed `uploadedThisSessionRef.current.clear()` from `handleSave` and updated the toast copy to be honest about local-state-only semantics. Storage cleanup now only runs on Cancel; outer Save Changes is responsible for finalizing persistence.

components/widgets/SmartNotebook/Widget.tsx
OPS-PIvers: Fixed in 8fcf9267 — `handleSaveObjectLink`/`Batch`/`Remove` now use `arrayUnion` / `arrayRemove` instead of read-modify-write. Concurrent paste-duplicate batches can no longer clobber each other's links.

components/widgets/SmartNotebook/components/PageEditor.tsx
OPS-PIvers: Fixed in 8fcf9267 — `onPaste` registers in capture phase and calls `stopImmediatePropagation` after the early-bail checks, so Dock's smart-paste listener no longer double-fires when pasting inside the editor.

components/widgets/SmartNotebook/components/PageEditor.tsx
OPS-PIvers: Fixed in 8fcf9267 — added an `isEditableTarget` helper and bailed early from the window keydown listener when the event target is an input/textarea/contenteditable/select. Cmd+Z/Y/C/[ /] keep their native semantics in other widgets.

components/widgets/SmartNotebook/components/PageEditor.tsx
OPS-PIvers: Fixed in 8fcf9267 — `emitChange` now tightens the undo cap to 5 entries when the previous baseline SVG is > 500KB (the data-URL-background case), keeping the heap bounded over long edit sessions.

components/widgets/SmartNotebook/components/PageEditor.tsx
OPS-PIvers: Fixed in 8fcf9267 — `rebuildEditorDom` now nulls `dragRef`/`strokeRef`/`shapeRef` alongside the handle/marquee refs, so undo mid-gesture can't leave them pointing at detached DOM.

components/widgets/SmartNotebook/components/PageEditor.tsx
OPS-PIvers: Fixed in 8fcf9267 — `pasteFromClipboard` now applies the same `+20,+20` translate that `duplicateSelected` uses, so same-page paste lands visibly offset instead of stacking under the source.

components/widgets/SmartNotebook/components/pageEditorTypes.ts
OPS-PIvers: Fixed in 8fcf9267 — added `DEFAULT_PEN_COLOR = '#e11d48'` (red) and switched the Overlay's `useState` default to use it. Palette ordering is preserved; default is decoupled from `PEN_COLORS[0]`.

components/widgets/SmartNotebook/components/PageEditor.tsx
OPS-PIvers: Fixed in 8fcf9267 — tracked `initialValue` + `isPlaceholder` in the `editing` state; when the user escapes or blurs without typing, the wrapper is removed instead of writing the literal 'Text' back.

components/widgets/SmartNotebook/components/PageEditorOverlay.tsx
OPS-PIvers: Fixed in 8fcf9267 — wrapped `onFollowLink` to clamp `targetPage` into `[0, pageUrls.length - 1]` mirroring `Viewer.tsx`, so stale links to deleted pages can't drive `currentPage` out of range.

components/widgets/random/RandomWidget.tsx
OPS-PIvers: Fixed in 8fcf9267 — restored the empty-roster early-return (`min(26cqw, 20cqh)`) for the no-roster / no-displayed-result / not-spinning case so "Ready?" fills the widget. Also aligned `maxWordLength`'s split with `displayedWordLength` (ASCII-only) to fix the NBSP pop-resize between spin and settle.

OPS-PIvers:
The 4 inline-comment fallbacks above (lines outside the diff hunk GitHub accepts for inline) are also fixed in 8fcf9267:

- **PageEditorOverlay.tsx:399 — notebook-switch stale DOM**: PageEditor's `key` now includes the active notebook id (\`${activeNotebookId}:${currentPage}\`), forcing a remount when switching notebooks at the same page index.
- **WorkSymbolsConfigurationModal.tsx:136 — stale-closure mutators**: \`removeSymbol\` / \`updateSymbolTitle\` / \`toggleBuilding\` (and the inline all-buildings handler) now use functional \`setSymbols(prev => ...)\` updaters, matching the \`handleFiles\` fix.
- **PageEditor.tsx:78 — module-scoped clipboard leak**: \`pageEditorClipboard\` is now a per-instance \`clipboardRef\`, so two notebook widgets can't cross-contaminate and copied content can't survive sign-out within the SPA.
- **RandomWidget.tsx:1385 — NBSP split inconsistency**: \`maxWordLength\` now uses the same ASCII-only whitespace regex as \`displayedWordLength\` so NBSP-joined names don't pop-resize between spin and settle.

All 15 review findings are now resolved. Lint, type-check, and 3698 unit tests pass.

PR #1726, opened 2026-05-28
fix(widgets): resolve UUID group IDs to human-readable names when sending groups to Stations

OPS-PIvers:
## Root cause

`buildStationsFromRandomGroups` in `components/widgets/Stations/nexus.ts` used `group.id?.trim()` directly as the station title. The random groupmaker (`groupMaker.ts`) assigns `id: crypto.randomUUID()` at five call sites — so every station title produced by "Send Groups → Stations" was a raw UUID string like `550e8400-e29b-4000-a000-426614174000` instead of "Group 1" / "Group 2". Teachers saw meaningless UUID labels in the Stations widget.

## Fix

- Added `UUID_RE` (v4 UUID regex) to detect synthetic group IDs.
- Added optional `sharedGroups?: SharedGroup[]` parameter to `buildStationsFromRandomGroups`, built into a lookup Map.
- UUID IDs are resolved through the Map; non-UUID IDs take the original legacy path (preserving custom names like "Reading Corner").
- Updated `RandomSettings.tsx` (`handleSendGroupsToStations`) to pass `activeDashboard?.sharedGroups`.

## Rejected band-aid

Simply replacing any UUID id with `"Group N"` — that would produce readable labels but lose the user's saved group names from `sharedGroups`. The root fix resolves through the shared groups map and falls back to "Group N" only when genuinely unnamed.

## Regression test

`tests/components/widgets/Stations/nexus.test.ts` — 4 new cases covering:
- UUID ids resolved via `sharedGroups`
- UUID ids fall back to "Group N" when no `sharedGroups` passed
- UUID ids fall back to "Group N" when matched entry has a blank name
- Non-UUID string ids still take the legacy path unchanged

All 9 tests (5 pre-existing + 4 new) pass. Test FAILS on `dev-paul` baseline, PASSES on this branch.

## Risk tag

**contained** — change is limited to `nexus.ts` and one call site in `RandomSettings.tsx`; no schema or data-format changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01T6TyuZzhXYCsuwYdvbQ81q)_


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-05-28

### Summary
Fixes the "Send Groups → Stations" flow so station titles aren't raw UUIDs. Adds a UUID-v4 regex + optional `sharedGroups` map lookup in `buildStationsFromRandomGroups`; falls back to "Group N" when a UUID id isn't resolvable. Preserves the legacy path for non-UUID ids.

### Code Quality
- ✓ Clean three-branch decision tree (no id → "Group N"; UUID → resolve via map or fall back; non-UUID → trim and use as-is).
- ✓ `UUID_RE` is case-insensitive and version-4-strict — appropriate for IDs from `crypto.randomUUID()`.

### Widget Standards
- ✓ Touches `components/widgets/Stations/nexus.ts` (helper file, not the widget component), so widget registry / scaling rules don't apply.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No new collections, no rule changes.

### Regression Risk
- ✓ Low risk. The legacy `group.id` → title path is preserved for any non-UUID string id; only UUIDs now route through the new resolver. Any caller that wasn't passing `sharedGroups` gets the "Group N" fallback (matches the previous unreadable behaviour, but now at least readable).

### Missing Tests
- ✓ Adequate coverage. 4 new tests added covering: UUID → name resolution, UUID → fallback when no `sharedGroups`, UUID → fallback when `sharedGroups` entry has blank name, and the non-UUID legacy path. Suite goes from 5 → 9.

### Overall
**Merge readiness:** Ready.
Self-contained bugfix with thorough regression coverage.

---

---

PR #1749, opened 2026-05-29
fix(dice): use refs to prevent stale-closure bug when props change mid-roll

OPS-PIvers:
## Root Cause

`DiceWidget`'s `roll()` function captured `diceCount`, `config`, and `widget.id` from the render closure at the moment `setInterval` was created. If props changed mid-roll (e.g. the user adjusted diceCount from 2→4 via the settings panel while the animation was running), the final `updateWidget` call at tick 12 used stale captured values — persisting a `lastRoll` array of 2 elements when the widget expected 4.

## Fix Summary

Assign `diceCountRef.current`, `configRef.current`, and `widgetIdRef.current` synchronously in the render body (the canonical React pattern for reading current prop values inside async callbacks). The `setInterval` callback reads from refs at the moment of execution, not from the closure.

## Rejected Band-Aid

Wrapping the interval in a `useEffect` that re-schedules whenever `diceCount` changes would restart the animation from scratch on every re-render mid-roll — worse UX than the original bug.

## Regression Test

`'persists lastRoll with the current diceCount when count changes mid-roll'` in `components/widgets/DiceWidget/Widget.test.tsx`:
- **Before fix**: `expected [ 1, 1 ] to have a length of 4 but got 2` — stale closure persisted old count
- **After fix**: 4/4 tests pass

## Risk

**Contained** — change is local to `DiceWidget`; refs are only read inside the interval callback, no rendering side-effects.

---
_Generated by [Claude Code](https://claude.ai/code/session_01B2bZb7zDM1bRRUzrV7LPcq)_

### DIFF
--- a/components/widgets/DiceWidget/Widget.test.tsx
+++ b/components/widgets/DiceWidget/Widget.test.tsx
@@ -97,4 +97,43 @@ describe('DiceWidget', () => {
 
     vi.useRealTimers();
   });
+
+  it('persists lastRoll with the current diceCount when count changes mid-roll', () => {
+    // Regression test for stale-closure bug: roll() captured diceCount from
+    // the render scope. If props changed from 2→4 while the interval was
+    // running, the final updateWidget call used the stale count (2) instead
+    // of the new count (4), persisting a lastRoll array with the wrong length.
+    vi.useFakeTimers();
+
+    const { rerender } = render(<DiceWidget widget={createWidgetData(2)} />);
+
+    const rollButton = screen.getByRole('button', { name: /Roll Dice/i });
+    act(() => {
+      rollButton.click();
+    });
+
+    // Advance mid-roll (interval fires every 80ms, 12 ticks total = 960ms)
+    act(() => {
+      vi.advanceTimersByTime(400);
+    });
+
+    // Change dice count to 4 while the roll is in-flight
+    rerender(<DiceWidget widget={createWidgetData(4)} />);
+
+    // Complete the roll
+    act(() => {
+      vi.advanceTimersByTime(1000);
+    });
+
+    // The last updateWidget call should persist a lastRoll array whose length
+    // matches the NEW diceCount (4), not the stale one (2).
+    const allCalls = mockUpdateWidget.mock.calls;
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-05-29

### Summary
Closes a stale-closure bug in `DiceWidget` where `roll()` captured `diceCount`, `config`, and `widget.id` from the render scope at `setInterval` creation. Mid-roll prop changes (e.g. count 2→4 via Settings) caused the terminal `updateWidget` call to persist a `lastRoll` array with the *old* length. Fix uses the canonical React ref-in-render pattern: refs are assigned synchronously in the render body (`diceCountRef.current = diceCount`) and read inside the interval callback at execution time.

### Code Quality
✓ Ref-in-render assignment is the right primitive for "always read latest prop in async callback" — preferable to wrapping in `useEffect` (would re-mount the interval on every prop change, restarting the animation mid-roll, which is exactly the band-aid the PR explicitly rejects). Inline comment is clear.

### Widget Standards
✓ Follows the CLAUDE.md guidance: "Assign refs directly in the render body: `myRef.current = value` — no effect needed."

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ Improves Firestore write integrity — the persisted `lastRoll` array length now reliably matches the current `diceCount`.

### Regression Risk
Low. The refs are read-only inside the interval; no rendering side-effects. Other DiceWidget code paths (initial state, manual settings flip) unaffected.

### Missing Tests
✓ Adequate. Adds a focused regression test that fakes timers, advances mid-roll, rerenders with new diceCount, and asserts the final `updateWidget` persists `lastRoll.length === 4`. Test pattern is reusable for any future timer-widget stale-closure regression.

### Regression Risk Audit (Wider)
The CLAUDE.md backlog note adds a follow-up to audit `TimerWidget`, `StopwatchWidget`, `CountdownWidget` for the same pattern. Worth doing — they share the setInterval-from-callback shape that's vulnerable here.

### Overall
**Merge readiness:** Ready
Surgical fix with the right primitive, clear test, and a thoughtful follow-up backlog entry.

---

---

PR #1795, opened 2026-06-01
Extract guarded grade-passback scaling into a shared, tested helper

OPS-PIvers:
## What

Extracts the "scale a student's earned points onto the Google Classroom attachment's frozen `maxPoints`" grade-passback math out of `QuizResults` and into a single pure helper:

```ts
scaleEarnedToMaxPoints(response, questions, session, total, maxPoints)
  -> { pseudonymUid, pointsEarned }
```

It lives in `components/widgets/QuizWidget/utils/quizScoreboard.ts`, right next to `getEarnedPoints`, and carries all three guards:

- **`Number.isFinite` guard** — a non-finite earned score (e.g. `getEarnedPoints` returning `NaN` for a response with missing/un-gradeable answers) is coerced to `0` rather than propagating `NaN` through the clamp and making the Cloud Function reject the grade entry.
- **zero-total guard** — `total <= 0` scales to `0` instead of dividing by zero.
- **round + clamp** to `[0, maxPoints]`.

`QuizResults.pushClassroomGrades` now calls the helper instead of inlining the scale.

## Why

The earned→`maxPoints` scaling was inlined in `QuizResults` with a comment documenting the exact `NaN`-rejection risk the `Number.isFinite` guard protects against. Inlined math like this invites a second passback caller (e.g. a teacher review route) to copy it *without* the guard and silently diverge — `Math.round(NaN)` would slip through the clamp and the CF would reject the entry. Funneling all callers through one helper removes that divergence risk and makes the math testable without rendering a results view.

## Note on scope

The task referenced a second, unguarded copy in `components/classroomAddon/TeacherReviewRoute.tsx`. **That file does not exist in the current tree** — the only `getEarnedPoints`-based scaling site today is `QuizResults` (already guarded), and the Video Activity results view uses a different percentage-based formula (`getStudentScore`) that has its own guard. So there is no current divergence to repair; this PR is the preventative extraction so any future passback caller shares the guarded math. Happy to wire up the helper at a second call site if/when one lands.

## Tests

Added a `scaleEarnedToMaxPoints` suite to `quizScoreboard.test.ts` covering:

- exact-total (`earned == total` → `maxPoints`)
- partial (scales onto a drifted denominator and rounds)
- over-total clamp (scaled value exceeds `maxPoints` → clamped)
- zero-total guard (→ `0`)
- non-finite earned (`getEarnedPoints` → `NaN` → `0`)

`type-check`, `eslint`, `prettier --check`, and the full `quizScoreboard.test.ts` suite (61 tests) all pass locally.

https://claude.ai/code/session_016bxfSdZDysV5eFsXkFzbxk

---
_Generated by [Claude Code](https://claude.ai/code/session_016bxfSdZDysV5eFsXkFzbxk)_

### DIFF
--- a/components/widgets/QuizWidget/components/QuizResults.tsx
+++ b/components/widgets/QuizWidget/components/QuizResults.tsx
@@ -58,6 +58,7 @@ import {
   getDisplayScore,
   getScoreSuffix,
   getEarnedPoints,
+  scaleEarnedToMaxPoints,
   isGamificationActive,
 } from '../utils/quizScoreboard';
 import { resolveResponseDisplayName } from '../utils/resolveDisplayName';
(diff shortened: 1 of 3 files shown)


OPS-PIvers:
Closing as redundant. This was opened against `main`, but `dev-paul` already implements this exact task — and better: `utils/classroomGradePush.ts` defines `buildQuizClassroomGradeEntries(...)` (the `Number.isFinite` guard, zero-total guard, and round+clamp to `[0, maxPoints]`), it's the single source of truth used by **both** `QuizResults.tsx` and `components/classroomAddon/TeacherReviewRoute.tsx`, and `tests/utils/classroomGradePush.test.ts` covers all the cases. Retargeting this PR to `dev-paul` would conflict and reintroduce a duplicate helper, so there's nothing to merge.

---

PR #1816, opened 2026-06-02
Harden classroom_course_links writes + add unlink correction path

OPS-PIvers:
Deferred from the PR #1813 release review. PR #1810 moved `classroom_course_links/{courseId}` writes behind the `linkClassroomCourse` Cloud Function (server-side teacher verification + no-hijack guard) and set the rules to read-only. This addresses the two follow-ups.

## 1. TOCTOU fix — `linkClassroomCourse` write is now transactional
The no-hijack check + write was a non-atomic read-then-write, so two co-teachers clicking **Link** on the *same never-linked* course within the same window could both pass the `exists?` check and write (last-writer-wins). It was benign (both racers still clear teacher-verification, so a non-teaching squatter can never win — it only decided *which authorized co-teacher* owned the link), but it's now deterministic:

- The check + write run inside `db.runTransaction`; the read locks the doc so the racers serialize — the first commits the create, the second's read sees it and is either **merged** (same teacher) or rejected with `already-exists` (different teacher).
- Teacher verification (`listTeacherCourseIds`) stays **outside** the transaction — it's a network call and the body may retry on contention.

## 2. Correction path — new `unlinkClassroomCourse` CF
Client deletes are blocked by the rules and there was no delete CF, so a wrong course→roster mapping (wrong `rosterId`, or a teacher who **left the district**) was permanent without Firestore Console access — and `already-exists` blocked even a legitimate co-teacher from re-linking.

- New `unlinkClassroomCourse` callable: **same trust anchor** as link (re-verify the caller teaches the Google course with their own `classroom.courses.readonly` token, **fail closed** on any verification error), identity from `request.auth.uid` only, **transactional read-then-delete**, **idempotent** (`removed: false` when there's nothing to delete).
- The common "wrong `rosterId`, same teacher" fix already works via a link re-merge (no unlink needed); this CF covers removal and the departed-teacher case.

### Documented decision — co-teacher takeover
A verified teacher of the course who is **not** the current owner (e.g. cleaning up after a colleague left the district) **is allowed** to remove/replace the link. This is safe and is the deliberate escape hatch the `linkClassroomCourse` no-hijack guard intentionally lacks because:
- it's still gated on **server-side teaching verification** — a non-teacher can never reach the delete (the squatting-fix invariant holds);
- removal is an **explicit, destructive** action on the same course the caller provably teaches — not a silent re-point during the normal link flow;
- it only **deletes** — it never assigns ownership to the caller; re-linking goes back through `linkClassroomCourse`, which re-runs verification and stamps a fresh `teacherUid`.

Each takeover (a different teacher's link removed) is logged for the audit trail. Restricting unlink to the owner only would re-create the exact "permanent stale link after a teacher leaves" gap this CF exists to close.

## Invariants preserved (squatting fix)
- ✅ Server-side teacher verification on every write/delete
- ✅ `teacherUid` always from `request.auth.uid`, never the client payload
- ✅ Fail-closed on any verification/upstream error
- ✅ No silent hijack (link still refuses a different teacher with `already-exists`)

## Other changes
- **Client:** wire an **Unlink** button into the "Link to Google Classroom" modal (shown when the roster is already linked) that calls the new CF and clears the roster's local `googleClassroomCourseId` mirror.
- **Rules:** updated the `classroom_course_links` comment to cover both CFs (writes *and* deletes are server-only).
- **Exports:** `unlinkClassroomCourse` re-exported from `functions/src/index.ts`.

## Tests / validation
- `functions/src/classroomAddonAuth.test.ts`: 47 → **55** tests. Added: link write happens inside a transaction (TOCTOU); unlink by owner (transactional); verified co-teacher takeover; non-teacher refused (no delete); fail-closed on verification error; unauthenticated rejected before any Classroom call; idempotent no-op; input validation. The mock now supports `runTransaction` + `doc().delete()`.
- Ran from repo root: `pnpm run type-check:all` ✓, `pnpm run lint` ✓, `pnpm run format:check` ✓, functions vitest ✓ (55/55).

---
_Generated by [Claude Code](https://claude.ai/code/session_01WSEvvUv3svziMSxeNQtxRY)_


OPS-PIvers (commented):
## Cross-PR dependency: `unlinkClassroomCourse` references `listTeacherCourseIds` which PR #1814 removes

`unlinkClassroomCourse` in `functions/src/classroomAddonAuth.ts` uses `classroomAddonNet.listTeacherCourseIds(accessToken)` as its trust anchor, and the new test cases stub it with `vi.spyOn(classroomAddonNet, 'listTeacherCourseIds')`.

PR #1814 (also targeting `dev-paul`) replaces `classroomAddonNet.listTeacherCourseIds` with `classroomAddonNet.verifyTeacherOfCourse` — a single `courses.teachers.get` call instead of a paginated `courseStates=ACTIVE` enumeration. Once either PR is merged and the other is rebased, `unlinkClassroomCourse` will reference a non-existent method and throw at runtime.

**Required update:** replace the trust-anchor call in `unlinkClassroomCourse` (and its spies in the test):

```typescript
// Replace this:
const teacherCourses = await classroomAddonNet.listTeacherCourseIds(accessToken);
if (!teacherCourses.ok) { throw new HttpsError('unauthenticated', ...); }
if (!teacherCourses.courseIds.includes(courseId)) { throw new HttpsError('permission-denied', ...); }

// With this (mirrors the linkClassroomCourse trust anchor after #1814):
const verification = await classroomAddonNet.verifyTeacherOfCourse(accessToken, courseId);
if (!verification.ok) { throw new HttpsError('unauthenticated', ...); }
if (!verification.isTeacher) { throw new HttpsError('permission-denied', ...); }
```

This also gains `verifyTeacherOfCourse`'s improvements for free:
- **State-agnostic** — a teacher of an ARCHIVED or PROVISIONED course can unlink it (with `listTeacherCourseIds` and `courseStates=ACTIVE`, those teachers would be wrongly refused)
- **One API call** instead of up to 25 paginated requests
- **Symmetric** with the link path's trust anchor, so both CFs share the same verification semantics

---

## `handleUnlink`: successful server-side unlink masked as failure when `updateRoster` throws

In `SidebarClasses.tsx`, `handleUnlink` makes two sequential awaits:

```typescript
await unlinkCourse({ accessToken, courseId });           // CF — deletes classroom_course_links doc
await updateRoster(linkingRoster.id, { googleClassroomCourseId: '' });  // Firestore write
```

If `unlinkCourse` succeeds but `updateRoster` throws (transient Firestore error), execution falls into the catch block, which shows "Failed to unlink from Google Classroom" and leaves `isUnlinking = false`. But the canonical link in `classroom_course_links/{courseId}` was already deleted — the unlink **did** succeed.

Consequences:
- Toast message is wrong (the user sees "failed" when the server-side operation worked).
- The modal stays open with `linkingRoster.googleClassroomCourseId` still set locally — Firestore won't auto-clear it, since only the `classroom_course_links` doc was deleted, not the roster doc's `googleClassroomCourseId` field.
- The Unlink button remains visible; clicking it again calls the CF which returns `{ removed: false }` (idempotent), then retries `updateRoster`. This can self-recover but the teacher may have already navigated away confused.

**Suggestion:** split the steps so a successful CF call always closes the modal and shows success, then cleans up the local mirror separately:

```typescript
await unlinkCourse({ accessToken, courseId });
// CF succeeded — close the modal and show success regardless of what follows.
closeLinkModal();
addToast(t('sidebar.classes.linkGoogleClassroom.unlinkSuccess', { defaultValue: 'Unlinked from Google Classroom.' }), 'success');
// Best-effort local mirror clear; failure here is non-critical (Firestore
// listeners will eventually re-read, and the canonical link is already gone).
try {
  await updateRoster(linkingRoster.id, { googleClassroomCourseId: '' });
} catch {
  // Ignore — the server-side link is deleted; local state will sync.
}
```

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-06-03

### Summary
Adds the `unlinkClassroomCourse` Cloud Function (the server-verified correction path the squatting fix deliberately omitted), wraps both link and unlink check-then-write in a transaction (TOCTOU fix), and adds an Unlink button to the link modal. Net security posture improves.

### Code Quality
✓ No issues found. The new i18n strings use inline `defaultValue` fallbacks (so no raw keys appear); consider adding them to the locale files in a follow-up for full DE/FR/ES coverage, but it is not blocking.

### Firestore & Security
✓ Strong. `firestore.rules` keeps `allow write: if false` (covering create/update **and** delete), so deletes are server-only. The unlink CF re-verifies teaching authority with the caller's own token and fails closed; the **co-teacher takeover** is a documented, deliberate decision (verified teacher of the same course may remove a departed colleague's link) that preserves the non-teacher-can-never-delete invariant. The transaction makes concurrent link/unlink deterministic.

### Regression Risk
⚠️ Overlaps with #1814 — both rewrite the `linkClassroomCourse` trust anchor / transaction body in `classroomAddonAuth.ts`. Verify the merge order so the TOCTOU transaction and the `verifyTeacherOfCourse` switch compose correctly (one shouldn't revert the other).

### Missing Tests
✓ Excellent — adds transaction-count assertions, co-teacher takeover, non-teacher denial, fail-closed, idempotent no-op, and arg-validation tests.

### Overall
Solid, well-tested security feature; coordinate the merge with #1814 and consider locale-file entries for the new strings.

---

---

PR #1862, opened 2026-06-04
Promote dev-paul → main: in-frame LTI deep-link picker

OPS-PIvers:
Promotes the in-frame Schoology LTI deep-link picker to production.

- `33e77bb6` feat(lti): run the picker **in-frame**; the top-level redirect is now a fallback. Google sign-in works in Schoology's iframe (scopes fixed), so the return POSTs inside the iframe → Schoology's content-return page runs natively → **no 'Redirecting to an external app' modal, no 'Duplicate timestamp/nonce' 401**. Launch code is exchanged after sign-in so the fallback can redirect with it unconsumed (5-min TTL).
- `08c21bf9` fix(lti): keep the picker shown once the deep-link context is loaded (review hardening — avoid a dead sign-in card on a transient token blip).

Reviewed by 3 parallel agents (state machine / fallback+single-use-lc / in-frame-return+security): clean after the hardening fix. Validation: type-check ✓ · lint ✓ · 4052 tests ✓. Only testable against a real Schoology launch (prod).

### DIFF
--- a/components/lti/LtiDeepLinkPicker.tsx
+++ b/components/lti/LtiDeepLinkPicker.tsx
@@ -378,7 +378,11 @@ const LtiDeepLinkFlow: React.FC = () => {
             </p>
           </div>
         </AddonCard>
-      ) : !teacherReady ? (
+      ) : !teacherReady && !deepLinking ? (
+        // Sign-in card only BEFORE the exchange. Once `deepLinking` is loaded we
+        // stay on the picker even if `teacherReady` momentarily blips false (e.g.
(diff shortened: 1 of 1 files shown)


OPS-PIvers:
No changelog entry — release contains only internal changes.

This promote is a single file / 2 commits, both reworking the internal mechanics of the Schoology deep-link picker's sign-in and return flow (in-frame vs. top-level fallback, launch-code timing, nonce handling, plus a review-hardening fix). No new teacher-facing capability, and the integration itself has never been announced in "What's New," so there's nothing here to surface to teachers.

---

PR #1913, opened 2026-06-09
fix(dock): add SELECT to smart-paste guard to prevent widget creation on select-element paste

OPS-PIvers:
## Root cause

The smart-paste `useEffect` in `components/layout/Dock.tsx` guards against intercepting clipboard events when the user is typing in a form field, but only checked `INPUT`, `TEXTAREA`, and `isContentEditable` — omitting `SELECT`. Since paste events bubble up to `window`, a paste fired while a `<select>` element has focus bypassed the guard. If the clipboard contained a YouTube URL, `detectWidgetType()` would match it as an Embed widget and call `addWidget('embed', ...)`, silently creating a widget the user never asked for.

Every other keyboard/paste guard in the codebase (`DashboardView.tsx`, `SettingsPanel.tsx`, `DraggableWindow.tsx`) already includes `SELECT`.

## Fix summary

- Added `target.tagName === 'SELECT'` to the `handlePaste` guard in `components/layout/Dock.tsx`
- Updated the guard comment from "input or textarea" to "input, textarea, or select"
- Added one regression test to `tests/components/layout/Dock.test.tsx`

## Band-aid rejected

Could have narrowed the `detectWidgetType` URL patterns to be less aggressive. Rejected — the guard is the right place to prevent interception of form-element paste events; narrowing detection patterns would just shift the problem.

## Regression test

`tests/components/layout/Dock.test.tsx` — new test: "Dock smart-paste – SELECT element is excluded from paste interception". Asserts `addWidget` is **not** called when paste target is a `<select>` and clipboard contains a YouTube URL.

**Before (dev-paul):** 5/6 pass — new test fails (`addWidget` was called once).
**After (this branch):** 6/6 pass.

## Risk

**Contained** — single guard condition change in `Dock.tsx`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FrCvf7ytfAG8jXWNypeMSH)_

### DIFF
--- a/components/layout/Dock.tsx
+++ b/components/layout/Dock.tsx
@@ -421,11 +421,12 @@ export const Dock: React.FC = () => {
         return;
       }
 
-      // Don't intercept if user is typing in an input or textarea
+      // Don't intercept if user is typing in an input, textarea, or select
       const target = e.target as HTMLElement;
       if (
         target.tagName === 'INPUT' ||
         target.tagName === 'TEXTAREA' ||
+        target.tagName === 'SELECT' ||
         target.isContentEditable ||
         e.defaultPrevented // Respect existing handlers (e.g. StickerBook)
       ) {
--- a/components/layout/Dock.tsx
+++ b/components/layout/Dock.tsx
@@ -422,12 +422,12 @@ export const Dock: React.FC = () => {
       }
 
       // Don't intercept if user is typing in an input, textarea, or select
-      const target = e.target as HTMLElement;
+      const target = e.target as HTMLElement | null;
       if (
-        target.tagName === 'INPUT' ||
-        target.tagName === 'TEXTAREA' ||
-        target.tagName === 'SELECT' ||
-        target.isContentEditable ||
+        target?.tagName === 'INPUT' ||
+        target?.tagName === 'TEXTAREA' ||
+        target?.tagName === 'SELECT' ||
+        target?.isContentEditable ||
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-06-09

### Summary
Adds `SELECT` to the `handlePaste` typing-field guard in `Dock.tsx` so a paste fired while a native dropdown is focused no longer bubbles to the window handler and silently creates a widget. Also hardens `e.target` access with optional chaining. Small, correct, consistent with the other paste/keyboard guards in the codebase.

### Code Quality
✓ No issues found. The `target?.` optional chaining (also requested by gemini-code-assist, thread now resolved/outdated) avoids null-target crashes in synthetic events.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found.

### Regression Risk
✓ Low risk. Single guard condition; existing behavior for INPUT/TEXTAREA/contentEditable unchanged.

### Missing Tests
✓ Adequate coverage. New regression test asserts `addWidget`/`addToast` are not called on a SELECT-targeted YouTube-URL paste. Minor nit (non-blocking): the test renders `<Dock />` twice (once before, once after installing the `addEventListener` spy) — only the second registration is asserted, so the first render is redundant and could be dropped for clarity.

### Overall
**Merge readiness:** Ready
Closes a real accidental-widget-creation path; aligns the guard with `DashboardView`/`SettingsPanel`.

---

---

PR #1960, opened 2026-06-13
fix(state): extract normalizeActivityWallLibraryEntry to prevent field-stripping on snapshot refresh

OPS-PIvers:
## Bug

`useActivityWallLibrary.ts`'s `onSnapshot` handler normalized each library entry using a hand-enumerated literal return:

```ts
return {
  id: storedId ?? docId,
  title: typeof title === 'string' ? title : '',
  // ... 8 more explicit fields ...
};
```

Every Phase 5A optional field (`classIds`, `rosterIds`, `periodNames`, `endedAt`, `sessionOptions`, etc.) not in the list was silently dropped on every live snapshot refresh. The initial `getDoc` call used raw `data as ActivityWallLibraryEntry`, so entries had all fields intact at first load. Once the first `onSnapshot` fired, `classIds` disappeared — breaking roster-based activity wall session filtering for ClassLink SSO deployments.

Same pattern previously fixed in:
- `normalizeVideoActivitySession` (#1902)
- `normalizeMiniAppSession` (#1912)

## Fix

Extracted `normalizeActivityWallLibraryEntry` to `utils/activityWallNormalize.ts`:

```ts
export function normalizeActivityWallLibraryEntry(
  data: Record<string, unknown>,
  docId: string
): ActivityWallLibraryEntry {
  const { id: storedId, title, ..., classId, createdAt, updatedAt, ...restData } = data;
  // explicit normalization guards for each field ...
  return {
    ...restData,   // preserves all future optional fields
    id: storedId ?? docId,
    title: ...,
    // explicit overrides take precedence
  };
}
```

Updated `useActivityWallLibrary.ts` to call the extracted function and updated `saveActivity` to use it for consistency.

## Regression test

`tests/utils/activityWallLibraryNormalize.test.ts` — adds `classIds: ['class-a', 'class-b']` to Firestore snapshot data and asserts they survive normalization.

- **FAILS on unpatched code**: `entry.classIds` is `undefined` (not in hand-enumerated literal)
- **PASSES after fix**: `entry.classIds` is `['class-a', 'class-b']` (spread from `restData`)

## Validation

| Check | Result |
|---|---|
| type-check | ✓ |
| eslint (hook + util + tests) | ✓ 0 warnings |
| format:check | ✓ |
| vitest (full suite) | ✓ all pass (3 pre-existing perf test timeouts unrelated to this change — confirmed by running editorPerf tests on unpatched code, same failures) |

---
_Generated by [Claude Code](https://claude.ai/code/session_01WWuzsxj5f2JoQPhfJEDGnU)_

### DIFF
--- a/hooks/useActivityWallLibrary.ts
+++ b/hooks/useActivityWallLibrary.ts
@@ -27,6 +27,7 @@ import {
 } from 'firebase/firestore';
 import { db } from '@/config/firebase';
 import type { ActivityWallLibraryEntry } from '@/types';
+import { normalizeActivityWallLibraryEntry } from '@/utils/activityWallNormalize';
 
 const COLLECTION = 'activity_wall_activities';
 
@@ -72,23 +73,12 @@ export const useActivityWallLibrary = (
     const unsub = onSnapshot(
       q,
       (snap) => {
-        const list: ActivityWallLibraryEntry[] = snap.docs.map((d) => {
-          const data = d.data() as Partial<ActivityWallLibraryEntry>;
-          const entry: ActivityWallLibraryEntry = {
-            id: data.id ?? d.id,
-            title: data.title ?? '',
-            prompt: data.prompt ?? '',
-            mode: data.mode ?? 'text',
-            moderationEnabled: !!data.moderationEnabled,
-            identificationMode: data.identificationMode ?? 'anonymous',
-            createdAt: data.createdAt ?? 0,
-            updatedAt: data.updatedAt ?? 0,
-          };
-          if (typeof data.classId === 'string' && data.classId.length > 0) {
-            entry.classId = data.classId;
-          }
-          return entry;
-        });
+        const list = snap.docs.map((d) =>
+          normalizeActivityWallLibraryEntry(
+            d.id,
+            d.data() as Partial<ActivityWallLibraryEntry>
+          )
+        );
         setActivities(list);
         setLoading(false);
       },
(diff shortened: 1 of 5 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-06-13

### Summary
Extracts `normalizeActivityWallLibraryEntry` to `utils/activityWallNormalize.ts`, replacing the hand-enumerated literal in `useActivityWallLibrary`'s `onSnapshot` that silently dropped Phase 5A optional fields (`classIds`, `rosterIds`) on every live refresh. `saveActivity` now spreads the full entry too. Same field-stripping class of bug previously fixed in #1902/#1912.

### Code Quality
✓ No issues found. The `...restData` spread-then-override pattern is the right fix — it future-proofs against newly-added optional fields. The empty-`classId` omission guard (preserving the `passesStudentClassGate` invariant) is retained, and `types.ts` documents the deprecation/relationship between `classId` and `classIds`.

### Widget Standards
✓ Not applicable — hook/util/type change, no front-face scaling.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. The `classId`-when-non-empty rule is preserved so an empty string never reaches Firestore; no rule changes required.

### Regression Risk
✓ Low risk. Both the read path (`onSnapshot`) and write path (`saveActivity`) now route through the shared logic, removing the drift that caused the bug.

### Missing Tests
✓ Adequate coverage — dedicated `activityWallLibraryNormalize.test.ts` (field preservation + all required-field defaults) plus hook-level snapshot and `saveActivity` regression tests.

### Overall
**Merge readiness:** Ready
Well-targeted data-loss fix with thorough tests.

---

---

PR #1974, opened 2026-06-15
fix(widgets/random): guard group-rename onBlur against stale Escape-cancel closure

OPS-PIvers:
## Root Cause

`GroupDropZone.cancel()` in `components/widgets/random/RandomGroups.tsx` called `setEditingName(false)` before setting any guard. React batches the state update; when it commits, the focused `<input>` is removed from the DOM. Browsers fire `blur` synchronously during DOM removal. The `onBlur={commit}` handler at that moment is the stale closure from the previous render — it still holds `draft = 'Edited Name'` (the text the user typed) rather than the reverted value. So pressing Escape to cancel a group rename persisted the unwanted name.

## Fix

Set `cancelledRef.current = true` synchronously at the top of `cancel()`, before any `setState`. `commit()` checks the flag on entry and short-circuits. The flag is a ref (not state), so the write is immediate and is visible to the stale `commit` closure when it fires during blur.

Same pattern as `DraggableWindow.isCancellingTitleRef` (#1965).

## Band-Aid Rejected

Adding `e.preventDefault()` to the Escape keyDown event — browsers only prevent the `blur` from focus-related actions, not from DOM removal. The blur from unmounting always fires regardless of `preventDefault`.

## Regression Test

`components/widgets/random/RandomGroups.test.tsx` — new file, 8 tests.

The critical test uses `act(() => { fireEvent.keyDown(Escape); fireEvent.blur(); })` to replicate the browser's synchronous blur-during-unmount order (jsdom does not fire blur automatically on DOM removal):

- **FAIL before fix**: `onRenameGroup` called with `['g1', 'Unwanted Name']`
- **PASS after fix**: `onRenameGroup` not called

## Risk Tag

**contained** — change is isolated to `GroupDropZone` in `RandomGroups.tsx`. No shared utilities or context modified.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J2yQ8dSA7nPUQdnvEUuKNa)_

### DIFF
--- a/components/widgets/random/RandomGroups.test.tsx
+++ b/components/widgets/random/RandomGroups.test.tsx
@@ -1,0 +1,190 @@
+/**
+ * Regression tests for RandomGroups / GroupDropZone.
+ *
+ * Bug: onBlur/Escape stale-closure in group rename input
+ * -------------------------------------------------------
+ * When the user pressed Escape to cancel a group rename, React queued the
+ * state updates from cancel() (setEditingName(false), setDraft(groupName))
+ * but the DOM blur event fired synchronously BEFORE those updates were
+ * applied.  The onBlur={commit} handler ran with a stale draft closure that
+ * still held the edited text, so pressing Escape persisted the value the
+ * user intended to discard.
+ *
+ * Fix: `cancelledRef.current = true` is set at the very start of cancel()
+ * (a ref write, synchronous and immediate), and commit() bails out early
+ * when it sees that flag.
+ */
+
+import React from 'react';
+import { act, render, screen, fireEvent } from '@testing-library/react';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { RandomGroups } from './RandomGroups';
+import type { RandomGroup } from '@/types';
+
+// dnd-kit needs minimal stubs so components render without a real drag context.
+vi.mock('@dnd-kit/core', () => ({
+  useDroppable: () => ({ isOver: false, setNodeRef: vi.fn() }),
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-06-15

### Summary
Fixes a stale-closure bug in `GroupDropZone`'s rename input where pressing Escape persisted the discarded name because the unmount-fired blur ran `commit()` with a stale `draft`. The `cancelledRef` guard (set synchronously in `cancel()`, checked + cleared in `commit()` and reset when a new rename session opens) is the correct, minimal fix and mirrors the established `DraggableWindow` #1965 pattern.

### Code Quality
✓ No issues found. Ref write is synchronous and reset on session start (`editingName` effect), so a cancelled rename won't suppress a later legitimate commit — covered by the "allows a subsequent rename after a cancelled one" test.

### Widget Standards
✓ No issues found — change is interaction logic, not CSS scaling/registry.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found — no data-layer changes.

### Regression Risk
✓ Low risk. Isolated to `GroupDropZone` in `RandomGroups.tsx`; no shared utils/context touched.

### Missing Tests
✓ Adequate coverage — 8 tests including the Enter/blur commit paths, the empty/unchanged no-ops, and the `act(() => { keyDown(Escape); blur(); })` regression that reproduces the synchronous blur-during-unmount.

### Overall
**Merge readiness:** Ready
Tight, well-tested fix consistent with the prior stale-closure remediations.

---

---

PR #1976, opened 2026-06-15
fix(migration): add range guards to proportionsValid in migrateWidgetToProportional

OPS-PIvers:
## Root Cause

`migrateWidgetToProportional` (`utils/migrateProportionalLayout.ts`) checked `Number.isFinite` on all four proportional fields before taking the "already migrated" early-return path. But `Number.isFinite(300)` is `true` — pixel values are finite. A widget in a partial-migration state (e.g. `xProp: 300, yProp: 150` from an interrupted migration, alongside valid `wProp: 0.15, hProp: 0.18`) would pass the `proportionsValid` check and return unchanged, leaving raw pixel coordinates in the proportional position fields.

`widgetNeedsProportionalMigration` already had the correct guard (`Math.abs(xProp) > 1.5` / `wProp > 1.5`). The two functions were mismatched: the guard correctly flagged the widget, but the migration function would immediately early-return without fixing it.

When these widgets were later hydrated back to pixels, `xProp: 300` was treated as a proportion (meaning 300× the safe viewport width), placing widgets wildly off-screen.

## Fix

Extended `proportionsValid` to include the same range checks used by `widgetNeedsProportionalMigration`:

```typescript
wProp <= 1.5 &&
hProp <= 1.5 &&
Math.abs(xProp) <= 1.5 &&
Math.abs(yProp) <= 1.5
```

Any field containing a pixel value (> 1.5) now falls through to the full re-migration path. Also added `typeof === 'number'` before `Number.isFinite` to satisfy TypeScript's control-flow analysis on `number | undefined` fields.

## Band-Aid Rejected

Only fixing the guard (`widgetNeedsProportionalMigration`) to not flag such widgets — that would hide the invalid state in Firestore rather than correcting it. The migration function must fix the root cause.

## Regression Tests

Added 2 regression tests and 2 happy-path arithmetic tests (`tests/utils/migrateProportionalLayout.test.ts`). 25 tests total (up from 21):

- `fixes pixel-valued xProp/yProp even when wProp/hProp are already valid proportions`: **FAIL before** (out.xProp = 300, fails `<= 1.5` assertion); **PASS after**
- `fixes pixel-valued wProp/hProp even when xProp/yProp are already valid proportions`: **FAIL before**; **PASS after**
- `converts known pixel coordinates to correct proportional values on 1920×1080`: happy-path arithmetic verification
- `round-trips pixel → proportion → pixel correctly (same viewport)`: end-to-end round-trip

## Risk Tag

**contained** — change is isolated to a single `proportionsValid` expression in `migrateProportionalLayout.ts`. The migration only runs on widgets flagged by `widgetNeedsProportionalMigration`, which already excluded healthy widgets.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J2yQ8dSA7nPUQdnvEUuKNa)_

### DIFF
--- a/tests/utils/migrateProportionalLayout.test.ts
+++ b/tests/utils/migrateProportionalLayout.test.ts
@@ -264,6 +264,110 @@ describe('migrateWidgetToProportional', () => {
     expect(out.version).toBe(5);
     expect(out.config).toBe(w.config); // structural identity preserved
   });
+
+  // Happy-path arithmetic tests: verify known pixel → proportional values.
+  it('converts known pixel coordinates to correct proportional values on 1920×1080', () => {
+    const vpW = 1920;
+    const vpH = 1080;
+    const { safeW, safeH } = getSafeViewport(vpW, vpH);
+    const w = baseWidget({ x: 200, y: 100, w: 400, h: 300 });
+    const out = migrateWidgetToProportional(w, vpW, vpH);
+    expect(out.xProp).toBeCloseTo((200 - PADDING) / safeW, 6);
+    expect(out.yProp).toBeCloseTo((100 - PADDING) / safeH, 6);
+    expect(out.wProp).toBeCloseTo(400 / safeW, 6);
+    expect(out.hProp).toBeCloseTo(300 / safeH, 6);
+    expect(out.aspectRatio).toBeCloseTo(400 / 300, 6);
+  });
+
+  it('round-trips pixel → proportion → pixel correctly (same viewport)', () => {
+    // Verifies that migrateWidgetToProportional + hydrateWidgetPixels
+    // (fill mode, same viewport) recovers the original pixel coordinates
+    // exactly (within rounding).
+    const vpW = 1920;
+    const vpH = 1080;
+    const pixelX = 200;
+    const pixelY = 100;
+    const pixelW = 400;
+    const pixelH = 300;
+    const w = baseWidget({ x: pixelX, y: pixelY, w: pixelW, h: pixelH });
+    const migrated = migrateWidgetToProportional(w, vpW, vpH);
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-06-15

### Summary
Closes a real correctness gap in `migrateWidgetToProportional`: the `proportionsValid` early-return used a `Number.isFinite`-only check, which passes for pixel values (e.g. `xProp = 300`), so a partially-migrated widget would skip re-derivation and end up wildly off-screen on hydration. Extending the check with the same `<= 1.5` range guards used by `widgetNeedsProportionalMigration` aligns the two functions.

### Code Quality
✓ No issues found. Destructuring before the comparisons to satisfy TS control-flow narrowing (`number | undefined` → `number`) is clean and the `typeof === 'number'` pre-checks are appropriate.

### Widget Standards
✓ Not applicable — pure migration utility.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. This only corrects values that hydrate back to pixels; no schema/rules impact.

### Regression Risk
✓ Low risk. The migration runs only on widgets already flagged by `widgetNeedsProportionalMigration`, which excludes healthy widgets, so the new fall-through can't touch correctly-migrated data. The round-trip test confirms pixel→proportion→pixel identity on the same viewport.

### Missing Tests
✓ Adequate coverage — 2 targeted regression tests (pixel-valued position vs. size fields) plus 2 happy-path arithmetic/round-trip tests; both regression cases assert FAIL-before/PASS-after behavior via `widgetNeedsProportionalMigration`.

### Overall
**Merge readiness:** Ready
Root-cause fix (not a guard-only band-aid) with strong, behavior-asserting tests.

---

---

PR #1988, opened 2026-06-16
ci(deploy): pin action-hosting-deploy to immutable v0.10.0

OPS-PIvers:
## What

Pin the production Firebase Hosting deploy step from the moving `@v0` tag to the immutable `@v0.10.0` release in `.github/workflows/firebase-deploy.yml`.

## Why

Supply-chain hygiene — pinning third-party actions in a deploy pipeline to an immutable tag avoids surprise changes from a moving major tag.

## Node 24 follow-up (deferred)

The recent production deploy logged a deprecation warning that this action runs on Node 20. There is **no released version that runs on Node 24 yet**: `v0.10.0` still declares `using: node20`. Upstream's Node 24 runtime bump ([PR #450](https://github.com/FirebaseExtended/action-hosting-deploy/pull/450)) is merged but **unreleased** (tracking: [issue #457](https://github.com/FirebaseExtended/action-hosting-deploy/issues/457)).

GitHub auto-runs Node 20 actions on Node 24 starting 2026-06-16, so this pin is forward-safe and the production deploy is unaffected. Re-pin to a clean Node 24 release once upstream cuts one. A code comment in the workflow documents this.

## Risk

Minimal — single-line action version pin, no change to deploy behavior. Only `firebase-deploy.yml` uses this action; the dev-deploy workflow shells out to the Firebase CLI directly.

### DIFF
--- a/.github/workflows/firebase-deploy.yml
+++ b/.github/workflows/firebase-deploy.yml
@@ -80,7 +80,12 @@ jobs:
           GOOGLE_APPLICATION_CREDENTIALS: ./service-account.json
 
       - name: Deploy to Firebase Hosting (Live)
-        uses: FirebaseExtended/action-hosting-deploy@v0
+        # Pinned to the latest immutable release instead of the moving @v0 tag.
+        # NOTE: v0.10.0 still declares Node 20. Upstream's Node 24 runtime bump
+        # (PR #450) is merged but unreleased (tracking: issue #457). GitHub
+        # auto-runs Node 20 actions on Node 24 from 2026-06-16, so this is
+        # forward-safe; re-pin to a clean Node 24 release once upstream cuts one.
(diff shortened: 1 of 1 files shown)


OPS-PIvers:
Thanks for the review! Addressed the non-blocking follow-up in 33a336e4: pinned `actions/checkout@v6` → `@v6.0.3` (the release `@v6` currently resolves to, so no behavior change). `firebase-deploy.yml`'s third-party actions are now consistently pinned — the only other `uses:` is the local `./.github/actions/setup` composite.

I kept it an immutable **version tag** rather than a commit SHA to stay consistent with the `@v0.10.0` pin in the same file. A full SHA-based hardening sweep across all workflows (including `setup-node`/`pnpm` inside the composite action) per OSSF Scorecard would be a reasonable separate follow-up if we want gold-standard pinning everywhere.

PR #2000, opened 2026-06-17
fix(videoActivityGrading): deduplicate question IDs in videoActivityMaxPoints + migration coverage

OPS-PIvers:
## Root cause

`videoActivityMaxPoints()` used a plain `reduce()` without deduplicating by question ID. `computeVideoActivityScorePct()` (numerator) was already deduplicating via a `Set<string>` to handle Drive-sync / `arrayUnion` races that can write the same question ID twice. The denominator had no such guard — so:

- numerator: counts each unique ID once ✓
- denominator: counts every occurrence → **inflated** (e.g. 4 instead of 2)
- LMS grade posted: `score / inflated_max` → fraction > 1 → rounds to wrong grade

## Fix summary

`utils/videoActivityGrading.ts` — mirror the dedup fence from `computeVideoActivityScorePct` into `videoActivityMaxPoints`:

```diff
-  return questions.reduce((sum, q) => sum + (q.points ?? 1), 0) || 100;
+  const seen = new Set<string>();
+  let total = 0;
+  for (const q of questions) {
+    if (seen.has(q.id)) continue;
+    seen.add(q.id);
+    total += q.points ?? 1;
+  }
+  return total || 100;
```

## Rejected band-aid

Deduplicating at call sites — this would have to be repeated across the grade picker and Results view; the helper exists precisely to avoid this drift.

## Regression test

New test in `tests/utils/videoActivityMaxPoints.test.ts`:

```
'deduplicates questions with the same id (Drive-sync duplicate guard)'
videoActivityMaxPoints([dup, dup])  // dup.id='x', dup.points=2
→ expected 2, old code returned 4 (FAIL before, PASS after)
```

Full suite: 5 tests pass (was 4).

## Bonus: migrateProportionalLayout coverage (commit 1)

The arithmetic in `migrateProportionalLayout.ts` is correct — no bug. However the backlog noted a coverage gap: no test for the specific case `(x=300, y=150, w=280, h=140)` on a real viewport. Three new tests added to pin the formula against concrete constants, covering 1920×1080, 1366×768, and a round-trip hydration check.

## Risk

**Contained.** Single utility function + tests. The fix only fires when duplicate question IDs are present (Drive-sync race condition). No API or schema changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_015gQT5dXu8DsCVoNhFDGFJ1)_

### DIFF
--- a/tests/utils/migrateProportionalLayout.test.ts
+++ b/tests/utils/migrateProportionalLayout.test.ts
@@ -298,6 +298,82 @@ describe('migrateWidgetToProportional', () => {
     expect(hydrated.h).toBe(pixelH);
   });
 
+  // Backlog-requested arithmetic test: the specific coordinate set cited in
+  // the PR #1976 backlog note. These exercise the division formula with a
+  // different aspect ratio (2:1) and non-trivial pixel coordinates so that
+  // a swap of safeW/safeH, or using raw vpW/vpH in place of the safe
+  // dimensions, would produce a wrong — and therefore failing — assertion.
+  it('converts (x=300, y=150, w=280, h=140) on 1920×1080 to exact proportional values', () => {
+    const vpW = 1920;
+    const vpH = 1080;
+    // safeW = vpW - PADDING*2, safeH = vpH - PADDING*2
+    const { safeW, safeH } = getSafeViewport(vpW, vpH);
+    const w = baseWidget({ x: 300, y: 150, w: 280, h: 140 });
+    const out = migrateWidgetToProportional(w, vpW, vpH);
+
+    // Position: pixel origin is offset by PADDING from the safe-board origin.
+    const expectedXProp = (300 - PADDING) / safeW;
+    const expectedYProp = (150 - PADDING) / safeH;
+    // Size: proportioned against the full safe dimension (no padding offset).
+    const expectedWProp = 280 / safeW;
+    const expectedHProp = 140 / safeH;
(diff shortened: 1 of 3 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-06-17

### Summary
Fixes an LMS grading bug: `videoActivityMaxPoints()` (the denominator) summed question points without de-duplicating by id, while `computeVideoActivityScorePct()` (the numerator) already deduped via a `Set`. Drive-sync/`arrayUnion` races that write a question id twice inflated the denominator, producing fractions >1 that round to the wrong grade. Adds a matching `Set<string>` dedup fence plus migration-coverage tests.

### Code Quality
✓ No issues found. The new loop mirrors the existing numerator dedup exactly, keeping numerator and denominator on the same counting basis; `|| 100` fallback preserved.

### Widget Standards / Admin Config
✓ Not applicable.

### Firestore & Security
✓ No issues found — pure utility logic; no schema or rules change.

### Regression Risk
✓ Low risk — single function; new behavior only differs when duplicate ids are present (the race condition). Existing tests updated to give questions distinct ids so the sum cases still validate.

### Missing Tests
✓ Adequate — adds the dedup regression (`[dup, dup]` → 2, was 4) and three `migrateProportionalLayout` arithmetic/round-trip tests across two viewports, closing the cited coverage backlog item.

### Overall
**Merge readiness:** Ready
Clean, well-targeted fix with strong regression coverage.

---

---

PR #2005, opened 2026-06-17
test(hooks): cover usePresetSubEmails preset-sub-email listener

OPS-PIvers:
## Summary

Adds `tests/hooks/usePresetSubEmails.test.ts` (11 tests) covering `hooks/usePresetSubEmails.ts`, the `/preset_sub_emails/{buildingId}` single-doc Firestore listener that feeds the ShareLinkCreatorModal's sub-email picker. This was an untested session hook tracked under the **HIGH** "hooks/ coverage — session hooks still missing tests" item in `docs/scheduled-tasks/test-coverage.md`.

## Coverage

- **Listener wiring** — targets `doc(db, 'preset_sub_emails', canonical)`, loading-true until first snapshot, canonicalizes a legacy building ID (`orono-high-school` → `high`, using the real `canonicalBuildingId` so the alias path is verified end-to-end), and short-circuits (no listener) on an empty `buildingId`.
- **Snapshot mapping** — exposes the `emails` array and clears loading; resolves a missing doc / non-array `emails` field to an empty list; filters non-string entries out of the array.
- **Error handling** — logs `[usePresetSubEmails] snapshot error:` and resolves to `{emails:[], loading:false}`.
- **Building change & cleanup** — re-subscribe tears down the prior listener and re-enters loading; a legacy alias that canonicalizes to the same building keeps the cached snapshot resolved; unmount fires the unsub.

Mocks `firebase/firestore` (`doc`/`onSnapshot`) and `@/config/firebase` per the existing `useSubstituteShares.test.ts` reference pattern.

## Verification

- `pnpm exec vitest run tests/hooks/usePresetSubEmails.test.ts` → 11 passing
- `pnpm exec tsc --noEmit` — clean
- `pnpm exec eslint tests/hooks/usePresetSubEmails.test.ts --max-warnings 0` — clean
- `pnpm exec prettier --check` — clean

Test-only change; no production code touched.

---
_Generated by [Claude Code](https://claude.ai/code/session_014GHcnhN65f2qNX6UQfFmRb)_


utils/migrateProportionalLayout.ts
```diff
@@ -75,11 +75,31 @@ export const migrateWidgetToProportional = (
...
+    typeof xProp === 'number' &&
+    typeof yProp === 'number' &&
+    typeof wProp === 'number' &&
+    typeof hProp === 'number' &&
+    Number.isFinite(xProp) &&
+    Number.isFinite(yProp) &&
+    Number.isFinite(wProp) &&
+    Number.isFinite(hProp) &&
+    wProp <= 1.5 &&
+    hProp <= 1.5 &&
```
OPS-PIvers: **Dimension props skip `Math.abs`, leaving a gap for negative corrupt values**

The position checks correctly wrap with `Math.abs`:
```ts
Math.abs(xProp) <= 1.5 &&
Math.abs(yProp) <= 1.5
```

But the dimension checks don't:
```ts
wProp <= 1.5 &&   // ← misses negative pixel values
hProp <= 1.5 &&   // ← misses negative pixel values
```

If Firestore data is corrupt with a negative pixel-valued `wProp` (e.g. `-400`), then `-400 <= 1.5` is `true`, so `proportionsValid` becomes `true` and the function returns early without fixing the bad field. The widget would be permanently stuck with an invalid negative dimension.

`widgetNeedsProportionalMigration` has the same gap at lines 35-36 (`w.wProp > 1.5` doesn't catch `-400` either), so the entire migration chain would skip a widget with a negative dimension prop. Both functions should use `Math.abs` for defense-in-depth:

```ts
// migrateWidgetToProportional
Math.abs(wProp) <= 1.5 &&
Math.abs(hProp) <= 1.5 &&

// widgetNeedsProportionalMigration (line 35-36)
Math.abs(w.wProp) > 1.5 ||
Math.abs(w.hProp) > 1.5 ||
```

Negative widget dimensions shouldn't occur through normal use, but corrupt Firestore documents are possible, and this is a cheap fix.

---

OPS-PIvers (commented):
Automated nightly review. One issue found across the diff.

---

utils/migrateProportionalLayout.ts
OPS-PIvers: Fixed in 77de6548. Wrapped the `wProp`/`hProp` range guards in `Math.abs` in both `migrateWidgetToProportional` (the `proportionsValid` early-return) and `widgetNeedsProportionalMigration`, for parity with the position checks. Added detection + repair regression tests covering a corrupt negative dimension (`wProp: -400`) — the widget is now flagged and re-derived into a valid positive proportion instead of being stranded. `tsc`, eslint, prettier, and the full `migrateProportionalLayout` suite (27 tests) all pass.

---

PR #2015, opened 2026-06-18
perf(widgets): add canvas-store selection/read-only selectors, migrate 8 widgets

OPS-PIvers:
## Summary

Fourth slice of the `DashboardContext` churn-reduction effort (after F9 [#1996](https://github.com/OPS-PIvers/SpartBoard/pull/1996), content-actions [#2004](https://github.com/OPS-PIvers/SpartBoard/pull/2004), globalStyle [#2008](https://github.com/OPS-PIvers/SpartBoard/pull/2008)).

Widgets that read `selectedWidgetId` or `isActiveBoardReadOnly` off the legacy `useDashboard()` value re-render on **every** unrelated dashboard op, because that context value re-creates on each commit. Both fields are already canvas-store slice fields, so this PR exposes them as mount-stable selectors and migrates the widgets that were blocked *only* on them.

## What changed

- **`context/dashboardCanvasStore.ts`** — two named boolean convenience hooks (mirroring the `useGlobalStyle` precedent):
  - `useIsActiveBoardReadOnly()` = `useDashboardCanvasSelector(s => s.isActiveBoardReadOnly)`
  - `useIsWidgetSelected(id)` = `useDashboardCanvasSelector(s => s.selectedWidgetId === id)`

  `useIsWidgetSelected` returns the **predicate**, not the raw `selectedWidgetId` — so a widget re-renders only when *its own* selected-ness flips; selecting a different widget bails via the `Object.is` snapshot cache. Both return primitives (required for the dedupe). No new provider/context; the selector's built-in legacy fallback already covers bare hosts (Subs/Student/tests).
- **8 widgets migrated** off `useDashboard()`: ConceptWeb, GraphicOrganizer (parent + its `EditableNode` sub-component), TextWidget, WorkSymbols, MaterialsWidget, SoundboardWidget, Webcam (pure actions), ActivityWall (actions + read-only).
- **4 colocated tests** updated to mock the stable surfaces (keeping the legacy `useDashboard` mock where an unmigrated child — `TextSettings`, ActivityWall's `ShareModal` — still consumes it).
- **Perf harness** — new `SelectionProbe` and `ReadOnlyProbe` families.

## Proof (perf harness, 17 scenarios)

**SelectionProbe** (5×, bound to an untouched widget `w-8`):

| scenario | selection probe |
|---|---|
| `bringToFront5` | **0** |
| `drag.release` | **0** |
| `selectOther` (select a *different* widget) | **0** |
| `selectSelf` (select `w-8`) | **5** |

→ true per-instance isolation: only the widget that actually becomes selected re-renders.

**ReadOnlyProbe** (5×):

| scenario | read-only probe |
|---|---|
| `bringToFront5` / `drag.release` | **0** |
| `loadReadOnlyBoard` (push a viewer board) | **5** |

**Falsifiability:** pointing `SelectionProbe` at the raw legacy `selectedWidgetId` makes `selectOther=5`, `bringToFront5=25`, `drag.release=5` — the isolation assertions fail. Both probe families have real positive controls so the 0-deltas can't pass trivially.

## Out of scope (next levers)

Remaining blocked widgets need surfaces this slice doesn't add: `activeDashboard.widgets`/`.background`/`.id` (cross-widget & background sync — Schedule, SoundWidget, NextUp, QR, Music, Drawing, …) and roster reads (`rosters`/`activeRosterId` — Checklist, LunchCount, Stations, SeatingChart, …). A `widgets`-selector surface and a roster surface would be the next two slices; Weather additionally needs `setBackground` added to the actions surface.

## Verification

- `pnpm run lint` / `prettier --check` on all changed files — clean
- Full unit suite green (the one editor-perf timeout in the run was a load-induced flake — passes 3/3 in isolation, unrelated to this change)
- Dashboard perf harness — all 17 scenarios pass

### DIFF
--- a/components/widgets/ActivityWall/Widget.test.tsx
+++ b/components/widgets/ActivityWall/Widget.test.tsx
@@ -64,11 +64,23 @@ const {
 
 let snapshotDocs: Record<string, unknown>[] = [];
 
-vi.mock('@/context/useDashboard', () => ({
-  useDashboard: () => ({
+vi.mock('@/context/dashboardCanvasStore', () => ({
+  useDashboardActions: () => ({
     addWidget: mockAddWidget,
     addToast: mockAddToast,
     updateWidget: mockUpdateWidget,
+  }),
+  // Tests don't set up a read-only board, so the widget renders in its
+  // default editable state (matching the prior mock, which omitted the flag).
+  useIsActiveBoardReadOnly: () => false,
+}));
+
+// ActivityWallShareModal (a child still on the legacy useDashboard() value)
+// reads `addToast` when the live view mounts the share UI, so keep the legacy
+// mock for that unmigrated component.
+vi.mock('@/context/useDashboard', () => ({
+  useDashboard: () => ({
+    addToast: mockAddToast,
   }),
 }));
 
--- a/components/widgets/ActivityWall/Widget.tsx
+++ b/components/widgets/ActivityWall/Widget.tsx
@@ -26,7 +26,10 @@ import {
   ActivityWallSubmission,
   ClassLinkClass,
 } from '@/types';
-import { useDashboard } from '@/context/useDashboard';
+import {
+  useDashboardActions,
+  useIsActiveBoardReadOnly,
+} from '@/context/dashboardCanvasStore';
 import { useAuth } from '@/context/useAuth';
 import { classLinkService } from '@/utils/classlinkService';
 import {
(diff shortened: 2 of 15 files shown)


OPS-PIvers:
Addressed the orthogonality suggestion in 02774740 — added `expect(metricFor('selectOther').readOnlyProbeRenders).toBe(0)` and the same for `selectSelf`, making explicit that a selection change never churns the read-only probe (the converse of `loadReadOnlyBoard`). The values were already 0 in the baseline; this just asserts them. Thanks!

PR #2041, opened 2026-06-21
perf(page-load): code-split teacher shell + lazy-init deep-link import hooks

OPS-PIvers:
## Summary

Two behavior-identical page-load optimizations for the signed-in teacher home (`DashboardView`), continuing the work from #2039 (harness) and #2040 (safe quick wins).

1. **Opt 1 — Code-split `Dock` + `Sidebar` out of the synchronous teacher mount.** Both are now `React.lazy` behind a Suspense boundary, so the board canvas paints first and the shell streams in on the next microtask. The fallback reserves the dock footprint (only when a dock will actually render — annotation/read-only boards show no skeleton). Both are `position:fixed` overlays, so deferring them a tick **cannot** shift the board's layout.
2. **Opt 2 — Lazy-init `DashboardView`'s deep-link import hooks.** The five Firestore‑listener hooks `useQuiz`, `useQuizAssignments`, `useVideoActivity`, `useVideoActivityAssignments`, `usePlcs` ran on **every** teacher load but DashboardView consumed **only** their import/peek callbacks (never their list data — the Quiz/VideoActivity widgets run their own independent listeners). They + their callbacks/effects/modal moved into a new `DeepLinkShareImporter`, mounted lazily only once a pending share id appears. On the common load, **zero** of those listeners are opened.

## Measurements — `pageLoadPerf.test.tsx`, `DashboardView`

Medians of the per-run medians (6 iterations/run × 3 runs); jsdom timings are noisy (±~2ms).

| Stage | `firstCommitMedianMs` (time-to-first-paint) | `medianMountMs` (settle-window sum) | commits |
|---|---|---|---|
| **Baseline** (origin/dev-paul) | ~16.3 (== total; single commit) | ~16.3 | 1 |
| **+ Opt 1** | **~6.5** | ~6.6 | 2 |
| **+ Opt 1 + Opt 2** | **~6.0** | ~6.1 | 2 |

**Which metric moved, and why:**
- **Opt 1 targets perceived load (time-to-first-paint).** I extended the harness to record `firstCommitMedianMs` (the first/synchronous commit) separately — this is the sanctioned proof (option *a*). It drops **~16.3ms → ~6.5ms (~57%)**: the board canvas paints without waiting on the Sidebar/Dock trees, which now land in a second commit. `firstCommitMedianMs` is stable regardless of the settle window (verified at 30ms and 200ms).
- **About `medianMountMs` for Opt 1:** the harness *sums* every commit in the settle window, so in a **real browser** total CPU work is unchanged (the Sidebar/Dock render is deferred past first paint, not removed). The apparent drop in the harness's sum is partly a jsdom artifact (the `React.lazy` dynamic import resolves at/after the 30ms settle boundary, so part of the deferred commit lands outside the window). I left the 30ms settle untouched (it's the harness-of-record's deterministic baseline). **Honest read: Opt 1's real win is first-paint, not total work.**
- **Opt 2's harness movement is marginal (within noise).** Expected: the harness *mocks* all five hooks, so their real cost — five live `onSnapshot` subscriptions — was never present in the harness to begin with. **The real win (five Firestore listeners removed from every common teacher load, plus the import-machinery callbacks/effects no longer running at mount) is a genuine cost reduction not visible to the mocked harness.**

## Threshold note (jsdom noise)
`firstCommitMedianMs` for DashboardView lands at **~6–7ms**, robustly under the 10ms guideline across all runs. It is **not** under 5ms, and given ±~2ms jsdom jitter I would not claim a robust <5ms. The remaining ~6ms first commit is the board-canvas shell + the ~10 other always-on hooks, outside this PR's scope.

## Behavior / cost safety
- **No new Firestore/Storage/AI usage.** Strictly *fewer* listeners on the common path; the import code paths issue exactly the same reads/writes as before.
- **Opt 2 preserves the import flow exactly:** the importer's effects fire on mount with the same pending id present, clear it synchronously (the established triple-import-race guard), and surface the same toasts / sync-vs-copy mode picker. The mount is **latched** because imports clear their pending id synchronously — a bare presence gate would tear the importer down mid-import.
- **Opt 1 keeps the global keydown listener and the `useGesture` surface on `DashboardView`,** so keyboard shortcuts and pan/zoom are unaffected by the shell mounting a tick later. No focus loss (Sidebar/Dock register their own listeners on mount), no shortcut breakage, no board layout jump.
- The self-referential Retry callbacks route recursion through refs synced in a `useEffect` (the `useQuizAssignments` `assignmentsRef` pattern) — required because the small extracted component is fully analyzed by the React-Compiler `react-hooks` lint rules (the original 1900-line DashboardView bailed out of that analysis).

## Validation
- `type-check`, `lint` (`--max-warnings 0`), `format:check` — all clean repo-wide.
- Full unit suite: **5711 passed** (the one initially-failing `DashboardView.test.tsx` "renders correctly" assertion was updated to `await findByTestId` for the now-lazy Sidebar/Dock).
- Page-load harness: **all 25 pages green**; only DashboardView's profile changed (commits 1→2).

https://claude.ai/code/session_01KMtn3eXsqqQ3fUSWT5SUAt

---
_Generated by [Claude Code](https://claude.ai/code/session_01KMtn3eXsqqQ3fUSWT5SUAt)_


OPS-PIvers:
Thanks — I evaluated this and it can't be applied here. Assigning these refs in the render body is exactly what I tried first, and the `react-hooks/refs` ESLint rule rejects it:

```
components/layout/DeepLinkShareImporter.tsx
  Cannot update ref during render   react-hooks/refs
```

That's a hard error under our `pnpm run lint --max-warnings 0` gate (the "Code Quality Checks" job), so the suggestion would fail CI.

The reason the CLAUDE.md "assign refs in the render body" guideline doesn't apply in *this* file: the React-Compiler-based `react-hooks` rules fully analyze this small, self-contained component and forbid render-phase ref writes. The large `DashboardView` (and other places that do `zoomRef.current = zoom` in render) bail out of that analysis, so the rule never fires there. For analyzed components the repo's own established pattern is to sync the ref in an effect — e.g. `assignmentsRef` in `hooks/useQuizAssignments.ts`, whose comment notes it's updated "via `useEffect` (rather than during render) to keep the `react-hooks/refs` lint happy."

No stale-closure risk here: `peekAndDispatchImportRef`/`runVideoActivityImportRef` are dereferenced **only** from the deferred "Retry" toast/modal `onClick` handlers, which can only fire after a user interaction — long after the effect (deps `[peekAndDispatchImport, runVideoActivityImport]`) has committed the latest callback. There is no code path that reads the ref in the render→effect-commit window. Leaving as-is.

---

OPS-PIvers:
Thanks — all three addressed in 8002c05.

**1. Measurement gap / wrong committed artifact (the real catch).** You were reading a stale artifact: the committed `page-load-baseline.json` had accidentally captured my *baseline* measurement run (a temporary run where I'd swapped in the pre-split DashboardView to record the baseline first-commit number), which is why it showed `commits: 1` / `18.442`. Regenerated against the optimized code — it now correctly shows the post-split profile:

```json
"commits": 2,
"medianMountMs": 6.846,
"firstCommitMedianMs": 6.694,
```

So jsdom *does* produce the 2-commit split with the optimized code; `commits: 2` confirms Sidebar/Dock land in a second commit. The first-paint comparison is baseline ~18ms (single commit) → ~6.7ms first commit.

On the threshold suggestion: I deliberately did **not** add a `<= X ms` assertion. The harness's founding contract (#2039) is "asserts only that metrics were produced — NO duration thresholds, so this can never be flaky on slow CI machines." `firstCommitMedianMs` is a recorded-for-human-comparison metric with exactly the same status as the existing `medianMountMs` (compare medians of 3 runs) — not a CI-enforced gate. Adding a per-page threshold would contradict that design and reintroduce the CI-flakiness it was built to avoid. The metric's value is the committed before/after artifact + the PR table, not an inequality assertion.

**2. `plcsLoading` deep-link latency.** Correct and behaviour-preserving as you noted — documented it inline at the `usePlcs()` call so the one-time extra snapshot round-trip on URL-param deep links is explicit. It's the intended trade for not opening that listener on every common teacher load.

**3. Stale `logError` tags.** Retagged all three `'DashboardView.*'` → `'DeepLinkShareImporter.*'`.

---

OPS-PIvers:
Good catch — fixed in 326b846. The cancel-path Retry now routes through `peekAndDispatchImportRef.current(shareId)`, matching the catch-path so a Retry clicked after cancel always uses the freshest dispatcher / `plcs` snapshot rather than the one captured when the toast was created.

---

PR #2092, opened 2026-06-27
fix(NumberLine): Escape now cancels edits in min/max/step inputs without saving

OPS-PIvers:
## Root Cause

`components/widgets/NumberLine/Settings.tsx` had three uncontrolled `<input type="number">` elements (Min Value, Max Value, Step) with `onBlur` save callbacks but no Escape cancel handler. In real browsers, pressing Escape while an input has focus triggers a native `blur` event synchronously. That blur fired `onBlur`, which called `updateConfig` with whatever the teacher had typed — silently persisting the cancelled value to Firestore.

This is the same stale-onBlur pattern previously fixed in DraggableWindow (#1965), RandomGroups (#1974), FolderTree/FolderSidebar (#1975), and PollWidget (#2064).

## Fix Summary

- Added `minCancelledRef`, `maxCancelledRef`, `stepCancelledRef` (`useRef(false)`) — one per input, assigned synchronously in render body.
- Added an `Escape` branch to each input's `onKeyDown` handler: sets the ref to `true`, resets `e.currentTarget.value` to the original prop value, then calls `e.currentTarget.blur()`.
- Each `onBlur` handler now checks its ref on entry — if `true`, resets the ref to `false` and returns early (skipping `updateConfig`). Normal blur/Enter saves are unaffected.
- Added `aria-label` attributes to all three inputs for accessibility and testability.

## Rejected Band-Aid

Wrapping the Escape key handler in a `useEffect` or debounce would hide the race but not fix it — the browser's synchronous blur fires before any async cleanup can run. The ref flag must be set synchronously before the blur, which is what this fix does.

## Regression Test

`components/widgets/NumberLine/Settings.test.tsx` — 10 new tests covering all three inputs:
- Escape does NOT call `updateWidget`
- DOM value reverts to original after Escape
- Normal blur after a previous Escape still saves correctly (ref reset confirmed)

**Before fix:** Test fails — `getByLabelText('Min Value')` throws because the original code had no `aria-label`; even if patched past that, `updateWidget` is called on Escape+blur.  
**After fix:** All 10 tests pass.

> Note: `tsc` strict mode and `eslint-typescript` disagree on the return type of `screen.getByLabelText()` (`HTMLElement` vs inferred `HTMLInputElement`). Three `// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion` comments resolve the conflict at the `.value`-accessing declarations. Both `pnpm run type-check` and `pnpm exec eslint ... --max-warnings 0` pass cleanly.

## Risk

**Contained** — change is limited to three `onKeyDown` handlers and three `onBlur` guards in a single settings component. No state shape changes, no Firestore schema changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_013MxYdKYysLuXL5x5mygD1V)_

### DIFF
--- a/components/widgets/NumberLine/Settings.test.tsx
+++ b/components/widgets/NumberLine/Settings.test.tsx
@@ -104,6 +104,7 @@ describe('NumberLineSettings — Escape-cancel regression', () => {
   it('resets the Min Value input to the original value after Escape', () => {
     render(<NumberLineSettings widget={baseWidget} />);
 
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
     const input = screen.getByLabelText('Min Value') as HTMLInputElement;
 
     fireEvent.change(input, { target: { value: '99' } });
@@ -165,6 +166,7 @@ describe('NumberLineSettings — Escape-cancel regression', () => {
   it('resets the Max Value input to the original value after Escape', () => {
     render(<NumberLineSettings widget={baseWidget} />);
 
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
     const input = screen.getByLabelText('Max Value') as HTMLInputElement;
 
     fireEvent.change(input, { target: { value: '999' } });
@@ -225,6 +227,7 @@ describe('NumberLineSettings — Escape-cancel regression', () => {
   it('resets the Step input to the original value after Escape', () => {
     render(<NumberLineSettings widget={baseWidget} />);
 
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
     const input = screen.getByLabelText('Step (Interval)') as HTMLInputElement;
 
     fireEvent.change(input, { target: { value: '5' } });
--- a/components/widgets/NumberLine/Settings.test.tsx
+++ b/components/widgets/NumberLine/Settings.test.tsx
@@ -1,24 +1,4 @@
-/**
- * Regression tests for the Escape-cancel bug in NumberLineSettings.
- *
- * Root cause: The min, max, and step inputs used `defaultValue` (uncontrolled)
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-06-27

### Summary
Adds an Escape-cancel path to the three uncontrolled number inputs (Min/Max/Step) in `NumberLine/Settings.tsx` so pressing Escape reverts the DOM value and skips the `onBlur` save. Well-scoped fix of a real data-persistence bug, consistent with the prior stale-onBlur fixes (#1965/#1974/#1975/#2064).

### Code Quality
✓ The `*CancelledRef` flags are assigned synchronously in the render body / event handlers — no `useEffect` misuse, matching the CLAUDE.md guidance. The three `@typescript-eslint/no-unnecessary-type-assertion` disables are justified (tsc vs eslint-typescript disagree on `getByLabelText` return type) and scoped to single lines. No console.log / floating promises.

### Widget Standards
✓ Not applicable — this is a back-face settings panel, which does not require container-query scaling. Adding `aria-label` to each input is a nice accessibility/testability improvement.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No data-model, rules, or auth changes. Save path semantics unchanged for normal blur/Enter.

### Regression Risk
✓ Low. Does not touch `WidgetRegistry.ts`, `types.ts`, or `DashboardContext.tsx`. The early-return guard resets the ref so a subsequent real edit still saves (explicitly tested).

### Missing Tests
✓ Adequate coverage — 10 tests covering Escape-no-save, DOM revert, save-after-prior-Escape, and the Enter path, across all three inputs.

### Overall
**Merge readiness:** Ready
Contained, correct, and well-tested; behavior matches the established pattern.

---

---

PR #2137, opened 2026-07-04
fix(state): gradeAnswer Ordering partial credit contradicted isCorrect with pointsEarned

OPS-PIvers:
## Root cause

For `Ordering` questions with `allowPartialCredit: true`, `gradeAnswer` (`hooks/useQuizSession.ts`) computes `pointsEarned` from `longestOrderedSubsequenceLength` (an LIS over the given vs. correct item arrays), but `isCorrect` was still derived from strict whole-string equality (`correct === given`) computed once at the top of the function, before the partial-credit fork. The two formulas diverge whenever the student's stored answer has an item that isn't in the current `correctAnswer` — most plausibly when a teacher edits a question (removes an item) after a student has already submitted a full answer under the old item set. `longestOrderedSubsequenceLength` simply skips the now-unknown item and can still report a full-length subsequence over the remaining items, so `pointsEarned` reaches `max` while `isCorrect` stays `false` — the same "isCorrect contradicts pointsEarned" bug class already fixed for Matching partial credit in #1950, but never mirrored to Ordering.

## Fix summary

`isCorrect` for the partial-credit branch now derives from the same `lis === correctItems.length` formula that produces `pointsEarned`, so the invariant `pointsEarned >= pointsMax ⟺ isCorrect` holds unconditionally for this function — matching the pattern already established for Matching.

## Band-aid rejected

Special-casing the export/publish paths to re-derive `isCorrect` from `pointsEarned >= pointsMax` downstream would paper over the contradiction at the display layer while leaving the pure grading function internally inconsistent for any other caller. Fixed at the source instead.

## Regression test

`tests/hooks/useQuizSession.test.ts` — "Ordering: isCorrect stays consistent with pointsEarned when the given answer has an extra item not in correctAnswer".

**Before fix:** fails (`{isCorrect: false, pointsEarned: 3, pointsMax: 3}`).
**After fix:** passes (106/106 tests in that file green); did not disturb any of the 5 existing Ordering partial-credit tests.

## Validation evidence

- `pnpm exec eslint context hooks utils types.ts --max-warnings 0` — clean
- `pnpm run format:check` — clean
- Full `pnpm run validate` on this branch: type-check:all, lint, format:check, root tests (551 files / 5986 tests passed), functions tests (37 files / 703 tests passed) — all green

## New backlog candidates (not fixed here)

- `utils/snapshotThrottle.ts` (`createLeadingTrailingThrottle`, used by `useQuizSession.ts` and `useVideoActivitySession.ts`) has no dedicated test file despite being shared, non-trivial state-sync infrastructure.
- The "partial-credit `isCorrect` must equal `pointsEarned >= pointsMax`" invariant (now fixed twice: Matching #1950, Ordering this run) should be spot-checked against any future new partial-credit question type before it ships.

## Risk

**Contained** — isolated to the Ordering partial-credit grading branch; no changes to the non-partial-credit path or to Matching.

---
🤖 Part of the nightly code-health routine (`docs/routines/debugger.md`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01DcK2bcoJQQTUyoJaZna31m)_

### DIFF
--- a/hooks/useQuizSession.ts
+++ b/hooks/useQuizSession.ts
@@ -490,7 +490,7 @@ export function gradeAnswer(
     const lis = longestOrderedSubsequenceLength(correctItems, givenItems);
     const pointsEarned =
       correctItems.length === 0 ? 0 : (lis / correctItems.length) * max;
-    // isCorrect must share pointsEarned's lis-based formula, not the strict whole-string equality above (mirrors Matching's fix in #1950).
+    // isCorrect must share pointsEarned's lis-based formula, not the strict whole-string equality above.
     const partialIsCorrect =
       correctItems.length > 0 && lis === correctItems.length;
     return { isCorrect: partialIsCorrect, pointsEarned, pointsMax: max };
--- a/hooks/useQuizSession.ts
+++ b/hooks/useQuizSession.ts
@@ -490,18 +490,7 @@ export function gradeAnswer(
     const lis = longestOrderedSubsequenceLength(correctItems, givenItems);
     const pointsEarned =
       correctItems.length === 0 ? 0 : (lis / correctItems.length) * max;
-    // Partial-credit isCorrect must be derived from the SAME formula as
-    // pointsEarned (lis === correctItems.length), not from the strict
-    // whole-string equality used above for the non-partial branch. The two
-    // diverge whenever `given` contains an item outside `correctItems` — the
-    // ordering UI never lets a student add items itself, but a question
-    // edited (items removed) after a student already submitted an answer
-    // leaves exactly that shape: `longestOrderedSubsequenceLength` skips the
-    // now-unknown item and still reports a full-length subsequence over the
-    // remaining ones, so `pointsEarned` reaches `max` while `correct===given`
-    // stays false on the now-shorter string. Reusing the strict `isCorrect`
-    // here produced the same isCorrect=false/pointsEarned=max contradiction
-    // already fixed for Matching partial credit (#1950).
+    // isCorrect must share pointsEarned's lis-based formula, not the strict whole-string equality above (mirrors Matching's fix in #1950).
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-04

### Summary
Fixes an `Ordering` partial-credit inconsistency where `isCorrect` was derived from strict whole-string equality while `pointsEarned` came from the LIS formula, so a full-credit answer could still report `isCorrect: false` (mirrors the Matching fix in #1950). Correct, minimal, and well-targeted.

### Code Quality
✓ Logic is correct — `partialIsCorrect = correctItems.length > 0 && lis === correctItems.length` now shares the exact denominator that produces `pointsEarned`, so the `pointsEarned >= pointsMax ⟺ isCorrect` invariant holds.

⚠️ **Minor (style):** the new 11-line comment block above the `return` in `hooks/useQuizSession.ts` exceeds the CLAUDE.md rule *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max"* — the same rule the Claude reviewer enforced on the sibling PRs #2135 and #2134 this run. The `it(...)` test comment has the same issue. Consider trimming both to a single line (the PR description already carries the full rationale). Not a merge blocker.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No changes.

### Regression Risk
✓ Low — isolated to the Ordering partial-credit branch; the non-partial path and Matching are untouched, and the existing 5 Ordering partial-credit tests still pass alongside the new case.

### Missing Tests
✓ Adequate — a targeted regression test covering the extra-item scenario was added.

### Overall
**Merge readiness:** Ready with minor notes. Correctness is solid; only the comment-block length is worth trimming for consistency with the sibling PRs.

---

---

PR #2147, opened 2026-07-05
fix(TimeTool): clamp hold-to-ramp duration to the shared ceiling

OPS-PIvers:
## Summary

**Root cause:** `TimeTool` has three ways to set a duration, and only two respected `TIME_TOOL_MAX_DURATION_SECONDS` (59999s = 999m59s): the Keypad (bounded by its 3-digit minutes field) and the admin building-defaults validator (`utils/adminBuildingConfig.ts`, which explicitly clamps to avoid overflowing the timer readout). The third path — `adjustTime` in `useTimeTool.ts`, driven by the +/- hold-to-ramp buttons (`useHoldAccelerate`, which ramps to a 5x tick multiplier while held) — had no upper bound: `Math.max(0, current + deltaSeconds)`. Holding "+" long enough pushes `elapsedTime`/`duration` past the app's own declared ceiling.

**Fix (in three parts, the last two added during review):**
1. Clamp both `next` and `nextDuration` in `adjustTime` to `TIME_TOOL_MAX_DURATION_SECONDS`, the same constant the Keypad and admin validator already anchor to.
2. Add a matching `disabled={displayTime >= TIME_TOOL_MAX_DURATION_SECONDS}` guard on the "+" `AdjustButton`, symmetric with the existing floor guard on "-" (`disabled={displayTime <= 0}`) — caught by review: without it, the button stayed clickable at the ceiling.
3. Add a no-op early-return in `adjustTime` when the clamped result doesn't change `elapsedTime`/`duration` — caught by review: while the timer is *running* at the ceiling, each write resets `startTime`, which restarts the RAF display loop and drops `runningDisplayTime` just below the ceiling within one frame, transiently re-enabling the button before the next 250ms hold-tick — without the guard this causes a Firestore write storm and a visually "frozen" display oscillating at the ceiling.

**Band-aid rejected:** Clamping only at the UI/display layer (disabling "+" near the ceiling, or capping the rendered string) — rejected because the persisted Firestore `duration`/`elapsedTime` values could still drift past the ceiling via this same code path (keyboard Enter/Space also calls `adjustTime`), repeating the "mask a bad stored value at display time" anti-pattern this codebase has fixed before.

**Risk:** contained.

## Test plan
- [x] `TimeToolWidget.test.tsx`: "clamps elapsedTime/duration to TIME_TOOL_MAX_DURATION_SECONDS instead of overflowing past it" — fails on unpatched code (60059 instead of 59999), passes after.
- [x] `TimeToolWidget.test.tsx`: "disables the '+' button once elapsedTime/duration reach the ceiling" — asserts the button is `disabled` and clicking it issues no write.
- [x] `TimeToolWidget.test.tsx`: "adjustTime is a no-op when running at the ceiling, avoiding a write-storm/startTime-reset loop" — uses `renderHook` to call `adjustTime` directly (bypassing the UI `disabled` gate, which would otherwise mask this path) while running at the ceiling; fails without the no-op guard (1 write), passes with it (0 writes).
- [x] All three verified independently by the orchestrator (revert → confirm fail → restore → confirm pass).
- [x] `pnpm run validate` — clean.
- [x] `pnpm run build` — clean.

🤖 Generated by the nightly SpartBoard code-health routine (subagent-authored, orchestrator-reviewed; two follow-up fixes from claude[bot] review incorporated).

### DIFF
--- a/components/widgets/TimeTool/TimeToolWidget.test.tsx
+++ b/components/widgets/TimeTool/TimeToolWidget.test.tsx
@@ -1,10 +1,18 @@
 import { describe, it, expect, vi, beforeEach, afterEach, Mock } from 'vitest';
-import { render, screen, fireEvent, cleanup } from '@testing-library/react';
+import {
+  render,
+  renderHook,
+  screen,
+  fireEvent,
+  cleanup,
+  act,
+} from '@testing-library/react';
 import { renderToStaticMarkup } from 'react-dom/server';
 import { useDashboard } from '@/context/useDashboard';
 import { useGlobalStyle } from '@/context/dashboardCanvasStore';
 import { WidgetData, TimeToolConfig, DEFAULT_GLOBAL_STYLE } from '@/types';
 import { TimeToolWidget } from './TimeToolWidget';
+import { useTimeTool } from './useTimeTool';
 import { DashboardContextValue } from '@/context/DashboardContextValue';
 
 // useTimeTool still reads the legacy context (updateWidget + activeDashboard);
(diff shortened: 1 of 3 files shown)


OPS-PIvers:
Good catch — added the missing `disabled={displayTime >= TIME_TOOL_MAX_DURATION_SECONDS}` guard on the "+" button, symmetric with the existing floor guard on "-". Updated the regression test to exercise the clamp math from below the ceiling (59980→59999) since the button is now disabled at the ceiling itself, and added a second test asserting the disabled state + that `updateWidget` isn't called when clicked there. `pnpm run validate` and `pnpm run build` both green.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-05

### Summary
Clamps the third TimeTool duration-entry path — `adjustTime` (the +/- hold-to-ramp buttons, which ramp to a 5× tick) — to `TIME_TOOL_MAX_DURATION_SECONDS`, closing an overflow the Keypad and admin validator already guarded against. Both `next` and `nextDuration` are clamped, so the persisted Firestore values can't drift past the ceiling.

### Code Quality
✓ No issues found. Anchoring to the shared `TIME_TOOL_MAX_DURATION_SECONDS` constant (rather than a local literal) keeps the three entry paths consistent.

### Widget Standards
✓ Not applicable — logic/state change only, no front-face scaling or registry surface touched.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found.

### Regression Risk
✓ Low. One thing worth a glance: `nextDuration = Math.min(MAX, Math.max(config.duration, next))` now also clamps a *pre-existing* over-cap `config.duration` down to the cap on any adjust — desirable (self-healing) but a behavior change for legacy over-cap docs. The `-` path relies on the existing `Math.max(0, …)` floor, which is fine.

### Missing Tests
✓ Adequate — the diff adds both a clamp test (overflow → 59999) and a UI test (the `+` button is `disabled` at the ceiling). Note the PR body describes only the `adjustTime` clamp; the diff also adds the `AdjustButton disabled={displayTime >= MAX}` UI guard and its test — a reasonable defense-in-depth pairing worth reflecting in the description.

### Overall
**Merge readiness:** Ready
Root-cause clamp at the persistence layer plus a UI affordance; well covered.

---

---

OPS-PIvers:
Confirmed and fixed — added the no-op early-return in `adjustTime` right after the clamp math (`if (next === config.elapsedTime && nextDuration === config.duration) return;`), matching your suggested guard exactly. Also fixed the stale comment on line 104 that described the old (unclamped) behavior.

Added a regression test using `renderHook` to call `adjustTime` directly while running at the ceiling — this bypasses the UI's `disabled` gate (which, per the AdjustButton's own internal check, would otherwise mask this exact scenario from a component-level test) and confirms `updateWidget` is not called. Verified it fails without the guard (1 write) and passes with it (0 writes).

`pnpm run validate` and `pnpm run build` both green.

---

OPS-PIvers:
Both confirmed and fixed — the first one especially is a real correctness regression, thank you for catching it.

1. **No-op guard baseline**: changed `next === config.elapsedTime` to `next === current` exactly as suggested. Verified the false-positive by writing a test that advances fake timers 60s on a running 300s timer (RAF genuinely ticks under vitest's fake timers, confirmed empirically), then calls `adjustTime(60)` — with the old comparison this silently dropped the write (`next(300) === config.elapsedTime(300)` even though `current` had decayed to 240); with the fix it correctly writes. Also re-verified the original ceiling no-op test still passes with the corrected formula (in that synchronous, no-time-elapsed scenario `current === config.elapsedTime`, so both formulas agree there).

2. **Disabled prop while running**: added `|| config.elapsedTime >= TIME_TOOL_MAX_DURATION_SECONDS` to the "+" button's disabled expression. Added a test that renders running-at-ceiling, advances fake timers 1s (so `displayTime` has genuinely decayed below the cap), and confirms the button is still disabled.

`pnpm run validate` green (44/44 tests in the file).

---

OPS-PIvers:
Good catch, and I agree with your own analysis of the trade-off — the `||`→`&&` swap reintroduces the write storm, and widening the no-op guard to also match `config.elapsedTime` reintroduces the false-positive-drop bug this same PR just fixed. A timer *started* at 999m59s (rather than reaching it via +/- adjustment) is the only path that hits this, and it's a real but narrow limitation (`+` stays disabled for that run's duration even as the display decays well below the cap).

Given the two more consequential bugs already fixed in this PR came from chasing exactly this kind of edge case, I'd rather not risk a fourth attempt without a more deliberate design (e.g. tracking the run's own decaying baseline separately, rather than overloading `config.elapsedTime` for both "persisted target" and "ceiling-reached" purposes). Logging it as a backlog item in `docs/routines/debugger.md` for a future pass rather than patching now under review pressure. Not blocking — accepting the trade-off as-is.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-06

### Summary
Clamps the TimeTool hold-to-ramp path (`adjustTime` in `useTimeTool.ts`) to `TIME_TOOL_MAX_DURATION_SECONDS`, the same ceiling the Keypad and admin validator already enforce, and disables the "+" button at the ceiling. Two follow-up fixes from review are folded in: a no-op early-return to stop a write-storm/`startTime`-reset loop while running at the ceiling, and gating the "+" disable on `config.elapsedTime` (not just the decaying `displayTime`).

### Code Quality
✓ Clean. `next`/`nextDuration` both clamped via `Math.min(MAX, Math.max(0, …))`; the no-op guard correctly compares against `current` (what `next` was derived from), avoiding the false-positive that would silently drop a legitimate press when the running display coincidentally equals the persisted baseline.

### Widget Standards
✓ Fix lives in the hook/config layer, not front-face content scaling. Anchors to the shared constant rather than a magic number.

### Firestore & Security
✓ No issues. The clamp also prevents an out-of-range `duration`/`elapsedTime` from being persisted, which was the point of fixing it at the source rather than at display time.

### Regression Risk
✓ Low risk. Non-ceiling behavior is unchanged (guard is a no-op there). Keyboard Enter/Space path also routes through `adjustTime`, so it's covered.

### Missing Tests
✓ Strong coverage — five targeted tests including the running-at-ceiling no-op (via `renderHook` to bypass the UI `disabled` gate) and the "don't drop a real press after decay" regression.

### Overall
**Merge readiness:** Ready
Root-cause fix with the display-layer band-aid explicitly rejected; the review follow-ups closed the two subtle running-state paths.

---

---

PR #2160, opened 2026-07-07
fix(import-wizard): guard against stale in-flight promises after close/reopen

OPS-PIvers:
## Root cause

`components/common/library/importer/ImportWizard.tsx` is always mounted by every caller (`QuizWidget/Widget.tsx`, `MiniApp/Widget.tsx`, `VideoActivityWidget/components/Creator.tsx`) — callers only toggle the `isOpen` prop, never unmount the component. The wizard resets its ephemeral state on any `isOpen` transition, but its five async handlers (`runParse`, `handlePickSheet`, `handleFilePicked`, `handleAiGenerate`, `handleConfirmSave`) had no cancellation guard. Closing the wizard mid-operation and reopening it before the promise settled let the stale `.then` continuation run after the reset and overwrite the freshly-reset state with the cancelled operation's result — jumping straight to Preview with old data, or in the save case, firing `onSaved`/closing the freshly-reopened wizard.

This was a previously-flagged, unverified backlog lead from a prior nightly run; this run confirmed and fixed it.

## Fix

Added a `sessionRef` counter bumped on every `isOpen` transition. All five async handlers capture the session at entry and check it after each `await` before applying any `setState` — the general, root-cause fix, covering all five handlers that shared the identical race.

## Rejected band-aid

Unmounting/remounting the wizard via a `key` prop tied to `isOpen` at each of the three call sites was considered and rejected: it would touch three separate widget files instead of the shared primitive, wouldn't prevent the race for any future caller of `ImportWizard`, and would lose in-progress UI state (e.g. an AI-assist prompt draft) unnecessarily on every close.

## Regression test

`tests/components/ImportWizard.test.tsx` — new test using a deferred promise to simulate a stale in-flight parse surviving a close/reopen cycle.

**Fail-before / pass-after** (independently re-verified by the orchestrator): swapped in the pre-fix component — the new test fails (`expected document not to contain element, found <div data-testid="preview">`); restored, passes.

## Verification

- `pnpm run validate` — clean (orchestrator-independent run: type-check, lint, format, 558 files/6727 tests root + 37 files/703 tests functions, test-count guard).
- `pnpm run build` — passes (orchestrator-independent run).

## Risk: contained

Isolated to one shared component's internal state-guard logic; behavior for the non-race happy path is unchanged.

## Not fixed (backlog)

`handleCreateTemplate`/`handleCopyTemplateUrl` share the same architectural gap but only touch lower-risk state (`creatingTemplate`/`parseError`) — left for a future sweep.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VEJTpeRWqRCPMhEfKbDNPZ)_

### DIFF
--- a/components/common/library/importer/ImportWizard.tsx
+++ b/components/common/library/importer/ImportWizard.tsx
@@ -111,11 +111,18 @@ export function ImportWizard<TData>({
 
   const fileInputRef = useRef<HTMLInputElement>(null);
 
+  // Bumped on every open/close transition so in-flight async handlers (parse,
+  // AI-assist, picker) can detect a stale session and skip applying their
+  // result — the wizard stays mounted across close/reopen, so promises
+  // started before a close can still resolve after a reopen.
+  const sessionRef = useRef(0);
+
   // Reset ephemeral state whenever the wizard opens.
   // Using adjust-state-while-rendering so we don't need useEffect.
   const [prevOpen, setPrevOpen] = useState(isOpen);
   if (prevOpen !== isOpen) {
     setPrevOpen(isOpen);
+    sessionRef.current += 1;
     if (isOpen) {
       setStep('source');
       setSheetUrl('');
@@ -145,10 +152,14 @@ export function ImportWizard<TData>({
     supportsCsv || supportsJson || supportsHtml || supportsFile;
 
   const runParse = async (payload: ImportSourcePayload): Promise<void> => {
+    const session = sessionRef.current;
     setLoading(true);
     setParseError(null);
     try {
       const result = await adapter.parse(payload);
+      // The wizard was closed and/or reopened while this parse was in
+      // flight — its result belongs to a cancelled session, discard it.
+      if (session !== sessionRef.current) return;
       setParsed(result.data);
       setWarnings(result.warnings);
       // If the adapter can extract a title from the parsed data (e.g. an
@@ -163,11 +174,12 @@ export function ImportWizard<TData>({
       }
       setStep('preview');
     } catch (err) {
+      if (session !== sessionRef.current) return;
       setParseError(
         err instanceof Error ? err.message : 'Failed to parse import source.'
       );
     } finally {
-      setLoading(false);
+      if (session === sessionRef.current) setLoading(false);
     }
   };
 
@@ -181,23 +193,26 @@ export function ImportWizard<TData>({
 
   const handlePickSheet = async (): Promise<void> => {
     if (!adapter.pickSheet) return;
+    const session = sessionRef.current;
     setParseError(null);
     setPicking(true);
     try {
       const picked = await adapter.pickSheet();
+      if (session !== sessionRef.current) return;
       // Null = the user dismissed the Picker — stay on the source step with no
       // error. Otherwise parse the picked sheet (it's now drive.file-readable).
       if (picked) {
         await runParse({ kind: 'sheet', url: picked.url });
       }
     } catch (err) {
+      if (session !== sessionRef.current) return;
       setParseError(
         err instanceof Error
           ? err.message
           : 'Failed to open the Google Drive picker.'
       );
     } finally {
-      setPicking(false);
+      if (session === sessionRef.current) setPicking(false);
     }
   };
 
@@ -208,6 +223,7 @@ export function ImportWizard<TData>({
     if (fileInputRef.current) fileInputRef.current.value = '';
     if (!file) return;
 
+    const session = sessionRef.current;
     const kind = inferKindFromFileName(file.name, adapter.supportedSources);
     if (kind === 'file') {
       await runParse({ kind: 'file', file });
@@ -215,8 +231,10 @@ export function ImportWizard<TData>({
     }
     try {
       const text = await file.text();
+      if (session !== sessionRef.current) return;
       await runParse({ kind, text, fileName: file.name });
     } catch (err) {
+      if (session !== sessionRef.current) return;
       setParseError(
         err instanceof Error ? err.message : 'Failed to read the selected file.'
       );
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-07

### Summary
Adds a `sessionRef` counter (bumped on every `isOpen` transition) and guards all five async handlers in the always-mounted `ImportWizard` so a promise started before a close/reopen can't clobber the freshly-reset state. Clean root-cause fix that generalizes across all callers instead of patching each call site.

### Code Quality
✓ No issues found. The `session` is captured at entry and re-checked after every `await`, including inside `finally` (`if (session === sessionRef.current) setLoading(false)`), which correctly avoids resetting the *new* session's flags. Comments are concise and explain the non-obvious "always-mounted" invariant.

### Widget Standards
✓ Not applicable (shared library primitive, not a widget).

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found — no persistence or auth surface touched.

### Regression Risk
✓ Low risk. Isolated to one component's internal state-guard logic; happy-path behavior is unchanged (the guard only short-circuits stale continuations).

### Missing Tests
✓ Adequate — the new deferred-promise test exercises the exact close/reopen-mid-parse race. As the PR notes, `handleCreateTemplate`/`handleCopyTemplateUrl` share the pattern but touch only low-risk state; reasonable to leave for a follow-up.

### Overall
**Merge readiness:** Ready
Focused, well-tested fix for a real state-leak race; no concerns.

---

---

components/common/library/importer/ImportWizard.tsx
```diff
@@ -111,11 +111,18 @@ export function ImportWizard<TData>({
 
   const fileInputRef = useRef<HTMLInputElement>(null);
 
+  // Bumped on every open/close transition so in-flight async handlers (parse,
+  // AI-assist, picker) can detect a stale session and skip applying their
+  // result — the wizard stays mounted across close/reopen, so promises
+  // started before a close can still resolve after a reopen.
```
OPS-PIvers: CLAUDE.md: *"one short line max"* — 4-line block. The PR description already explains the mounted-wizard / stale-promise rationale; the comment just needs the non-obvious invariant:

```suggestion
  // Bumped on every open/close transition; in-flight handlers check it to drop results from a cancelled session.
  const sessionRef = useRef(0);
```

---

components/common/library/importer/ImportWizard.tsx
```diff
@@ -111,11 +111,18 @@ export function ImportWizard<TData>({
...
+  const sessionRef = useRef(0);
+
   // Reset ephemeral state whenever the wizard opens.
   // Using adjust-state-while-rendering so we don't need useEffect.
```
OPS-PIvers: CLAUDE.md: *"one short line max"* — 2-line block. Both lines state what the code does, not why. Remove entirely (the `sessionRef.current += 1` and the `if (isOpen !== prevIsOpen)` guard are self-explanatory) or collapse to:

```suggestion
  // Adjust-state-while-rendering: reset on open without useEffect.
```

---

components/common/library/importer/ImportWizard.tsx
```diff
@@ -145,10 +152,14 @@ export function ImportWizard<TData>({
...
 
   const runParse = async (payload: ImportSourcePayload): Promise<void> => {
+    const session = sessionRef.current;
     setLoading(true);
     setParseError(null);
     try {
       const result = await adapter.parse(payload);
+      // The wizard was closed and/or reopened while this parse was in
+      // flight — its result belongs to a cancelled session, discard it.
+      if (session !== sessionRef.current) return;
```
OPS-PIvers: CLAUDE.md: *"one short line max"* — same pattern repeats at lines ~202, ~237, ~289, ~296, ~327. Each 2-line stale-guard block should be a single line:

```suggestion
      // Stale session — wizard closed/reopened while this parse was in flight.
      if (session !== sessionRef.current) return;
```

The other repeated instances (at each `await` boundary in `handlePickSheet`, `handleFilePicked`, `handleAiGenerate`, `handleConfirmSave`) need the same treatment.

---

OPS-PIvers (commented):
CLAUDE.md comment-length sweep — the `sessionRef` comment, the 2-line reset block above it, and the repeating stale-guard 2-liners throughout all five async handlers.

---

components/common/library/importer/ImportWizard.tsx
OPS-PIvers: Collapsed this one and the sibling in `handleConfirmSave` (line ~325-326) — those two were the only sites that actually had a multi-line comment to collapse.

Checked the other five `if (session !== sessionRef.current) return;` sites you listed (~201, ~208, ~234, ~289, ~296, ~331) — none currently have a comment at all, so there's nothing to collapse there. I didn't add new one-line comments to them: the invariant is already documented once at the `sessionRef` declaration, and repeating an identical comment at every call site of an established pattern is the kind of thing CLAUDE.md's "default to no comments" guidance argues against, not a violation of "one line max." Happy to revisit if you'd rather have per-site markers for readability.

---

tests/components/ImportWizard.test.tsx
```diff
@@ -484,6 +484,63 @@ describe('ImportWizard', () => {
...
+    await act(async () => {
+      resolveParse?.({ data: { rows: ['stale-row'] }, warnings: [] });
+      await Promise.resolve();
+      await Promise.resolve();
+    });
+
+    // The stale result must not push the freshly-reopened wizard to Preview.
+    expect(screen.queryByTestId('preview')).not.toBeInTheDocument();
+    expect(screen.getByLabelText('Google Sheet URL')).toBeInTheDocument();
+  });
```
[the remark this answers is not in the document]
OPS-PIvers: Added, using the `renderWizard` helper (already wires `onClose`/`onSaved`) instead of raw `render` for consistency with the rest of the file. Verified fail-before/pass-after by swapping in the pre-fix component: fails with `onSaved` called once (`['My Quiz']`), passes with the shipped `sessionRef` guard. Full `pnpm run validate` green.

---

components/common/library/importer/ImportWizard.tsx
OPS-PIvers: Fixed — collapsed to the suggested single line: `// Bumped on every open/close transition; in-flight handlers check it to drop results from a cancelled session.` Thread is outdated because this landed in a follow-up commit.

---

components/common/library/importer/ImportWizard.tsx
OPS-PIvers: Fixed — collapsed to the suggested single line: `// Adjust-state-while-rendering: reset on open without useEffect.` Thread is outdated because this landed in the same follow-up commit.

---

PR #2178, opened 2026-07-11
fix(ImportWizard): guard 2 remaining async handlers against stale close/reopen results

OPS-PIvers:
## Root cause
`ImportWizard.tsx` is always mounted by all 3 callers — only `isOpen` toggles. A prior fix (#2160) added a `sessionRef` cancellation-guard counter to 5 of its async handlers, closing a race where a stale promise resolving after a close→reopen cycle could overwrite freshly-reset state. Two handlers, `handleCreateTemplate` and `handleCopyTemplateUrl`, were deliberately left out of that PR as "lower-risk state."

This run confirmed both are genuinely reachable: they're wired to real UI buttons (Create Template / Copy Template URL) rendered whenever `adapter.templateHelper` is set. Without the guard, closing the wizard mid-request and reopening it lets a stale rejection from the *old* session call `setParseError(...)` on the *freshly reset* session — surfacing an unrelated "Failed to create template" / "Failed to copy template URL" banner the user never triggered on the new screen.

## Fix
Mirrored the exact pattern already established for the other 5 handlers in this file: capture `const session = sessionRef.current` at entry, and check `if (session !== sessionRef.current) return;` before any state-setting call after an `await`, on both the success and error paths, and in `finally`.

## Rejected band-aid
Did not just suppress the banner or unconditionally clear `parseError` on reopen (it's already cleared on reopen today) — the actual race is that the stale write can land *after* the reset has already run, so the fix has to prevent the stale write itself via the session-token check, not paper over its symptom.

## Regression test
`tests/components/ImportWizard.test.tsx`: two new tests using the same deferred-promise + close/reopen pattern as the file's existing `sessionRef` tests.
- **Fail before**: both new tests failed — stale "Template service unavailable" / "Clipboard denied" banners found in the DOM after reopen.
- **Pass after**: 19/19 tests pass.

## Evidence
Independently reproduced by the orchestrator: checked out `dev-paul`'s `ImportWizard.tsx` with the new tests — 2 failed / 17 passed. Restored the fix — 19/19 passed. `pnpm run validate` and `pnpm run build` both pass on this branch (one `tests/hooks/useRosters.test.ts` failure seen during a contended validate run was confirmed to be the known pre-existing flaky-test race — fixed independently in the companion Build & Tooling PR from tonight's run — and unrelated to this diff; it passes 35/35 in isolation).

**Risk: contained**

---
_Generated by [Claude Code](https://claude.ai/code/session_01QSdrNnarDUKYMw6dFAhe4t)_

### DIFF
--- a/tests/components/ImportWizard.test.tsx
+++ b/tests/components/ImportWizard.test.tsx
@@ -598,6 +598,99 @@ describe('ImportWizard', () => {
     expect(screen.getByLabelText('Google Sheet URL')).toBeInTheDocument();
   });
 
+  it('does not surface a stale handleCreateTemplate error after close/reopen', async () => {
+    let rejectCreate: ((err: Error) => void) | null = null;
+    const createPromise = new Promise<{ url: string }>((_resolve, reject) => {
+      rejectCreate = reject;
+    });
+    const { adapter } = makeAdapter({
+      templateHelper: {
+        createTemplate: () => createPromise,
+        instructions: <p>Template instructions here.</p>,
+      },
+    });
+    const { rerender } = renderWizard(adapter);
+
+    fireEvent.click(
+      screen.getByRole('button', { name: /template .* format help/i })
+    );
+    fireEvent.click(screen.getByRole('button', { name: /create template/i }));
+
+    // Close the wizard, then reopen it before the stale create-template
+    // request settles — ImportWizard is always mounted; only isOpen toggles.
+    rerender(
+      <ImportWizard<FakeData>
+        isOpen={false}
+        onClose={vi.fn()}
+        adapter={adapter}
+      />
+    );
+    rerender(
+      <ImportWizard<FakeData>
+        isOpen={true}
+        onClose={vi.fn()}
+        adapter={adapter}
+      />
+    );
+
+    await act(async () => {
+      rejectCreate?.(new Error('Template service unavailable'));
+      await Promise.resolve();
+      await Promise.resolve();
+    });
+
+    // The freshly-reopened wizard must not show an error from the cancelled
+    // pre-reopen request.
+    expect(
+      screen.queryByText('Template service unavailable')
+    ).not.toBeInTheDocument();
+  });
+
+  it('does not surface a stale handleCopyTemplateUrl error after close/reopen', async () => {
+    let rejectCopy: ((err: Error) => void) | null = null;
+    const createPromise = new Promise<{ url: string }>((_resolve, reject) => {
+      rejectCopy = reject;
+    });
+    Object.assign(navigator, {
+      clipboard: { writeText: vi.fn().mockResolvedValue(undefined) },
+    });
+    const { adapter } = makeAdapter({
+      templateHelper: {
+        createTemplate: () => createPromise,
+        instructions: <p>Template instructions here.</p>,
+      },
+    });
+    const { rerender } = renderWizard(adapter);
+
+    fireEvent.click(
+      screen.getByRole('button', { name: /template .* format help/i })
+    );
+    fireEvent.click(screen.getByRole('button', { name: /copy template url/i }));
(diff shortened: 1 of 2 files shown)


tests/components/ImportWizard.test.tsx
OPS-PIvers, line 1: The two new tests cover the stale-**error** paths (a `parseError` from a cancelled session doesn't surface after reopen). That's the right thing to lock in given the PR description.

However, the stale-**success** path for `handleCreateTemplate` — specifically, that the blank tab opened synchronously by `window.open('about:blank', ...)` is closed when the session changes before `createTemplate` resolves — isn't tested. This is exactly the behaviour the (now-resolved, outdated) Gemini reviewer comment called out: a stale blank tab left permanently open in the user's browser. The code correctly handles it (`if (newTab && !newTab.closed) newTab.close(); return;`), but there's no regression test pinning that behaviour.

A test would look roughly like:

```tsx
it('closes the stale blank tab if the session changes before createTemplate resolves', async () => {
  let resolveCreate: ((v: { url: string }) => void) | null = null;
  const createPromise = new Promise<{ url: string }>((resolve) => {
    resolveCreate = resolve;
  });
  const mockClose = vi.fn();
  vi.spyOn(window, 'open').mockReturnValueOnce({ closed: false, close: mockClose, location: { href: '' } } as unknown as Window);

  const { adapter } = makeAdapter({
    templateHelper: {
      createTemplate: () => createPromise,
      instructions: <p>Instructions</p>,
    },
  });
  const { rerender } = renderWizard(adapter);

  // Open template panel and click Create Template
  fireEvent.click(screen.getByRole('button', { name: /template .* format help/i }));
  fireEvent.click(screen.getByRole('button', { name: /create template/i }));

  // Close + reopen before the promise settles
  rerender(<ImportWizard isOpen={false} onClose={vi.fn()} adapter={adapter} />);
  rerender(<ImportWizard isOpen={true}  onClose={vi.fn()} adapter={adapter} />);

  await act(async () => {
    resolveCreate?.({ url: 'https://example.com/tpl' });
    await Promise.resolve();
    await Promise.resolve();
  });

  // The stale blank tab must have been closed, not navigated
  expect(mockClose).toHaveBeenCalledTimes(1);
});
```

Without this, the tab-cleanup guard can be silently removed in the future with no failing test.

---

OPS-PIvers (commented):
The session guards on both handlers are correctly implemented and mirror the existing pattern from #2160. The stale-error tests are solid. Noting a gap in the stale-success test coverage above.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-12

### Summary
Extends the `sessionRef` cancellation-guard pattern (from #2160) to the two remaining `ImportWizard` async handlers — `handleCreateTemplate` and `handleCopyTemplateUrl` — closing the stale-promise-after-close/reopen race for the Create Template / Copy Template URL buttons. Also closes the synchronously-opened `about:blank` tab on the stale-session early return.

### Code Quality
✓ No issues. Captures `const session = sessionRef.current` at entry and checks `session !== sessionRef.current` before every post-`await` state write, on success, error, and `finally` paths — an exact mirror of the established 5-handler pattern in the same file. The `newTab.close()` on the stale early return prevents a permanently-orphaned blank tab.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues.

### Regression Risk
✓ Low — additive guards on two handlers; no change to the happy path.

### Missing Tests
✓ Adequate. Three new tests cover both stale-error paths and the stale-blank-tab cleanup (the latter added in response to the review thread). Fail-before 2/19 → pass-after 19/19.

### Overall
**Merge readiness:** Ready
Completes the cancellation-guard coverage for this component.

---

---

PR #2211, opened 2026-07-15
fix(GuidedLearning): count building sets in "All items" sidebar badge

OPS-PIvers:
## Root cause

`GuidedLearningManager.tsx`'s `folderItemCounts` was computed via `countItemsByFolder(sets)` — personal sets only. But the Library tab's "All items" view (`filterSourcedEntriesByFolder` with `selectedFolderId === null`) always shows personal + building sets together, and `FolderSidebar`'s "All items" badge sums every bucket in `itemCounts`. Since building sets never contributed to any bucket, the badge undercounted by `buildingSets.length` whenever any building sets existed — even though "All items" is exactly the view that shows them.

## Fix

Fold `buildingSets.length` into the root bucket (`ROOT_FOLDER_COUNT_KEY`), since building sets have no `folderId` and are never rendered as their own folder row — only summed into the "All items" total.

## Rejected band-aid

Considered recomputing the total directly inside `FolderSidebar` via a separately-passed "total override" prop. Rejected because `FolderSidebar`/`folderFilters.ts` is a shared primitive used by Quiz/VideoActivity/GuidedLearning/MiniApp, and none of the sibling widgets mix a "building set" concept into personal sets the way GuidedLearning does — special-casing the shared component for one caller would leak a GuidedLearning-specific concept into generic library infrastructure. Fixing it at the source keeps the shared primitives generic and honest.

## Regression test

`tests/components/widgets/GuidedLearningManager.folderCounts.test.tsx` — renders `GuidedLearningManager` with 1 personal set + 1 building set, asserts the "All items" badge equals 2.

**Fail before / pass after:** badge shows `1` pre-fix (fails — only counts the personal set), `2` post-fix (passes).

## Validation

- `pnpm run validate` — passing (703 functions tests, full root suite, test:counts guard)
- `pnpm run build` — passing

## Risk

**Contained** — single widget-local memo change, additive test coverage, no shared-component API changes.

---
🌙 Nightly code-health run — Widgets area

---
_Generated by [Claude Code](https://claude.ai/code/session_01SJSN9QBeEFg6nyXGBW6mGb)_

### DIFF
--- a/components/widgets/GuidedLearning/components/GuidedLearningManager.tsx
+++ b/components/widgets/GuidedLearning/components/GuidedLearningManager.tsx
@@ -65,6 +65,7 @@ import { LibraryPreviewPane } from '@/components/common/library/LibraryPreviewPa
 import {
   countItemsByFolder,
   filterSourcedEntriesByFolder,
+  ROOT_FOLDER_COUNT_KEY,
 } from '@/components/common/library/folderFilters';
 import { useFolders } from '@/hooks/useFolders';
 import { ScaledEmptyState } from '@/components/common/ScaledEmptyState';
@@ -439,9 +440,15 @@ export const GuidedLearningManager: React.FC<GuidedLearningManagerProps> = ({
     setSelectedFolderId(null);
   }
 
-  // Only personal sets participate in folders. Building sets are always at
-  // root (they're shared at the building level and have no folderId field).
-  const folderItemCounts = useMemo(() => countItemsByFolder(sets), [sets]);
+  // Building sets have no folderId; fold them into the root bucket so "All items" counts them.
+  const folderItemCounts = useMemo(() => {
+    const counts = countItemsByFolder(sets);
+    if (buildingSets.length > 0) {
+      counts[ROOT_FOLDER_COUNT_KEY] =
+        (counts[ROOT_FOLDER_COUNT_KEY] ?? 0) + buildingSets.length;
+    }
+    return counts;
+  }, [sets, buildingSets]);
 
   const allEntries = useMemo(
     () =>
--- a/tests/components/widgets/GuidedLearningManager.folderCounts.test.tsx
+++ b/tests/components/widgets/GuidedLearningManager.folderCounts.test.tsx
@@ -1,0 +1,92 @@
+// Regression test: "All items" badge must count building sets, not just personal sets.
+
+import React from 'react';
+import { describe, it, expect, vi } from 'vitest';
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-15

### Summary
This PR fixes an undercount in `GuidedLearningManager`'s "All items" sidebar badge by folding `buildingSets.length` into the root count bucket (`ROOT_FOLDER_COUNT_KEY`) after `countItemsByFolder(sets)`. I verified the fix targets the correct bucket and does not double-count: `sets` passed to `countItemsByFolder` is personal-only (building sets are mapped separately with `source:'building'` in `buildLibraryEntries`), and `FolderSidebar.totalItemCount` sums `Object.values(itemCounts)`, so the "All items" total correctly rises by the building count while no real folder row (keyed by `folder.id`) is affected.

### Code Quality
✓ No issues found. The memo now returns the augmented `counts` object and its dependency array correctly adds `buildingSets` alongside `sets`. Uses the exported `ROOT_FOLDER_COUNT_KEY` constant rather than the literal `'root'` string, keeping it in sync with the shared primitive. No console/promise/lint concerns.

### Widget Standards
✓ No issues found. Change is confined to a data-computation `useMemo` in the manager's library sidebar (not front-face scaled content), so container-query/cqmin rules don't apply. No `WidgetRegistry`, `lazyNamed`, or `skipScaling` surface touched.

### Admin Config Standards
✓ Not applicable — no `components/admin/` changes.

### Firestore & Security
✓ No issues found. Purely a client-side derived-count calculation; no reads, writes, rules, or auth involved.

### Regression Risk
✓ Low risk. The fix is additive and guarded by `if (buildingSets.length > 0)`, so the personal-only case is unchanged. It deliberately fixes at the source rather than special-casing the shared `FolderSidebar`/`folderFilters` primitive — the right call, since sibling widgets (Quiz/VA/MiniApp) don't mix a "building set" concept. No shared API, types union, or three-map consistency concerns.

### Missing Tests
✓ Adequate coverage. `tests/components/widgets/GuidedLearningManager.folderCounts.test.tsx` renders the manager with 1 personal + 1 building set and asserts the "All items" badge reads `2` (would be `1` pre-fix). Non-blocking suggestion: a case with a foldered personal set plus a building set would additionally lock in that the building count lands in root and not a named folder bucket, but the current test adequately covers the reported regression.

### Overall
Correct, minimal, well-reasoned fix at the proper layer with a real fail-before/pass-after regression test; only an optional extra test case is suggested.

---

---

PR #2215, opened 2026-07-15
fix(rules): close moderation-bypass on activity_wall_sessions submissions

OPS-PIvers:
## Root cause

**Security.** The `activity_wall_sessions/{session}/submissions` create rule only checked `request.resource.data.status in ['approved', 'pending']` — accepting either value regardless of the session's `moderationEnabled` flag. The client always writes `status: moderationEnabled ? 'pending' : 'approved'` (`ActivityWallStudentApp.tsx:365`), but nothing server-side enforced that contract. Any submitter (including anonymous/PIN-based students) could set `status: 'approved'` directly on a moderated session and skip the teacher's review queue entirely — a moderation bypass.

## Fix

Added `awSessionModerationEnabled()` helper (mirrors the existing `awSessionClassIds()` pattern in the same rule block) and pinned `create`'s `status` field to the value the moderation flag implies:
```
request.resource.data.status == (awSessionModerationEnabled() ? 'pending' : 'approved')
```

## Rejected band-aid

Considered leaving the rule as an `in [...]` check and instead relying on the client to "always do the right thing." Rejected — the whole point of Firestore security rules is to not trust the client for anything security-relevant; the client-side ternary is UX, not enforcement.

## Regression test

`tests/rules/studentRoleClassGate.test.ts` — new `activity_wall_sessions/submissions — moderation status gate` suite (4 tests): cannot self-approve when moderation is enabled, can submit pending when moderation is enabled, auto-approves when moderation is disabled, cannot submit pending when moderation is disabled. Also updated 2 pre-existing describe blocks whose seeded sessions have no `moderationEnabled` field (rules default `false`) to expect `'approved'` instead of `'pending'`, since their old expectation was only ever passing because the old rule accepted both values unconditionally.

**Fail before / pass after:** ran the new suite in isolation against unfixed `firestore.rules` — 2/4 new tests fail (the two bypass-prevention assertions: self-approve-when-moderated succeeds when it should fail, submit-pending-when-open succeeds when it should fail). Full `pnpm run test:rules` suite: 1116/1116 passing post-fix (1112 baseline + 4 new).

## Validation

- `pnpm run validate` — passing (703 functions tests, full root suite, test:counts guard)
- `pnpm run test:rules` — passing, 47 files / 1116 tests (Firestore emulator)
- `pnpm run build:all` — passing

## Risk

**Contained** — single rule-file change scoped to one collection's create path, verified against the real Firestore emulator with full before/after evidence, no client code changes required (the client was already writing the correct value).

---
🌙 Nightly code-health run — Build & Tooling area

---
_Generated by [Claude Code](https://claude.ai/code/session_01SJSN9QBeEFg6nyXGBW6mGb)_

### DIFF
--- a/tests/rules/studentRoleClassGate.test.ts
+++ b/tests/rules/studentRoleClassGate.test.ts
@@ -967,7 +967,7 @@ describe('activity_wall_sessions/submissions — moderation status gate', () => 
   const col = 'activity_wall_sessions';
   const MODERATED_SESSION = 'session-moderated';
   const OPEN_SESSION = 'session-open';
-  const subFor = (session: string, status: 'approved' | 'pending') => ({
+  const subFor = (status: 'approved' | 'pending') => ({
     id: 'sub-1',
     content: 'Hello world',
     submittedAt: 1000,
@@ -997,7 +997,7 @@ describe('activity_wall_sessions/submissions — moderation status gate', () => 
     await assertFails(
       addDoc(
         collection(asStudentA(), `${col}/${MODERATED_SESSION}/submissions`),
-        subFor(MODERATED_SESSION, 'approved')
+        subFor('approved')
       )
     );
   });
@@ -1006,7 +1006,7 @@ describe('activity_wall_sessions/submissions — moderation status gate', () => 
     await assertSucceeds(
       addDoc(
         collection(asStudentA(), `${col}/${MODERATED_SESSION}/submissions`),
-        subFor(MODERATED_SESSION, 'pending')
+        subFor('pending')
       )
     );
   });
@@ -1015,7 +1015,7 @@ describe('activity_wall_sessions/submissions — moderation status gate', () => 
     await assertSucceeds(
       addDoc(
         collection(asStudentA(), `${col}/${OPEN_SESSION}/submissions`),
-        subFor(OPEN_SESSION, 'approved')
+        subFor('approved')
       )
     );
   });
@@ -1024,7 +1024,7 @@ describe('activity_wall_sessions/submissions — moderation status gate', () => 
     await assertFails(
       addDoc(
         collection(asStudentA(), `${col}/${OPEN_SESSION}/submissions`),
-        subFor(OPEN_SESSION, 'pending')
+        subFor('pending')
       )
     );
   });
--- a/tests/rules/studentRoleClassGate.test.ts
+++ b/tests/rules/studentRoleClassGate.test.ts
@@ -823,11 +823,12 @@ describe('activity_wall_sessions/submissions — student-role gate', () => {
   const col = 'activity_wall_sessions';
   const subCol = (session: string) =>
     collection(asStudentA(), `${col}/${session}/submissions`);
+  // seedSessions() has no moderationEnabled field, so rules default to false → 'approved'.
   const validSub = () => ({
     id: 'sub-1',
     content: 'Hello world',
     submittedAt: 1000,
-    status: 'pending' as const,
+    status: 'approved' as const,
   });
 
   beforeEach(async () => {
@@ -890,11 +891,12 @@ describe('activity_wall_sessions/submissions — student-role gate', () => {
 // above, which seeds `classId` and still passes unchanged).
 describe('activity_wall_sessions/submissions — Phase-5A compat (classIds list)', () => {
   const col = 'activity_wall_sessions';
+  // No moderationEnabled field seeded → rules default false → 'approved'.
   const validSub = () => ({
     id: 'sub-1',
     content: 'Hello world',
     submittedAt: 1000,
-    status: 'pending' as const,
+    status: 'approved' as const,
   });
 
   beforeEach(async () => {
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-15

### Summary
This PR closes a real moderation-bypass in the `activity_wall_sessions/{session}/submissions` create rule. Previously `status` was validated only with `in ['approved', 'pending']`, so a submitter (including anonymous/PIN students) could write `status: 'approved'` on a moderated session and skip the teacher's review queue. The fix adds an `awSessionModerationEnabled()` helper and pins `status` to the value the flag implies. The logic is sound and the change is well-scoped and well-evidenced.

### Code Quality
✓ No issues found. The new `awSessionModerationEnabled()` helper mirrors the existing `awSessionClassIds()` pattern in the same block and uses `.data.get('moderationEnabled', false)`, so the default (absent flag → `false` → `'approved'`) matches the documented client contract (`ActivityWallStudentApp.tsx` writing `moderationEnabled ? 'pending' : 'approved'`). Both `get()` calls hit the same session document, and Firestore rules caches document reads within a single evaluation, so no additional billed read is introduced.

### Widget Standards
✓ Not applicable — no `components/widgets` files touched.

### Admin Config Standards
✓ Not applicable — no `components/admin` files touched.

### Firestore & Security
Core fix is correct: `status == (awSessionModerationEnabled() ? 'pending' : 'approved')` exactly enforces the intended contract and removes client trust for a security-relevant field. Two items worth verifying (both outside this diff, non-blocking):
- **Update path:** this diff only hardens the `create` rule. If the submissions `update`/`allow write` rule permits mutating `status`, a submitter could still flip `pending → approved` after creation. Confirm the update rule for this subcollection already forbids student-side `status` changes; if not, the bypass is only half-closed.
- **Sibling collections:** other session-submission collections with a moderation concept (e.g. `mini_app_sessions/submissions`, immediately below in the same file) should be checked for the same `in [...]` pattern so this class of bug isn't left open elsewhere. Only flag/fix if they actually carry a `moderationEnabled`-style flag.

### Regression Risk
✓ Low risk. Single rule-file change to one collection's create path. The two pre-existing describe blocks were correctly updated to expect `'approved'` (their seeds have no `moderationEnabled` field → default `false`); their prior `'pending'` expectation only passed because the old rule accepted both values unconditionally, so this is a correctness improvement, not a masking change. Client already writes the correct value, so no client change is required.

### Missing Tests
✓ Adequate coverage for the create path — the new 4-test suite covers both moderation states × both status values (self-approve-when-moderated fails, pending-when-moderated succeeds, approved-when-open succeeds, pending-when-open fails), with documented fail-before/pass-after evidence. If the update path is found to allow `status` mutation, add a companion test there.

### Overall
Security fix is correct, minimal, and well-tested; before merge, just confirm the submissions `update` rule can't be used to re-approve a `pending` submission and that no sibling session collection needs the same pinning.

---

---

OPS-PIvers:
Checked both open items from the review:

**Update path**: the `submissions` `allow update` rule (same match block, just below `create`) is already gated to `isAdmin() || sessionId.matches(request.auth.uid + '_.*')` — since `activity_wall_sessions` doc IDs are `{teacherUid}_...`, a student submitter can never match that pattern, so they cannot reach `update` at all, only the owning teacher or an admin. `status in ['approved', 'pending']` there is intentional: it lets the teacher flip a submission during review. No bypass on this path.

**Sibling collections**: swept the whole rules file for `moderat` and `status in [`. `mini_app_sessions/submissions` (the collection called out) has no `status`/moderation field at all. Every other `status in [...]` occurrence gates an unrelated session-lifecycle state machine (`active`/`paused`/`inactive`, `in-progress`/`completed`, `accepted`/`declined`, `draft`/`submitted`, `active`/`ended`) — none of them combine a moderation-style boolean flag with a client-settable review-state field the way `activity_wall_sessions` does. `moderationEnabled` only exists on this one collection.

Both confirmed clean — no follow-up needed.

---

tests/rules/studentRoleClassGate.test.ts
```diff
@@ -953,6 +955,81 @@ describe('activity_wall_sessions/submissions — Phase-5A compat (classIds list)
...
+// ---------------------------------------------------------------------------
+// activity_wall_sessions/submissions — moderation status is server-enforced
+// ---------------------------------------------------------------------------
+// The client always writes status: moderationEnabled ? 'pending' : 'approved'
+// (see ActivityWallStudentApp.tsx). Prior to this gate the create rule only
+// checked `status in ['approved', 'pending']`, so a submitter could set
+// status: 'approved' directly on a moderated session and skip the teacher's
+// review queue entirely.
+describe('activity_wall_sessions/submissions — moderation status gate', () => {
+  const col = 'activity_wall_sessions';
```
OPS-PIvers, line 967: `session` is accepted but never referenced inside the returned object — all four callers pass the session-specific constant but it has no effect. Drop the first parameter and update the callsites:

```suggestion
  const subFor = (status: 'approved' | 'pending') => ({
```

Then update the four call sites, e.g. `subFor(MODERATED_SESSION, 'approved')` → `subFor('approved')`.

---

OPS-PIvers: Valid — already addressed in 97975ee ("test: drop unused session param from subFor helper"). `subFor` now takes only `(status)` and the four call sites pass `subFor('approved')` / `subFor('pending')`. Resolving.

---

PR #2228, opened 2026-07-18
fix(FolderTree): dismiss the folder overflow menu on outside click/Escape

OPS-PIvers:
## Root cause

`FolderRow`'s kebab overflow menu (Rename / New subfolder / Move to root / Delete) in `components/common/library/FolderTree.tsx` only closed via its own kebab toggle or one of its own menu items — there was no outside-click or Escape handler, unlike every other kebab/overflow menu in this codebase (e.g. `SidebarPlcs`'s `PlcRow`, which uses `useClickOutside` plus an Escape `keydown` listener).

`FolderSidebar`/`FolderTree` is wired into `QuizManager`, `VideoActivityManager`, `GuidedLearningManager`, and `MiniAppManager` — all live, everyday teacher surfaces — so opening a folder's actions menu and then clicking anywhere else in the widget (another folder, the item grid, etc.) left the menu floating open indefinitely.

## Fix

Added `useClickOutside` (ignoring the kebab button) and an Escape `keydown` listener to `FolderRow`, matching the established `PlcRow` pattern exactly. Both are no-ops while the menu is closed (`menuRef` is null).

## Band-aid rejected

None needed — this is the already-proven, established pattern used by every other overflow menu in the codebase; there was no alternative shortcut worth taking instead of applying it here too.

## Regression test

`tests/components/common/library/FolderTree.menuDismiss.test.tsx` — renders `FolderSidebar` (the real stateful owner), opens a folder's menu, and asserts it closes on (1) a click elsewhere in the sidebar and (2) Escape.
- **Before fix:** both assertions fail — the menu stays open in both cases.
- **After fix:** both pass; full pre-existing `FolderTree`/`FolderSidebar` suite (11/11) still passes, no regressions.
- Verified independently by the orchestrator: reverted only the source file to `dev-paul`, confirmed both new assertions fail; restored the fix, confirmed both pass.

## Note on scope

An earlier draft of this fix targeted `PeriodSelector.tsx`'s Escape handling (same missing-guard bug class), but that component currently has zero live consumers anywhere in the app (dead/staged code pending future wiring) — discarded in favor of this genuinely reachable bug and logged as a backlog item instead.

## Validation

- `pnpm run validate` — pass (type-check, lint, format, root + functions test suites, test-count guard)
- `pnpm run build` — pass

**Risk:** contained (single shared component, isolated dismissal-guard addition mirroring an existing pattern).

---
🤖 Nightly code-health run — orchestrated by Claude Code.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DYzHXdbjXHWv3AsJ1RjMpt)_

### DIFF
--- a/components/common/library/FolderTree.tsx
+++ b/components/common/library/FolderTree.tsx
@@ -22,6 +22,8 @@ import {
 } from 'lucide-react';
 import { useDroppable } from '@dnd-kit/core';
 import type { LibraryFolder } from '@/types';
+import { useClickOutside } from '@/hooks/useClickOutside';
+import { isEscapeFromWidgetInput } from '@/utils/domHelpers';
 import { folderDroppableId, type FolderDropData } from './folderDropTargets';
 import { useFolderPanelMode } from './LibraryFolderPanelContext';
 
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-18

### Summary
Adds outside-click and Escape dismissal to `FolderRow`'s kebab overflow menu in `FolderTree.tsx`, which previously only closed via its own toggle/items — leaving it floating open after a stray click. `FolderSidebar` is live in the Quiz/Video Activity/Guided Learning/Mini App managers, so this is a genuinely reachable UX papercut. The fix reuses the established `useClickOutside` + Escape pattern from `SidebarPlcs.PlcRow`.

### Code Quality
✓ No issues. Matches the existing codebase convention exactly, including the `isEscapeFromWidgetInput` guard so Escape from an input doesn't collide. The `useEffect` here is a correct use of the escape hatch — it subscribes to a `document` keydown listener (an external system) and cleans it up on unmount.

### Widget Standards
✓ Not applicable — shared library component, no widget registry/scaling/appearance surface.

### Firestore & Security
✓ No issues — no data-layer changes.

### Regression Risk
✓ Low. Both listeners are no-ops while the menu is closed (`menuRef` is null), and the kebab button is excluded from the outside-click ref so toggling still works.

### Missing Tests
✓ Adequate. `FolderTree.menuDismiss.test.tsx` renders the real `FolderSidebar` owner and asserts both dismissal paths; the PR notes the pre-existing 11/11 suite still passes.

### Overall
**Merge readiness:** Ready
Established-pattern fix to a live surface with a targeted regression test.

---

---

PR #2242, opened 2026-07-19
fix(rules): SECURITY — gate shared_activity_walls read on revoked/expiresAt

OPS-PIvers:
## Root cause

`firestore.rules`' `shared_activity_walls/{shareId}` READ rule was `allow read: if request.auth != null;` — unconditional for any authenticated caller. The doc schema carries `expiresAt` and `revoked` fields whose entire purpose (per `ActivityWallGalleryView.tsx`) is to gate access — the client shows "revoked"/"expired" states when these trip, but that enforcement was **UI-only**. A caller bypassing the app (direct Firestore SDK/REST call) could read a gallery share's `title`, `prompt`, `sessionId`, `identificationMode`, etc. forever, regardless of revoke or expiry.

Sibling collections `/shared_boards` and `/shared_collections` already enforce an `expiresAt` cutoff on their substitute-mode reads (see #2150) — `shared_activity_walls` was the asymmetric sibling that never got the same treatment, matching a recurring pattern in this routine's history.

Confirmed the affected client reads are all single-doc `getDoc()` calls (`ActivityWallGalleryView.tsx`, `ShareModal.tsx`) — no live `onSnapshot` list queries against this collection, so this fix does not carry the "list query silently breaks" risk documented for time-gated rule changes on other collections (#2150).

## Rejected band-aid

Widening the client-side expiry check would do nothing — the whole point is a non-client caller bypasses the client entirely. The only real fix is in the security rule itself.

## Fix

`allow read` now requires: original author (management/cleanup) OR admin OR (`revoked` is not `true` AND (`expiresAt` unset OR in the future)).

## Regression test

`tests/rules/sharedActivityWalls.test.ts` — 5 new tests: live share readable, revoked share denied, expired share denied, host can still read a revoked share, admin can still read an expired share.

**Fail-before** (orchestrator-verified against unmodified `dev-paul` rules with the new tests applied, real Firestore emulator):
```
FAIL ... > viewer cannot read a revoked share
Error: Expected request to fail, but it succeeded.
FAIL ... > viewer cannot read a share past its expiresAt
Error: Expected request to fail, but it succeeded.
Tests  2 failed | 24 passed (26)
```

**Pass-after**: 26/26 passed. Full `pnpm run test:rules` suite (orchestrator re-ran independently): **47 files / 1121 tests, all passing** — no regressions in sibling rule suites.

## Verification

- `pnpm run validate` — green (type-check, lint, format, tests, functions tests, test:counts)
- `pnpm run build:all` — green (Vite build + SSR prerender + functions `tsc` build)
- `pnpm run test:rules` — green, full suite, orchestrator-verified independently against the real Firestore emulator

## Follow-up flagged (not fixed here, out of this area's glob)

`activity_wall_sessions.publiclyShared` is set `true` on share but never flipped back to `false` on revoke/expire — the underlying submission photos/text (a separate collection from the share doc fixed here) stay world-readable indefinitely once shared, independent of this share doc's own revoke/expiry state. Needs both a rules change (or a scheduled sweep function, mirroring `expireSubShares.ts`) and client wiring — spans outside `firestore.rules`/`functions/**` alone; recommend a dedicated cross-area follow-up.

## Risk

**Architectural** (security-rule tightening on a live collection) — but verified narrowly contained: single-doc reads only, no live-listener breakage risk, full sibling rules suite green.

---
🌙 Nightly code-health run — run 30 (2026-07-19)

---
_Generated by [Claude Code](https://claude.ai/code/session_01RenmR9eXRsJqw7oUtbA6cX)_

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -1097,7 +1097,23 @@ service cloud.firestore {
     // doc for the sessionId + toggles (allowComments/allowCommentResponses/
     // allowLikes). Readable by any signed-in caller (incl. anonymous students).
     match /shared_activity_walls/{shareId} {
-      allow read: if request.auth != null;
+      // READ: any authed caller (incl. anonymous gallery viewers) may read a
+      // live share; host and admin can still read past revoke/expiry for
+      // management/cleanup. Mirrors the expiresAt cutoff already enforced on
+      // /shared_boards and /shared_collections — without this, the
+      // "Available until X" / revoke UX in ActivityWallGalleryView is
+      // cosmetic only, since a caller that bypasses the client UI can keep
+      // reading a gallery's title/prompt/sessionId/identificationMode
+      // forever after the teacher revokes it or its expiresAt passes.
+      allow read: if request.auth != null && (
+        resource.data.get('originalAuthor', null) == request.auth.uid
+        || isAdmin()
+        || (
+          resource.data.get('revoked', false) != true
+          && (resource.data.get('expiresAt', null) == null
+              || resource.data.get('expiresAt', 0) > request.time.toMillis())
+        )
+      );
 
       // Identity fields + the sessionId-ownership check force the doc to be
       // created by the teacher who owns the underlying session (can't publish
--- a/tests/rules/sharedActivityWalls.test.ts
+++ b/tests/rules/sharedActivityWalls.test.ts
@@ -235,6 +235,18 @@ describe('shared_activity_walls — read gating (expiresAt / revoked)', () => {
     );
   });
 
+  it('viewer can read a permanent share (expiresAt null / unset)', async () => {
+    await testEnv.withSecurityRulesDisabled(async (ctx) => {
+      await setDoc(
+        doc(ctx.firestore(), `shared_activity_walls/${SHARE_ID}`),
+        sharedDocPayload()
+      );
+    });
+    await assertSucceeds(
+      getDoc(doc(asViewer(), `shared_activity_walls/${SHARE_ID}`))
+    );
+  });
+
   it('viewer cannot read a revoked share', async () => {
     await testEnv.withSecurityRulesDisabled(async (ctx) => {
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-19

### Summary
Tightens the `shared_activity_walls/{shareId}` READ rule so revoke/expiry is enforced server-side rather than being UI-only, mirroring the `expiresAt` cutoff already on `/shared_boards` and `/shared_collections`. Well-scoped security fix with a dedicated regression suite.

### Code Quality
✓ No issues. Rule logic is correct: author OR admin OR (`revoked != true` AND (`expiresAt` unset OR in the future)). Author/admin escape hatch preserves management/cleanup access.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ Sound. One verification note (non-blocking): the rule compares `resource.data.get('expiresAt', 0) > request.time.toMillis()`, which assumes `expiresAt` is stored as epoch **milliseconds** (a number), not a Firestore `Timestamp`. The new tests write it as `Date.now() ± N` (a number), which matches — just confirm the client write path (`ShareModal`/gallery publish) also persists millis, consistent with the sibling collections it mirrors.

### Regression Risk
Low-to-moderate — a rule tightening on a live collection, but the PR correctly confirms all reads are single-doc `getDoc()` (no `onSnapshot` list queries that would silently break under a time-gated rule), and the full rules suite is green. The flagged `activity_wall_sessions.publiclyShared` follow-up is a legitimately separate concern, correctly left out of scope.

### Missing Tests
✓ Adequate — 5 new tests cover live/permanent/revoked/expired viewer reads plus author-revoked and admin-expired escape hatches.

### Overall
Correct, well-tested security fix; only confirm the `expiresAt`-as-millis assumption on the write side.

---

---

PR #2248, opened 2026-07-20
fix(DraggableWindow): cancel touch long-press timers on unmount

OPS-PIvers:
## Root cause

The 1-finger (screenshot) and 2-finger (annotation-toggle) touch long-press `setTimeout`s in `DraggableWindow.tsx` were only cleared in `onPointerUp`/`onPointerCancel`. If a widget is removed while a touch is still held down (Firestore-driven delete from another session, dashboard switch, admin force-remove), the host DOM node detaches, so `onPointerUp` never fires and nothing clears the timer. 600ms later the stale callback fires anyway, calling `takeScreenshot()`/`setIsAnnotating()` plus `handleCloseTools()` — the latter calls the dashboard-wide `setSelectedWidgetId(null)`, which can clear whatever widget the user has since selected. Confirmed reachable — wired via `onPointerDownCapture`/`onPointerUpCapture` on every widget, not dead code.

## Fix

A write-once-read-only `isUnmountedRef` boolean, set by a dedicated unmount effect and checked at the top of both timer callbacks — matches the file's existing defense-in-depth pattern for drag/resize gesture cleanup (`activeGestureCleanupRef`).

## Band-aid rejected

A naive `useEffect` cleanup reading `longPressTimer.current`/`twoFingerLongPressTimer.current` directly (`clearTimeout` in the cleanup function) trips `eslint-plugin-react-hooks` v7's `react-hooks/immutability` rule once those refs are read in an effect, and destabilizes 5 pre-existing, necessary `react-hooks/refs` disable-comments elsewhere in this file. The write-once-read-only ref sidesteps this without touching unrelated lint suppressions.

## Regression test

`cancels the pending long-press screenshot timer when the widget unmounts mid-press` — fires a touch `pointerdown`, unmounts before the 600ms timer fires, advances fake timers 700ms, asserts `takeScreenshot`/`setSelectedWidgetId` were never called. Orchestrator independently re-verified: reverted just the source fix, confirmed the test fails (`expected ... not to be called ... called 1 times`); restored the fix, confirmed all 71 tests in the file pass.

## Evidence

Subagent-run `pnpm run validate` + `pnpm run build` both green. Orchestrator re-ran the full test file (71/71 passing) after the fail-before/pass-after check.

## Risk

**Contained** — localized to two `setTimeout` callbacks in one widget-chrome component, matches an established in-file pattern, no API/type/behavior change for the non-buggy path.

## Backlog (investigated, not pursued)

- `components/layout/sidebar/SidebarWidgets.tsx` has no `canAccessWidget` permission gate on its `TOOLS` list (unlike `Dock.tsx`), but the component has zero live consumers (`Sidebar.tsx` never imports it) — dead code, not user-reachable today.

---
🤖 Nightly code-health run — automated bug hunt + orchestrator review.

---
_Generated by [Claude Code](https://claude.ai/code/session_01M33BZjYmSCA3LhUBqhaCX3)_

### DIFF
--- a/components/common/DraggableWindow.test.tsx
+++ b/components/common/DraggableWindow.test.tsx
@@ -51,6 +51,10 @@ interface GlassCardProps {
   children: React.ReactNode;
   className?: string;
   onPointerDown?: (e: React.PointerEvent) => void;
+  onPointerDownCapture?: (e: React.PointerEvent) => void;
+  onPointerMoveCapture?: (e: React.PointerEvent) => void;
+  onPointerUpCapture?: (e: React.PointerEvent) => void;
+  onPointerCancelCapture?: (e: React.PointerEvent) => void;
   onClick?: (e: React.MouseEvent) => void;
   onKeyDown?: (e: React.KeyboardEvent) => void;
   onTouchStart?: (e: React.TouchEvent) => void;
@@ -76,6 +80,10 @@ vi.mock('./GlassCard', () => {
         children,
         className,
         onPointerDown,
+        onPointerDownCapture,
+        onPointerMoveCapture,
+        onPointerUpCapture,
+        onPointerCancelCapture,
         onClick,
         onKeyDown,
         onTouchStart,
@@ -92,6 +100,10 @@ vi.mock('./GlassCard', () => {
         data-testid="draggable-window"
         className={className}
         onPointerDown={onPointerDown}
+        onPointerDownCapture={onPointerDownCapture}
+        onPointerMoveCapture={onPointerMoveCapture}
+        onPointerUpCapture={onPointerUpCapture}
+        onPointerCancelCapture={onPointerCancelCapture}
         onClick={onClick}
         onKeyDown={onKeyDown}
         onTouchStart={onTouchStart}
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-20

### Summary
Adds a write-once `isUnmountedRef` guard to the two touch long-press `setTimeout` callbacks in `DraggableWindow.tsx`, so a widget removed mid-press (Firestore-driven delete, dashboard switch, admin force-remove) can't have a stale 600ms callback fire `takeScreenshot()`/`setIsAnnotating()` and clobber `setSelectedWidgetId(null)` after unmount. Correct, minimal fix.

### Code Quality
✓ The ref is written only by the dedicated unmount effect and read only in the two timer callbacks — this is a legitimate external-sync use of `useEffect` (unmount lifecycle), not derived-state misuse. Matches the file's existing defense-in-depth gesture-cleanup convention, and the PR body documents why the simpler direct-`clearTimeout` cleanup was rejected (it destabilizes the pre-existing `react-hooks/refs` suppressions).

### Widget Standards
✓ Not applicable (widget chrome, not front-face content).

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No changes.

### Regression Risk
✓ Low. The guard is a no-op on the normal (mounted) path — both callbacks still null out their timer ref and run their existing logic. No API/type change.

### Missing Tests
✓ Adequate. New regression test fires a touch `pointerdown`, unmounts before the 600ms timer, advances fake timers, and asserts neither `takeScreenshot` nor `setSelectedWidgetId` fired.

### Overall
**Merge readiness:** Ready
Small, correct lifecycle-safety fix with a targeted regression test.

---

---

PR #2249, opened 2026-07-20
fix(rosters): assignPins no longer collides with manually-set PINs

OPS-PIvers:
## Root cause

`assignPins()` (`hooks/useRosters.ts`) backfilled a blank student PIN using pure array position — `String(i + 1).padStart(2, '0')` — with no check against PINs other students in the same roster already had. The roster editor's duplicate-PIN warning runs against pre-assignment (still-blank) state, so it can never catch a collision between a manually-typed PIN and one that will later be auto-derived from position. Example: student #5 left blank, student #7 manually set to `"05"` — the editor shows no warning, but `assignPins()` silently gives student #5 `"05"` too. Both feed `syncRosterPinIndex()`, which writes the sidecar the PIN-based ClassLink/SSO login flow uses to resolve student identity — a collision there is a real correctness/identity-resolution bug on login, not cosmetic.

## Fix

Extracted `assignPins` to `utils/rosterPins.ts` (mirrors the existing `utils/rosterRestrictions.ts` extraction pattern for testability). The new implementation builds a `Set` of every PIN already in use (manual or previously assigned) and advances a running counter past any taken value before handing out a fallback PIN.

## Band-aid rejected

Widening `findDuplicatePins` to warn in the editor UI was considered and rejected — it only nags after the fact, doesn't prevent the collision if the teacher ignores/misses it, or if the save routes through a different path that doesn't hit that specific editor hook (ClassLink re-sync, Drive import merge). Fixing `assignPins()` itself closes the hole at the one chokepoint every save path already goes through.

## Regression test

`tests/utils/rosterPins.test.ts` (5 cases) — key case reproduces the exact blank-then-manual-collision scenario and asserts no duplicate PINs result. Orchestrator independently re-verified: reverted `rosterPins.ts` to the old position-only logic, confirmed 2/5 tests fail (`expected 3 to be 4` / duplicate PINs); restored the fix, confirmed all 5 pass.

## Evidence

Subagent-run `pnpm run validate` + `pnpm run build` both green (root 6991/6991, functions 706/706).

## Risk

**Contained** — pure, localized change to a fallback-value algorithm inside one utility function; no call-signature or consumer-behavior change beyond the returned PINs now being guaranteed collision-free.

## Backlog (investigated, ruled out)

- `useCatalystSets.ts:46`'s lexicographic sort — confirmed unreachable: the admin UI hard-caps catalyst sets at 4, so a 10th set (needed to trigger the mis-sort) can never be created.

---
🤖 Nightly code-health run — automated bug hunt + orchestrator review.

---
_Generated by [Claude Code](https://claude.ai/code/session_01M33BZjYmSCA3LhUBqhaCX3)_

### DIFF
--- a/hooks/useRosters.ts
+++ b/hooks/useRosters.ts
@@ -17,6 +17,7 @@ import { db, functions, isAuthBypass } from '@/config/firebase';
 import { useGoogleDrive } from '@/hooks/useGoogleDrive';
 import { getLocalIsoDate } from '@/utils/localDate';
 import { mapWithConcurrency } from '@/utils/mapWithConcurrency';
+import { assignPins } from '@/utils/rosterPins';
 
 /**
  * Phase 3 — rebuild the per-roster pin_index sidecar after a roster save.
@@ -81,17 +82,6 @@ async function syncRosterPinIndex(
   } catch (err) {
     console.warn('[useRosters] commitRosterPinIndexV1 failed:', err);
   }
-}
-
-/**
- * Assigns zero-padded sequential PINs to students that don't have one yet.
- * Returns a new array — does not mutate the input.
- */
-function assignPins(students: Student[]): Student[] {
-  return students.map((s, i) => ({
-    ...s,
-    pin: s.pin || String(i + 1).padStart(2, '0'),
-  }));
 }
 
 /**
--- a/tests/utils/rosterPins.test.ts
+++ b/tests/utils/rosterPins.test.ts
@@ -1,0 +1,67 @@
+import { describe, it, expect } from 'vitest';
+import { assignPins } from '@/utils/rosterPins';
+import type { Student } from '@/types';
+
+const student = (id: string, pin = ''): Student => ({
+  id,
+  firstName: id,
+  lastName: '',
+  pin,
+});
+
+describe('assignPins', () => {
+  it('leaves students with an existing pin untouched', () => {
+    const result = assignPins([student('a', '07')]);
+    expect(result[0].pin).toBe('07');
+  });
+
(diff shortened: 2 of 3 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-20

### Summary
Extracts `assignPins` from `hooks/useRosters.ts` into `utils/rosterPins.ts` and hardens it so a position-based fallback PIN can never collide with a PIN already present in the roster (manual or previously assigned). This closes a real identity-resolution bug: a blank student earlier in the roster plus a manually-set `"05"` later would silently produce two students sharing `"05"`, corrupting the PIN-based login/SSO `pin_index` sidecar. Overall a well-scoped, correct fix.

### Code Quality
✓ Clean pure function, no mutation of the input array, no `any`. The `used` Set + advancing `next` counter is correct and O(n) amortized. Import wired in `useRosters.ts` with the local copy removed — no dead export left behind.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No collection or rules changes. Fix improves correctness of the downstream `syncRosterPinIndex` sidecar write.

### Regression Risk
✓ Low. Behavior is preserved for the common all-blank roster (`01, 02, 03…`); only the collision case changes, and it changes toward correctness. No call-signature change.

### Missing Tests
✓ Adequate. Five cases including the exact blank-then-manual collision scenario and the multi-manual skip case; asserts uniqueness via `new Set(pins).size`.

### Overall
**Merge readiness:** Ready
Focused, correct, well-tested extraction with no behavioral regression for existing rosters.

---

---

PR #2265, opened 2026-07-23
fix(widgets): NextUp auto-expiry never re-checks day rollover without a config write

OPS-PIvers:
## Root cause

`NextUpWidget`'s auto-expiry `useEffect` (`components/widgets/NextUp/Widget.tsx`) compared `config.createdAt`'s date against `new Date()` inside an effect whose dependencies were `[config.isActive, config.createdAt, widget.id, updateWidget, config]` — no time-based dependency. A queue left `isActive: true` on an idle classroom display overnight (no student joining, no other config write) never re-ran the day check, so it stayed "active" indefinitely past midnight instead of auto-expiring.

Same root-cause class as two previously-fixed bugs in this codebase: `CountdownWidget` (#1774) and `CalendarWidget` (#1955) — a date comparison with no ticking dependency to re-trigger it across a midnight boundary.

## Fix

Extracted the comparison into a pure, testable `shouldExpireNextUpQueue(isActive, createdAt, now)` in `nextUpQueueUtils.ts`, and added a `nowTick` state ticked every 60s via `setInterval` (mirrors `CountdownWidget`'s `todayMidnightMs` pattern), wired into the effect's dependency array.

## Band-aid rejected

Patching the effect to also fire on the existing Firestore `entries` listener callback (which already calls `updateWidget` when students join). Rejected because it doesn't cover the actual failure scenario — a queue with zero student activity overnight would still never re-check. The fix needed a real time-based tick, not a symptom patch tied to unrelated activity.

## Regression test

`components/widgets/NextUp/Widget.test.tsx` — "auto-expires an idle overnight session once the clock ticks past midnight", plus 5 new pure-function unit tests for `shouldExpireNextUpQueue` in `nextUpQueue.test.ts`.

- **Fail-before** (orchestrator-verified against `dev-paul` source): `TypeError: shouldExpireNextUpQueue is not a function` / assertion failures — 6 failed, 9 passed (15 total).
- **Pass-after**: 15/15 passed.

## Validation

`pnpm run validate` (type-check root+functions, lint `--max-warnings 0`, format-check, root tests 581 files/7052 tests, functions tests 38 files/719 tests, test:counts guard) and `pnpm run build` both pass with exit code 0. Orchestrator independently re-ran the fail-before/pass-after swap against `dev-paul` and confirmed the same result.

## Risk

**Contained** — single widget, isolated pure-function extraction + one new `setInterval` ticker, no cross-widget or persisted-schema impact.

---

### DIFF
--- a/components/widgets/NextUp/Widget.test.tsx
+++ b/components/widgets/NextUp/Widget.test.tsx
@@ -1,0 +1,121 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/**
+ * Regression test for the NextUp auto-expiry day-rollover check.
+ *
+ * Bug: the widget's auto-expiry `useEffect` only re-ran when
+ * `config.isActive`/`config.createdAt` changed identity — there was no
+ * time-based re-trigger. A session left active on an idle classroom display
+ * overnight (no student joining the queue, no other config write) stayed
+ * "active" indefinitely past midnight instead of auto-expiring at the next
+ * calendar day, because nothing ever caused the effect to re-run and
+ * re-evaluate `new Date()` against the stored `createdAt`. Same root cause
+ * as CountdownWidget (#1774) and CalendarWidget (#1955): a date comparison
+ * with no ticking dependency.
+ *
+ * Fix: added a `nowTick` state that ticks every 60s via `setInterval` and is
+ * included in the auto-expiry effect's dependency array, so the day
+ * comparison (extracted to the pure, independently-tested
+ * `shouldExpireNextUpQueue` in nextUpQueueUtils.ts) re-runs periodically
+ * regardless of whether `config` changed.
+ */
+import '@testing-library/jest-dom';
+import React from 'react';
+import { act, render } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { NextUpWidget } from './Widget';
+import { NextUpConfig, WidgetData } from '@/types';
+
+const mockUpdateWidget = vi.fn();
+
+vi.mock('@/context/useDashboard', () => ({
+  useDashboard: () => ({
+    updateWidget: mockUpdateWidget,
+    activeDashboard: { widgets: [] },
+  }),
+}));
+
+vi.mock('@/context/useAuth', () => ({
+  useAuth: () => ({ user: { uid: 'teacher-1' } }),
+}));
+
+vi.mock('@/hooks/useGoogleDrive', () => ({
+  useGoogleDrive: () => ({ driveService: null }),
+}));
+
+const buildWidget = (config: Partial<NextUpConfig>): WidgetData =>
+  ({
+    id: 'nextup-widget',
+    type: 'nextUp',
+    x: 0,
+    y: 0,
+    w: 350,
+    h: 500,
+    z: 1,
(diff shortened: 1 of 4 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-23

### Summary
This PR fixes a real day-rollover bug in `NextUpWidget`: the auto-expiry `useEffect` had no time-based dependency, so an idle overnight queue never re-evaluated `new Date()` and stayed `isActive: true` past midnight. The fix extracts the comparison into a pure `shouldExpireNextUpQueue(isActive, createdAt, now)` helper and drives it from a `nowTick` state ticked every 60s via `setInterval`. The approach mirrors the previously-fixed `CountdownWidget` (#1774) / `CalendarWidget` (#1955) pattern and is well-scoped.

### Code Quality
- ✓ `useState`/`useEffect` already imported — no import churn beyond the added helper.
- ✓ `setInterval` is correctly cleaned up via `clearInterval` in the effect return; ticker effect uses an empty `[]` dep array (correct — it only sets up the external timer).
- ✓ `nowTick` uses a lazy `useState(() => Date.now())` initializer; `new Date(nowTick)` is passed to the pure helper.
- ✓ Auto-expiry effect deps are exhaustive-deps-correct (`nowTick` added). `useEffect` here is a legitimate external-system (timer) use, not derived-state — consistent with project convention.
- ✓ No floating promises, no `console.log`, no `any` introduced. The pre-existing redundant `config` + `config.isActive`/`config.createdAt` deps are carried over unchanged (harmless, not introduced by this PR).
- Minor: expiry can lag up to ~60s past midnight (interval granularity). Acceptable for this use case.

### Widget Standards
- ✓ N/A — no new widget, no front-face rendering changes, no container-query/CSS scaling touched. No `text-*`/fixed-size regressions introduced.

### Admin Config Standards
- ✓ N/A — no admin config, `feature_permissions`, or building-defaults changes. Bug fix only.

### Firestore & Security
- ✓ No Firestore schema, rules, or query changes. `updateWidget` call is unchanged in shape (`{ config: { ...config, isActive: false } }`); no new persisted fields. `nowTick` is local component state, never written to Firestore.

### Regression Risk
- ✓ Contained to a single widget: one isolated pure-function extraction plus one 60s ticker. No cross-widget or persisted-schema impact. Ticker runs for the widget's lifetime even while inactive, but the work per tick is trivial (a `setState` + a `toDateString` compare).
- Note: PR is still marked **draft** — confirm it's intended for merge before proceeding.

### Missing Tests
- ✓ Good coverage. 6 new tests: 1 integration test in `Widget.test.tsx` ("auto-expires an idle overnight session once the clock ticks past midnight", plus a same-day no-expiry case) using fake timers, and 5 pure-function unit tests in `nextUpQueue.test.ts` covering today/rolled-over/inactive/missing-`createdAt`/before-vs-after-midnight. Fail-before / pass-after (6 failed → 15/15) is documented.
- Minor verification note: `Widget.test.tsx` mocks `useDashboard`/`useAuth`/`useGoogleDrive` but not `useDialog` (used by the component); PR reports all tests green, so CI has verified this resolves fine.

### Overall
Solid, well-targeted fix with a genuinely pure/testable helper and correct timer cleanup; just clear the draft flag before merging.

---

---

PR #2266, opened 2026-07-23
fix(dock): stop Escape leaking from popovers into the global widget-minimize handler

OPS-PIvers:
## Root cause

`DashboardView.tsx` installs a global `window`-level Escape keydown handler that, when it finds no active typing field and no `.widget` ancestor for the focused element, falls back to targeting the topmost z-index widget and dispatches a `'widget-keyboard-action'` Escape event — which `DraggableWindow` interprets as "minimize this widget."

The dock's `ToolDockItem` "restore minimized widgets" popover had **no Escape handler at all**, and lives outside any `.widget` ancestor (it's portalled). An unhandled Escape bubbled straight through to the global handler. **A teacher opening the dock's restore-widget popover and pressing Escape to dismiss it would instead silently minimize an unrelated widget on their live board** (e.g. a running timer, mid-lesson).

Auditing sibling popovers (`RemoteControlMenu.tsx`, `ClassRosterMenu.tsx`) found the identical leak: both already had `onClose()` Escape handlers but never called `stopPropagation()`, so the same bug occurred there too.

## Fix

All three now call `event.stopPropagation()` in their Escape handler before closing. `ToolDockItem` also gained the handler itself (it had none), plus focus-return to the trigger button on dismiss.

## Band-aid rejected

Just adding a close-on-Escape handler to `ToolDockItem` (matching what `RemoteControlMenu`/`ClassRosterMenu` already had) would still leak propagation to the global handler — exactly the latent bug those two already had. The real root cause is the missing `stopPropagation()`, not the missing close call, so all three needed the same fix rather than patching only the component that started the investigation.

## Regression tests

- `tests/components/layout/ToolDockItem.test.tsx` — "closes the popover when Escape is pressed" + "does not leak the Escape keydown to window-level listeners"
- `tests/components/layout/RemoteControlMenu.escapePropagation.test.tsx`
- `tests/components/layout/ClassRosterMenu.escapePropagation.test.tsx`

**Fail-before** (orchestrator-verified against `dev-paul` source for all 3 files): 4/4 new tests fail — popover stays open / a stand-in `window` keydown spy fires once.
**Pass-after**: 4/4 pass.

## Validation

`pnpm run validate` (583 root test files/7049 tests, 38 functions test files/719 tests, 0 lint/type errors) and `pnpm run build` both pass. Orchestrator independently re-ran the fail-before/pass-after swap against `dev-paul` and confirmed the same result.

## Risk

**Contained** — event-propagation fix confined to 3 popover/menu components; no state-shape or persistence changes.

---

### DIFF
--- a/components/layout/ClassRosterMenu.tsx
+++ b/components/layout/ClassRosterMenu.tsx
@@ -48,6 +48,12 @@ const ClassRosterMenu: React.FC<Props> = ({
     const handleKeyDown = (event: KeyboardEvent) => {
       if (event.key !== 'Escape') return;
       if (isEscapeFromWidgetInput(event)) return;
+      // This menu is portalled outside any `.widget` DraggableWindow, so
+      // without stopping propagation an Escape here also bubbles up to
+      // DashboardView's global window-level Escape handler, which falls
+      // back to minimizing the topmost widget on the board — a completely
+      // unrelated widget disappearing just because the menu was dismissed.
+      event.stopPropagation();
       onClose();
     };
 
--- a/components/layout/RemoteControlMenu.tsx
+++ b/components/layout/RemoteControlMenu.tsx
@@ -69,6 +69,12 @@ const RemoteControlMenu: React.FC<Props> = ({ onClose, anchorRect }) => {
     const handleKeyDown = (event: KeyboardEvent) => {
       if (event.key !== 'Escape') return;
       if (isEscapeFromWidgetInput(event)) return;
+      // This menu is portalled outside any `.widget` DraggableWindow, so
+      // without stopping propagation an Escape here also bubbles up to
+      // DashboardView's global window-level Escape handler, which falls
+      // back to minimizing the topmost widget on the board — a completely
+      // unrelated widget disappearing just because the menu was dismissed.
+      event.stopPropagation();
       onClose();
     };
 
--- a/components/layout/dock/ToolDockItem.tsx
+++ b/components/layout/dock/ToolDockItem.tsx
@@ -1,10 +1,11 @@
-import React, { useState, useRef } from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import { createPortal } from 'react-dom';
 import { RefreshCcw, Trash2, Plus, X } from 'lucide-react';
 import { useSortable } from '@dnd-kit/sortable';
 import { CSS } from '@dnd-kit/utilities';
 import { useClickOutside } from '@/hooks/useClickOutside';
 import { useLongPress } from '@/hooks/useLongPress';
+import { isEscapeFromWidgetInput } from '@/utils/domHelpers';
 import { GlassCard } from '@/components/common/GlassCard';
 import { DockIcon } from './DockIcon';
 import { DockLabel } from './DockLabel';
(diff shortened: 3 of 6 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-23

### Summary
This PR fixes Escape-key leakage from three portalled popovers (`ToolDockItem` restore-widgets popover, `RemoteControlMenu`, `ClassRosterMenu`) into `DashboardView`'s global window-level Escape handler, which would otherwise minimize an unrelated topmost widget. The fix adds `event.stopPropagation()` to each Escape handler and gives `ToolDockItem` a handler it previously lacked (with focus-return to the trigger button). The root-cause reasoning and the decision to fix all three rather than band-aid the one component are correct.

### Code Quality
✓ — The propagation model is sound and verified against source. All three popover handlers attach to `document` in the bubble phase; `DashboardView`'s global handler attaches to `window` in the bubble phase (`window.addEventListener('keydown', ...)`, no capture — confirmed at DashboardView.tsx:1054). Because `document` bubbles before `window`, `stopPropagation()` in the document-level handler reliably prevents the window-level minimize handler — the fix works as claimed. `stopPropagation()` (not `stopImmediatePropagation()`) is the right choice: sibling document-level listeners (e.g. the menus' own click-outside handlers) are unaffected. The new `useEffect` in `ToolDockItem` is a legitimate external-system subscription (not derived-state misuse), guards on `showPopover`, and correctly returns `undefined` vs. the cleanup remover. Deps `[showPopover, buttonRef]` are complete and `buttonRef` is a stable ref. The `isEscapeFromWidgetInput()` early-return before `stopPropagation()` is intentional and consistent across all three — an Escape originating in a widget text input is correctly left to bubble.

### Widget Standards
✓ — Not applicable. No widgets added or restructured; changes are confined to layout/dock popover components. (Note: the referenced `spart-new-widget` skill file was not present at the given path — noted per instructions, no impact on this review.)

### Admin Config Standards
✓ — Not applicable. No admin feature-permission or widget-config surface touched. (`spart-widget-admin-config` skill file also absent at the given path.)

### Firestore & Security
✓ — No Firestore reads/writes, security rules, auth, or persistence changes. Purely a client-side event-propagation fix.

### Regression Risk
✓ — Low and contained. Adding `stopPropagation()` only suppresses the global window-level Escape when one of these menus actively handles the key — the correct "the menu consumed this Escape" semantics. Other document-level Escape consumers still fire (no `stopImmediatePropagation`). Minor consistency note (non-blocking): only `ToolDockItem` adds focus-return to its trigger; `RemoteControlMenu`/`ClassRosterMenu` do not restore focus on Escape dismiss — reasonable to leave out of scope, but a follow-up could align them.

### Missing Tests
✓ — Coverage matches the claim: 4 new tests across 3 files. Each asserts both halves (popover/menu closes AND a stand-in window-level keydown spy does not fire), which genuinely exercises the propagation fix rather than just the close call. The `ToolDockItem` suite avoids mounting the full `DashboardView` by simulating the window listener directly — a sensible isolation choice. `fireEvent.keyDown(document, ...)` targets `document`, which `isEscapeFromWidgetInput` correctly treats as non-widget-input (not an `Element`), so the guard doesn't short-circuit the assertions.

### Overall
**Merge readiness:** Ready
Correct root-cause fix with matching regression tests and no state, persistence, or security surface changes; the only note (focus-return parity across the two menus) is optional and out of scope.

---

---

PR #2267, opened 2026-07-23
fix(auth): decode raw HTML entity in InviteAcceptance default error message

OPS-PIvers:
## Root cause

`components/auth/InviteAcceptance.tsx`'s `getErrorContent()` default-branch error message was authored as a plain JS string literal containing the raw HTML entity `&rsquo;`: `'We couldn&rsquo;t accept this invitation. Try again, or contact your admin.'`. JSX only decodes HTML entities in literal JSX text children, not in ordinary string literals interpolated via `{message}`.

Any `/invite/:token` claim failure whose Firebase callable error code isn't one of the four explicitly handled (`not-found`, `failed-precondition`, `deadline-exceeded`, `permission-denied`) — e.g. `internal`, `unavailable`, `invalid-argument`, or any transient/unexpected code — rendered the literal text "We couldn&rsquo;t accept this invitation…" on screen instead of a real apostrophe.

## Fix

Swapped the string to a double-quoted literal with a real apostrophe. Swept `components/admin/**`, `components/auth/**`, and `config/**` for the same anti-pattern (HTML entities embedded in JS string literals rather than JSX text) to rule out sibling drift — no other instance found; the other `&amp;`-family hits in-repo are legitimate HTML-escaping utility code inside JSX children, unrelated to this bug.

## Band-aid rejected

None needed — already a single-line, one-string root cause with no broader pattern in scope.

## Regression test

`tests/components/auth/InviteAcceptance.errorEntities.test.tsx` — mocks an unhandled Firebase callable error code and asserts the rendered error text contains a real apostrophe, not the literal substring `&rsquo;`.

**Fail-before** (orchestrator-verified against `dev-paul` source): fails — `Received: "We couldn&rsquo;t accept this invitation..."`.
**Pass-after**: passes.

## Validation

`pnpm run validate` (root tests 581 files/7046 tests, functions tests 38 files/719 tests, 0 lint/type errors) and `pnpm run build` both pass. Also ran a key-by-key diff of `plcDashboard`, `importSharedCollection`/`shareCollection` i18n namespaces against DE/ES/FR (all clean, zero drift found — not included in this diff). Orchestrator independently re-ran the fail-before/pass-after swap against `dev-paul` and confirmed the same result.

## Risk

**Contained** — single string literal fix plus one new test file.

---

### DIFF
--- a/components/auth/InviteAcceptance.tsx
+++ b/components/auth/InviteAcceptance.tsx
@@ -249,7 +249,7 @@ const getErrorContent = (
       return {
         title: 'Something went wrong',
         message:
-          'We couldn&rsquo;t accept this invitation. Try again, or contact your admin.',
+          "We couldn't accept this invitation. Try again, or contact your admin.",
       };
   }
 };
--- a/tests/components/auth/InviteAcceptance.errorEntities.test.tsx
+++ b/tests/components/auth/InviteAcceptance.errorEntities.test.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+
+// Regression: InviteAcceptance's default-branch error message is built from a
+// plain JS string literal containing the raw HTML entity `&rsquo;`
+// ("We couldn&rsquo;t accept this invitation..."). Unlike JSX text children
+// (which the JSX transform decodes as HTML), a plain string interpolated via
+// `{message}` is inserted into the DOM verbatim — so users hit with any
+// unrecognized claim-error code see the literal characters "&rsquo;" instead
+// of an apostrophe. Reproduce by forcing the claim callable to reject with a
+// code outside the four explicitly-handled cases (falls through to the
+// "Something went wrong" default branch).
+
+const mockClaim = vi.fn();
+
+vi.mock('@/config/firebase', () => ({
+  functions: {},
+}));
+
+vi.mock('firebase/functions', () => ({
+  httpsCallable: () => mockClaim,
+}));
+
+vi.mock('@/context/useAuth', () => ({
+  useAuth: () => ({
+    user: { uid: 'u1', email: 'teacher@example.com' },
+    loading: false,
+    signOut: vi.fn(),
+  }),
+}));
+
+import { InviteAcceptance } from '@/components/auth/InviteAcceptance';
+
+describe('InviteAcceptance default error message', () => {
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-23

### Summary
This PR fixes a real user-facing bug: the `getErrorContent()` default branch in `components/auth/InviteAcceptance.tsx` held the raw HTML entity `&rsquo;` inside a plain JS string literal, which — unlike JSX text children — is inserted into the DOM verbatim via `{message}`, so any unhandled Firebase claim-error code (`internal`, `unavailable`, etc.) rendered the literal characters on screen. The fix swaps in a real apostrophe in a double-quoted literal, and a new regression test covers the case. The change is a single-line string fix plus one test file — contained and correct.

### Code Quality
✓ Replacement is a genuine U+0027 apostrophe in a double-quoted literal (`"We couldn't accept this invitation..."`), not another entity. The double-quote choice cleanly avoids escaping. The remaining `&rsquo;` occurrences in the file (`we&rsquo;re linking your account`, `You&rsquo;re invited`) are legitimate JSX **text children**, where the JSX transform decodes entities correctly — those are not instances of the bug and were correctly left untouched. No other entity-in-JS-string-literal remains in the touched file.

### Widget Standards
✓ N/A — auth flow component, no widget involved. (Note: the two referenced skill files were not present in this environment; evaluated against repo conventions instead.)

### Admin Config Standards
✓ N/A — no admin/feature-permission surface touched.

### Firestore & Security
✓ No Firestore, security-rule, or auth-logic changes. Error-message text only; no data path or permission impact.

### Regression Risk
✓ Low. Single string-literal change on a rarely-hit default error branch; no control flow, type, or API change. The four explicitly-handled error codes are unaffected.

### Missing Tests
✓ New `tests/components/auth/InviteAcceptance.errorEntities.test.tsx` forces an unhandled callable code (`functions/internal`) to fall through to the default branch and asserts the rendered `textContent` excludes `&rsquo;` and includes `couldn't accept this invitation` — it validates the actual rendered DOM text, which is exactly the right assertion for this bug. Minor note: the shared `mockClaim` also backs the `resolveOrgForUser` call, which rejects and is swallowed by `resolveInviteOrgId`'s fallback before the `claimOrganizationInvite` rejection drives the error state — indirect but valid; the test correctly reaches the default branch.

### Overall
**Merge readiness:** Ready
Correct, minimal, well-tested fix for a genuine rendering bug with negligible regression risk.

---

---

PR #2268, opened 2026-07-23
fix(functions): paginate expireActivityWallShares lapsed-share lookup

OPS-PIvers:
## Root cause

`functions/src/expireActivityWallShares.ts` (the hourly sweep that re-locks an Activity Wall session once every gallery share pointing at it has lapsed) queried `shared_activity_walls` for expired-by-date and revoked shares with a single `.limit(500)` call and **no cursor**. Lapsed share docs are never deleted (owners/admins can still read them for management, mirroring the sibling rule). Once more than 500 lapsed shares accumulated, every subsequent run re-fetched the identical top-500 (by document-ID order) forever, and any session whose only lapsed share sorted past position 500 was **never** relocked.

This is a real, unbounded privacy gap: `publiclyShared` stays `true` on the session doc, so a caller who captured a `sessionId` while a share was live could keep reading every submission and photo indefinitely past the teacher's expiration/revoke, once share volume passed the page cap. Flagged (not fixed) on #2251; this run implements the fix.

## Fix

Paginated both lookups with a `startAfter` cursor, mirroring the identical fix already applied in `gcPlcOrphans.ts` (#2230) for the same bug class:
- Expired-by-date query: `orderBy('expiresAt').orderBy(FieldPath.documentId())` (Firestore requires ordering the range-filtered field first; the document-id tiebreaker needs no additional composite index — Firestore implicitly appends it to any single-field-ordered query).
- Revoked query: `orderBy(FieldPath.documentId())` alone.
- `MAX_LAPSED_SHARES_PER_RUN` (500 → 5000) is now a safety ceiling with a `console.warn` if hit, not a silent hard cap; `LAPSED_SHARE_PAGE_SIZE` (500) is the per-page size.

## Band-aid rejected

Simply raising `MAX_LAPSED_SHARES_PER_RUN` to a bigger fixed number — this only pushes the same silent cliff further out without eliminating it. Rejected in favor of the root-cause cursor-pagination fix already proven correct for this exact bug class in `gcPlcOrphans.ts`.

## Regression tests

`functions/src/expireActivityWallShares.test.ts` — two new tests seeding 505 lapsed shares (expired and revoked variants) with the target session's share sorting last:
- "relocks a session whose only lapsed (expired) share sorts past LAPSED_SHARE_PAGE_SIZE"
- "relocks a session whose only lapsed (revoked) share sorts past LAPSED_SHARE_PAGE_SIZE"

**Fail-before** (orchestrator-verified against `dev-paul` source): both fail — `expected 1 to be NaN` (pre-fix code has no `LAPSED_SHARE_PAGE_SIZE` export, only found the hardcoded 500-doc cap; the target session's share, sorted past it, was left unlocked).
**Pass-after**: 15/15 tests pass in the file — all 505 lapsed shares found, every session (including the one past the old page boundary) relocked.

## Validation

`pnpm run validate` (root tests 580 files/7045 tests, functions tests 38 files/721 tests, 0 lint/type errors) and `pnpm run build` both pass. `firestore.rules` untouched, so `test:rules` doesn't apply here. Also swept `functions/src/` for the sibling SSR F redirect-guard pattern (`maxRedirects: 0` on blocklisted-URL fetches) — confirmed both existing guarded call sites (`fetchExternalProxy`, `checkUrlCompatibility`) are correct and no ungated call site exists. Orchestrator independently re-ran the fail-before/pass-after swap against `dev-paul` and confirmed the same result.

## Risk

**Architectural** — this closes an unbounded-accumulation gap in a privacy-relevant scheduled sweep; tagging architectural (not contained) since it changes the query/pagination contract of a production Cloud Function, even though the change itself mirrors an already-shipped, proven pattern.

---

### DIFF
--- a/functions/src/expireActivityWallShares.test.ts
+++ b/functions/src/expireActivityWallShares.test.ts
@@ -19,7 +19,12 @@ import { describe, it, expect, vi } from 'vitest';
 vi.mock('firebase-admin', () => ({
   apps: [{ name: '[DEFAULT]' }],
   initializeApp: vi.fn(),
-  firestore: vi.fn(),
+  // The sweep paginates with admin.firestore.FieldPath.documentId(); expose
+  // it as the '__name__' sentinel so the stub CollectionRef.orderBy gets a
+  // stable value (mirrors gcPlcOrphans.test.ts's stub).
+  firestore: Object.assign(vi.fn(), {
+    FieldPath: { documentId: vi.fn(() => '__name__') },
+  }),
 }));
 
 vi.mock('firebase-functions/v2', () => ({
@@ -33,6 +38,7 @@ vi.mock('firebase-functions/v2/scheduler', () => ({
 import {
   isLapsedShare,
   runExpireActivityWallShares,
+  LAPSED_SHARE_PAGE_SIZE,
 } from './expireActivityWallShares';
 
 const NOW = 1_700_000_000_000;
@@ -77,11 +83,16 @@ interface StubShareDoc {
 /**
  * In-memory Firestore mirroring the Admin SDK surface
  * `runExpireActivityWallShares` uses:
- *   db.collection('shared_activity_walls').where(f, op, v).limit(n).get()
+ *   db.collection('shared_activity_walls').where(f, op, v)
+ *     .orderBy(field).limit(n).startAfter(docSnap).get()
  *   db.collection('activity_wall_sessions').doc(id).update(data)
  * Only the where clauses this sweep issues are supported (le/eq on
  * expiresAt/revoked/sessionId) — enough to faithfully exercise the sweep
- * without pulling in the Firestore emulator.
+ * without pulling in the Firestore emulator. `orderBy`/`startAfter` mirror
+ * `gcPlcOrphans.test.ts`'s stub: sort by the ordered field(s) (falling back to
+ * doc id for the '__name__' sentinel) and slice past the cursor doc's
+ * ordering-field values — this is what actually exercises cross-page
+ * pagination in the regression test below.
  */
 function makeStubDb(seed: {
   shares?: StubShareDoc[];
@@ -93,10 +104,10 @@ function makeStubDb(seed: {
   };
   const updates: Record<string, Record<string, unknown>[]> = {};
 
-  function matches(
-    doc: StubShareDoc,
-    filters: Array<[string, string, unknown]>
-  ): boolean {
+  type Filter = [string, string, unknown];
+  type DocSnap = { id: string; data: () => Record<string, unknown> };
+
+  function matches(doc: StubShareDoc, filters: Filter[]): boolean {
     return filters.every(([field, op, value]) => {
       const actual = doc.data[field];
       if (op === '<=')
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-07-23

### Summary
This PR closes a real, unbounded privacy gap in the hourly `expireActivityWallShares` Cloud Function: both `shared_activity_walls` lookups (expired-by-date and revoked) previously used a single un-cursored `.limit(500)`, so once more than one page of never-deleted lapsed shares accumulated, any session whose only lapsed share sorted past position 500 was never relocked (`publiclyShared` stayed `true`). The fix introduces a `startAfter` cursor loop (`LAPSED_SHARE_PAGE_SIZE=500` per page, `MAX_LAPSED_SHARES_PER_RUN` raised 500→5000 as a warn-on-hit ceiling) that faithfully mirrors the already-shipped `gcPlcOrphans.ts` (#2230) fix for the identical bug class. Two well-targeted regression tests seed 505 lapsed shares (expired + revoked variants) with the target sorting past the first page.

### Code Quality
Clean. `fetchLapsedSharesPaginated` is correct and readable; the pagination loop terminates unconditionally (each iteration either breaks on `page.size === 0`, breaks on a short final page `< LAPSED_SHARE_PAGE_SIZE`, or advances `visited` by a full page toward the 5000 ceiling — max ~10 iterations, no infinite-loop path), and the cursor always advances via `startAfter(lastDoc)` so no page is re-fetched. `console.warn` is lint-clean (`functions/eslint.config.mjs` sets `no-console: 'off'`) and matches the reference sweep's usage; no TS/ESLint issues. Minor cosmetic (also present in `gcPlcOrphans.ts`, not a blocker): when the total lapsed count is an exact multiple of the page size at the ceiling, the final full page doesn't trigger the short-page break, so the `visited >= MAX` warn can fire as a benign false positive without an extra empty-page probe.

### Widget Standards
✓ (N/A — backend Cloud Function; no widget, UI, or container-query scaling surface touched.)

### Admin Config Standards
✓ (N/A — no admin config modal, feature-permission, or building-defaults wiring in scope.)

### Firestore & Security
✓ Query shapes are index-compatible: the expired query correctly orders the range-filtered field first (`where('expiresAt','<=',now).orderBy('expiresAt').orderBy(documentId())`, satisfying Firestore's inequality-field-ordered-first rule), and the revoked query is `where('revoked','==',true).orderBy(documentId())`; the `__name__` tiebreaker is implicitly appended and needs no additional composite index. `startAfter(lastDoc)` passes the full snapshot so cursor values are extracted for all ordered fields regardless of query shape. `firestore.rules` correctly untouched. This directly remediates the privacy gap (stale `publiclyShared: true`), and the two paginated queries run in parallel via `Promise.all` well within the scheduled-function timeout.

### Regression Risk
The exported `runExpireActivityWallShares(db, now)` signature and its `{ lapsedShares, sessionsRelocked }` return shape are unchanged, and the `onSchedule` wrapper is untouched — no downstream break. `MAX_LAPSED_SHARES_PER_RUN` changes value (500→5000) and semantics (hard cap → warn-on-hit ceiling); intended and safe. **Coordination note:** the PR targets `dev-paul`, whose current tip still holds the pre-pagination version, and sibling PR #2262 (`dev-paul`→`main`) carries that same pre-pagination file. Land #2268 into `dev-paul` *before* #2262 promotes to `main`, otherwise production ships the still-vulnerable version first. Also note this PR is still marked **draft**.

### Missing Tests
✓ Strong, on-point coverage — both regression tests exercise true cross-page pagination through the stub's `startAfter`/`orderBy` and assert the past-the-boundary session is relocked (verified fail-before / pass-after per the description). Optional-only: no test drives the `MAX_LAPSED_SHARES_PER_RUN` ceiling / `console.warn` path (low value; mirrors the untested ceiling branch in `gcPlcOrphans.ts`).

### Overall
Technically sound and faithful to a proven pattern; before merge, undraft and sequence the `dev-paul`/#2262 merge order so the fix reaches `main` ahead of (or together with) the pre-pagination version.

---

---

PR #2315, opened 2026-07-31
nightly(build-tooling): stop generateVideoActivity tests from hitting the real Gemini API

OPS-PIvers:
## Root cause
`functions/src/index.test.ts`'s `generateVideoActivity — accessLevel enforcement` describe block never mocked `@google/genai`. An admin caller short-circuits every permission/usage-limit gate in `generateVideoActivity` (`functions/src/aiGeneration.ts` ~line 1550-1660) and reaches the real `new GoogleGenAI({apiKey}).models.generateContent(...)` call. Since `GEMINI_API_KEY.value()` resolves to the truthy literal `'mock-GEMINI_API_KEY'` (via the `firebase-functions/params` mock), the `if (!apiKey)` guard never trips, so the SDK makes a genuine HTTPS round-trip to `generativelanguage.googleapis.com`. Google normally rejects the fake key in well under 100ms, but under CPU/network contention (multiple concurrent nightly worktrees, which is normal operating conditions for this repo's nightly harness) that round-trip can exceed Vitest's 5000ms default test timeout — a real defect (a unit test depending on live network reachability to pass), discovered tonight when it intermittently failed the orchestrator's baseline `pnpm run validate` run.

## Fix
Added `vi.mock('@google/genai', ...)` that preserves every real export (the `Type` enum is used at runtime to build response schemas) but replaces the `GoogleGenAI` class with a mock whose `models.generateContent` is a hoisted, deterministic spy. The two "throws permission-denied" tests now assert the mock was never called (gate short-circuits before any AI call); the two "does not throw" tests assert it was called exactly once (proving the handler reached the AI-call code path — the thing that used to hit the network — without ever touching it).

Implementation note: `GoogleGenAI` is invoked with `new` in `aiGeneration.ts`, so the mock had to be an ordinary `function` expression, not an arrow function (arrow functions have no `[[Construct]]` and throw "is not a constructor").

## Band-aid rejected
Bumping the test timeout — that leaves the test dependent on live network reachability, which is the actual defect, and only makes failures rarer/slower to reproduce rather than eliminating them.

## Regression test
**Fail-before** (orchestrator-verified: took the fixed test file, removed only the `vi.mock('@google/genai', ...)` block, kept the new assertions):
```
FAIL src/index.test.ts > generateVideoActivity — accessLevel enforcement > does not throw accessLevel errors for an admin regardless of accessLevel setting
AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
Test Files  1 failed (1)
     Tests  1 failed | 3 passed | 72 skipped (76)
```
(Confirms: without the mock, the real network path is taken and the mock spy is never invoked.)

**Pass-after** (orchestrator-verified, full fix restored):
```
Test Files  1 passed (1)
     Tests  4 passed | 72 skipped (76)
```
Error trace under the fix now shows `Error: generateContentMock: no response configured` — proof the real network call no longer happens at all, vs. the pre-fix trace showing a live `ApiError: API key not valid` response from Google's actual API.

## Validate + build
Full `pnpm run validate` and `pnpm run build:all` both green in the subagent's worktree: type-check (root + functions), lint (0 errors/warnings), format:check, 587 root test files / 7108 tests, 40 functions test files / 775 tests, test:counts guard, production build + functions build succeeded.

## Additional issues noticed, not fixed (left as backlog)
- The sibling "does not throw for a beta user" test's own comment was already inaccurate before this change (claimed it fails due to a mock secret, when it actually throws from an unrelated Firestore-mock gap — `ai_usage` collection doesn't stub `.doc()`). Corrected the misleading comment; did not add the Firestore mock support needed to make that test faithfully exercise the full beta-user + usage-transaction path — a small, separate, pre-existing test-fidelity gap.
- Confirmed `generateWithAI`/`generateGuidedLearning` and other `@google/genai`-using handlers aren't driven through full end-to-end tests in this file today, so weren't exposed to this same flakiness; the new module-level mock also covers them if that changes.

## Risk
**Contained** — test-file-only change (mocking, plus assertion additions), no production code touched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01C2sEXSNMrYg5yBKFCVDEe2)_

### DIFF
--- a/functions/src/index.test.ts
+++ b/functions/src/index.test.ts
@@ -85,6 +85,15 @@ const geminiConfigDocGet = vi.fn(() =>
   })
 );
 
+// Hoisted mock for `@google/genai`'s `Models.generateContent`. Any handler
+// that reaches the actual Gemini call (e.g. an admin caller who clears every
+// permission gate) must hit this stub, never the real network — a unit test
+// must never depend on live network reachability to pass. Defaults to a fast
+// rejection; tests that need a specific response call `mockResolvedValueOnce`.
+const generateContentMock = vi.fn(() =>
+  Promise.reject(new Error('generateContentMock: no response configured'))
+);
+
 // Storage mocks for archiveActivityWallPhoto size-guard tests. Each
 // stub is independently resettable so tests can install different
 // metadata responses without rebuilding the whole storage tree.
@@ -373,6 +382,32 @@ vi.mock('firebase-functions/params', () => ({
 
 // Mock axios
 vi.mock('axios');
+
+// Mock `@google/genai` so any code path that reaches the real Gemini call
+// (e.g. an admin caller bypassing every permission gate in
+// `generateVideoActivity`) hits a deterministic, network-free stub instead
+// of the live Generative Language API. Without this, `GEMINI_API_KEY.value()`
+// resolves to the truthy `mock-GEMINI_API_KEY` string (see the
+// `firebase-functions/params` mock above), so the real SDK is constructed
+// and makes a real HTTP round-trip — usually rejected fast by Google with an
+// invalid-key error, but under CPU/network contention that round-trip can
+// exceed the test timeout, making the test flaky rather than deterministic.
+vi.mock('@google/genai', async (importOriginal) => {
+  // Keep every real export (e.g. the `Type` enum used at runtime to build
+  // response schemas) — only the network-calling `GoogleGenAI` class itself
+  // needs to be replaced.
+  const actual = await importOriginal<typeof import('@google/genai')>();
+  return {
+    ...actual,
+    // `GoogleGenAI` is invoked with `new` in aiGeneration.ts — the mock
+    // implementation must be an ordinary function (not an arrow function),
+    // since arrow functions have no [[Construct]] and throw "is not a
+    // constructor" when called via `new`.
+    GoogleGenAI: vi.fn().mockImplementation(function GoogleGenAIMock() {
+      return { models: { generateContent: generateContentMock } };
+    }),
+  };
+});
 
 // Import the function under test
 import {
@@ -3104,6 +3139,8 @@ describe('generateVideoActivity — accessLevel enforcement', () => {
     await expect(handler(VALID_DATA, { auth: NON_ADMIN_AUTH })).rejects.toThrow(
       'Gemini functions are currently restricted to administrators.'
     );
+    // The permission gate must short-circuit before the AI call.
+    expect(generateContentMock).not.toHaveBeenCalled();
   });
 
   it('throws permission-denied for non-beta user when accessLevel is "beta"', async () => {
@@ -3122,14 +3159,16 @@ describe('generateVideoActivity — accessLevel enforcement', () => {
     await expect(handler(VALID_DATA, { auth: NON_ADMIN_AUTH })).rejects.toThrow(
       'You do not have access to Gemini beta functions.'
     );
+    // The permission gate must short-circuit before the AI call.
+    expect(generateContentMock).not.toHaveBeenCalled();
   });
 
   it('does not throw for a beta user when accessLevel is "beta"', async () => {
     // Arrange: caller is in betaUsers — should pass the accessLevel gate.
-    // runTransaction mock won't call its callback, so the function will
-    // resolve (no Gemini API call needed — the function throws from the
-    // try/catch wrapper when the transaction is a no-op). We just need to
-    // confirm it does NOT throw a permission-denied error.
+    // `runTransaction` is a no-op mock (doesn't invoke its callback), so the
+    // usage-counter step is skipped and the handler proceeds to the (mocked)
+    // Gemini call. We just need to confirm it does NOT throw a
+    // permission-denied error.
     geminiConfigDocGet.mockResolvedValue({
       exists: true,
       data: () =>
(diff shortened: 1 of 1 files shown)


functions/src/index.test.ts
```diff
@@ -3171,5 +3210,15 @@ describe('generateVideoActivity — accessLevel enforcement', () => {
...
+    // .generateContent(...)` call in `generateVideoActivity` (aiGeneration.ts)
+    // with no gates in between. Without mocking `@google/genai`, this test
+    // makes a REAL network call to the Generative Language API — normally
+    // rejected fast with an invalid-key error, but under CPU/network
+    // contention that round-trip can exceed the test timeout, producing
+    // flaky failures. Asserting the mock was invoked (once) proves the
+    // handler took the code path that used to hit the network, without ever
+    // actually reaching it.
+    expect(generateContentMock).toHaveBeenCalledTimes(1);
   });
```
OPS-PIvers, line 3223: `generateContentMock` is a module-scoped `vi.fn()` whose call count accumulates across all `it()` calls in the file without an explicit reset. `toHaveBeenCalledTimes(1)` passes today because tests 1–3 in this describe block short-circuit before reaching the Gemini path, but it's fragile: if a future test anywhere in this file triggers the mock before this admin test runs, the count will be off and the assertion will give a false pass or fail.

Adding a `beforeEach(() => generateContentMock.mockClear())` inside this describe block would make the assertion robust against test ordering and future additions.

---

OPS-PIvers (commented):
One test-robustness issue found — see inline comment.

---

functions/src/index.test.ts
OPS-PIvers: Checked this empirically rather than just reasoning about it: this describe block already has its own `beforeEach(() => { vi.clearAllMocks(); ... })` (line ~3115), which resets `mock.calls`/`mock.results` for every `vi.fn()` in the file — including `generateContentMock` — before each of this block's own tests, not just once at file start.

To confirm, I temporarily inserted a test directly before the admin test that called `generateContentMock()` three times and asserted `toHaveBeenCalledTimes(3)`, then ran the block: that test passed (3 calls seen), and the admin test immediately after it *still* passed with `toHaveBeenCalledTimes(1)` — the prior calls didn't leak through, because `clearAllMocks()` fires again in the admin test's own `beforeEach` right before it runs. (Reverted the temp test after confirming — no diff on the branch.)

So the specific fragility described — "a future test anywhere in this file triggers the mock before this admin test runs" — can't actually manifest as long as this local `beforeEach` stays as-is, and every other assertion in this same block (`adminDocGet`, `geminiConfigDocGet` calls, etc.) already depends on that same reset, not just this one. Adding a second, `generateContentMock`-only `mockClear()` would be redundant rather than a fix for a live gap, so I'm leaving it as-is. Appreciate you flagging it though — worth having verified rather than assumed.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-01

### Summary
This PR adds a hoisted `vi.mock('@google/genai')` to `functions/src/index.test.ts` so that any handler reaching the real Gemini call (notably an admin caller in `generateVideoActivity`, who clears every permission gate) hits a deterministic, network-free stub instead of the live Generative Language API. It also adds `expect(generateContentMock)...` assertions that verify the permission gates short-circuit before the AI call and that the admin path reaches it exactly once. The fix is correct, well-scoped, and mirrors the file's existing mock conventions.

### Code Quality
✓ No issues found.
- The mock intercepts the exact runtime path: `aiGeneration.ts` does `new GoogleGenAI({ apiKey })` then `ai.models.generateContent(...)`, and the stub returns `{ models: { generateContent: generateContentMock } }` — shape matches.
- Spreading `...actual` correctly preserves the `Type` enum, which `aiGeneration.ts` imports and uses at runtime to build response schemas. Replacing only `GoogleGenAI` is the minimal correct surface.
- Using a **named regular function** for the `mockImplementation` is correct and necessary — `GoogleGenAI` is invoked with `new`, and an arrow function has no `[[Construct]]`. The inline comment documents this.
- No hoisting/TDZ hazard: `generateContentMock` is referenced only inside the nested closure (executed lazily at `new` time), never during factory execution — matching the established `firebase-admin` factory pattern in the file.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found — test-only change; no production code, rules, or data paths touched.

### Regression Risk
✓ Low risk.
- The mock is file-scoped to `index.test.ts` and does not leak to other test files.
- Mock isolation is sound: each relevant `describe` uses `vi.clearAllMocks()` in `beforeEach`, which resets call counts (so `toHaveBeenCalledTimes` is accurate) without wiping the default `vi.fn(() => Promise.reject(...))` implementation — the correct choice (`resetAllMocks` would have erased the default reject and reintroduced flakiness).
- No other test depended on the real SDK; the only test that previously hit the network (the admin case) is exactly the one this fix stabilizes.
- The default rejection is caught by the handler's `try/catch`, so no unhandled-rejection noise.

### Missing Tests
✓ Adequate coverage (minor, non-blocking notes). The change adds the right guard assertions (`not.toHaveBeenCalled()` on gated paths, `toHaveBeenCalledTimes(1)` on the admin path). A follow-up could add a `mockResolvedValueOnce(...)` happy-path assertion on the parsed success payload now that the mock enables it, but that is outside this PR's scope.

### Overall
**Merge readiness:** Ready
A correct, well-documented, minimally-scoped test-mocking fix that eliminates a real network dependency and its associated flakiness, following the file's existing mock patterns with no regression risk.

---

---

PR #2316, opened 2026-07-31
nightly(state-data): duplicateQuiz/duplicateActivity drop authored Behavior Settings

OPS-PIvers:
## Root cause
In `hooks/useQuiz.ts`'s `duplicateQuiz()` and `hooks/useVideoActivity.ts`'s `duplicateActivity()`, the hand-rolled metadata object for the newly-created copy explicitly preserved `sourceMeta.folderId` but never `sourceMeta.behavior` (the `QuizBehaviorSettings`/`VideoActivityBehaviorSettings` sub-object — tab-switch warnings, shuffle questions/answers, result/correct-answer visibility, speed bonus, attempt limit, etc.). Every other write path in these two hooks (`saveQuiz`/`saveActivity`'s preserve-on-omit fallback, `pullSyncedQuiz`/`pullSyncedVideoActivity`) carefully threads `behavior` through and is thoroughly tested for it — only the Duplicate path silently reset it to defaults. A teacher who configures Behavior Settings (e.g. turns off "show correct answer to student" for exam integrity) and then duplicates that quiz/activity for another class period gets a copy that silently reverts to defaults, with no indication anything changed.

## Fix
Added `...(sourceMeta.behavior !== undefined ? { behavior: sourceMeta.behavior } : {})` to both metadata constructions (`hooks/useQuiz.ts` ~line 572, `hooks/useVideoActivity.ts` ~line 396), mirroring the existing `folderId` pattern and `saveQuiz`'s established `!== undefined` spread convention.

## Band-aid rejected
Could have unconditionally copied `behavior` without the `undefined` guard, but that would write a `behavior: undefined` key rather than omitting it — the codebase treats "no key at all" as meaningfully different from an explicit `undefined` elsewhere in these same files (Firestore semantics), so the guard matches the established convention rather than diverging from it.

## Regression tests
Added to the existing `tests/hooks/useQuiz.behavior.test.ts` and `tests/hooks/useVideoActivity.behavior.test.ts` (established mock-Drive/mock-Firestore harness) — each seeds the mock Drive store via `saveQuiz`/`saveActivity`, builds a `sourceMeta` with `behavior` set, calls `duplicateQuiz`/`duplicateActivity`, and asserts the returned metadata + `setDoc` payload carry the same `behavior`.

**Fail-before** (orchestrator-verified via file-swap of `hooks/useQuiz.ts` and `hooks/useVideoActivity.ts` to `dev-paul`'s pre-fix source):
```
FAIL tests/hooks/useVideoActivity.behavior.test.ts > duplicateActivity — preserves behavior on the copy > carries source.behavior onto the duplicated metadata
AssertionError: expected undefined to deeply equal { sessionMode: 'auto', ... }
Test Files  2 failed (2)
     Tests  2 failed | 19 passed (21)
```

**Pass-after** (orchestrator-verified):
```
Test Files  2 passed (2)
     Tests  21 passed (21)
```

## Validate + build
Full `pnpm run validate` and `pnpm run build` both green in the subagent's worktree: type-check (root + functions), lint (0 errors/warnings), format:check, 587 root test files / 7112 tests, 40 functions test files / 775 tests, test:counts guard, production build succeeded.

## Additional issues noticed, not fixed (backlog candidates)
- `DialogContext.openNext()`'s React-18-batching risk re-traced (4th consecutive night) — mechanism still real, still unreached by any current `showPrompt()` call site.
- `useRosters.ts`'s `assignPins()` display/index drift — traced the ClassLink import path; both `assignPins` + `syncRosterPinIndex` always run together on save, so the persisted/displayed pins can't actually diverge in the traced flow. Still unconfirmed as reachable.
- `DashboardContext.updateWidget()`'s double-`JSON.stringify` diff — measured directly (50KB config, 1000 iterations, ~0.4ms/call); not a bottleneck at realistic widget config sizes. Reporting as measured-clean.
- `utils/adminBuildingConfig.ts`'s `soundboard` case concatenates sound-ID arrays without dedup, but the consuming widget filters by `.includes()`, not by iterating — confirmed harmless.

## Risk
**Contained** — additive `behavior` spread mirroring an existing pattern, plus new tests only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01C2sEXSNMrYg5yBKFCVDEe2)_

### DIFF
--- a/hooks/useVideoActivity.ts
+++ b/hooks/useVideoActivity.ts
@@ -396,8 +396,6 @@ export const useVideoActivity = (
           ...(source.folderId !== undefined
             ? { folderId: source.folderId }
             : {}),
-          // Preserve authored Behavior Settings on duplicate — see the
-          // function doc comment above.
           ...(source.behavior !== undefined
             ? { behavior: source.behavior }
             : {}),
--- a/hooks/useVideoActivity.ts
+++ b/hooks/useVideoActivity.ts
@@ -362,7 +362,9 @@ export const useVideoActivity = (
    * freshly-created Drive file id and roll it back on Firestore failure
    * — `saveActivity` only surfaces the id on success and would leak an
    * orphan Drive file otherwise. Mirrors the rollback path in
-   * `useQuiz.duplicateQuiz`. Reviewer flag from PR #1587.
+   * `useQuiz.duplicateQuiz`. Preserves `folderId` AND `behavior` from the
+   * source, same rationale as `useQuiz.duplicateQuiz`. Reviewer flag from
+   * PR #1587.
    */
   const duplicateActivity = useCallback(
     async (source: VideoActivityMetadata): Promise<VideoActivityMetadata> => {
@@ -393,6 +395,11 @@ export const useVideoActivity = (
           // Preserve folder placement on duplicate.
           ...(source.folderId !== undefined
             ? { folderId: source.folderId }
+            : {}),
+          // Preserve authored Behavior Settings on duplicate — see the
+          // function doc comment above.
+          ...(source.behavior !== undefined
+            ? { behavior: source.behavior }
             : {}),
         };
         await setDoc(
--- a/hooks/useQuiz.ts
+++ b/hooks/useQuiz.ts
@@ -527,11 +527,13 @@ export const useQuiz = (userId: string | undefined): UseQuizResult => {
   /**
    * Duplicate a quiz. Loads canonical content from the source's Drive
    * file, then writes a fresh quiz (new id, new Drive file, new title
-   * suffix). Preserves `folderId` when the source carries one — so a
-   * teacher who duplicates a quiz inside a folder sees the copy land
-   * in the same folder, matching their mental model. Does NOT inherit
-   * `sync` linkage (duplicates are private forks; PLC sharing is a
-   * separate explicit step).
+   * suffix). Preserves `folderId` AND `behavior` when the source carries
+   * them — so a teacher who duplicates a quiz inside a folder, or one
+   * with custom Behavior Settings (tab warnings, shuffle, result
+   * visibility, etc.), gets a copy that matches their mental model of
+   * "an independent copy of what I had," not a copy reset to defaults.
+   * Does NOT inherit `sync` linkage (duplicates are private forks; PLC
+   * sharing is a separate explicit step).
    *
    * The write is hand-rolled (not via `saveQuiz`) so we can observe the
    * freshly-created `driveFileId` and roll it back if the Firestore
@@ -569,6 +571,9 @@ export const useQuiz = (userId: string | undefined): UseQuizResult => {
           // field); reading it back yields the same shape.
           ...(sourceMeta.folderId !== undefined
             ? { folderId: sourceMeta.folderId }
+            : {}),
+          ...(sourceMeta.behavior !== undefined
+            ? { behavior: sourceMeta.behavior }
             : {}),
         };
         await setDoc(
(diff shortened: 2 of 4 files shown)


hooks/useVideoActivity.ts
```diff
@@ -394,6 +396,11 @@ export const useVideoActivity = (
           ...(source.folderId !== undefined
             ? { folderId: source.folderId }
             : {}),
+          // Preserve authored Behavior Settings on duplicate — see the
```
OPS-PIvers: The function's JSDoc was already updated this PR to explain the `behavior` preservation rationale — this inline comment duplicates it. The equivalent spread added to `useQuiz.ts` in this same PR has no inline comment. Per project conventions, comments should be reserved for non-obvious WHY that doesn't already appear in the surrounding doc comment. Remove these two lines to keep the two hooks consistent.

---

OPS-PIvers (commented):
One style nit — see inline comment.

---

hooks/useVideoActivity.ts
OPS-PIvers: Confirmed and fixed in `4eff2172` — removed the redundant comment so `useVideoActivity.ts` matches `useQuiz.ts`'s convention (no inline comment on this spread; the JSDoc already covers it).

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-01

### Summary
This PR fixes `duplicateQuiz`/`duplicateActivity` dropping authored Behavior Settings by conditionally spreading the source's `behavior` field onto the freshly-written metadata doc, mirroring the pre-existing `folderId` preservation pattern. The fix is correct, symmetric across both hooks, and backed by matching new tests. It is a minimal, additive change with low regression risk.

### Code Quality
✓ No issues found. The conditional-spread idiom (`...(sourceMeta.behavior !== undefined ? { behavior: sourceMeta.behavior } : {})`) is identical to how `folderId`/`sync` are already handled in `saveQuiz`/`pullSyncedQuiz`, so the two duplication paths stay consistent. The undefined-omit guard correctly avoids writing an explicit `behavior: undefined` key, and the deliberate decision to still NOT carry `sync` linkage onto a duplicate is preserved.

Minor (non-blocking): `behavior` is a nested object (contains `sessionOptions`), and the spread copies the reference, so in memory `duplicated.behavior === sourceMeta.behavior`. This is not a persistence bug — `setDoc` serializes each doc independently, so stored copies are fully separate — and it matches how `saveQuiz`/`pullSyncedQuiz` already return behavior by reference. Worth awareness only if a future consumer mutates a returned `behavior` object in place.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. Writes target the owner-scoped `users/{userId}/{quizzes|video_activities}/{id}` collections under an existing `if (!userId)` auth guard; no rules, scopes, or shared/global collections are touched. The Drive-file rollback-on-Firestore-failure path is unchanged.

### Regression Risk
✓ Low risk. Purely additive — only adds a field when the source carries one, no-op otherwise. Duplicates with no behavior settings produce byte-identical output to before (verified by the "does NOT write a behavior key" tests). No public signatures changed.

### Missing Tests
✓ Adequate coverage. Each hook gains two focused tests: (1) `behavior` carried onto both the returned metadata and the `setDoc` payload, and (2) no `behavior` key written when the source has none.

### Overall
**Merge readiness:** Ready
Correct, symmetric, well-tested bug fix that restores expected "duplicate preserves my settings" behavior with negligible regression surface.

---

---

PR #2324, opened 2026-08-01
fix(schedule): add events below the list, drop delete confirm, keep settings open

OPS-PIvers:
## Problem

Building a schedule by hand forced a scroll on every single event. The **Add** button sat *above* the event list while a **New Schedule** button sat *below* it — so the control nearest the last row was the one that did **not** add an event.

Separately, deleting one event opened a confirmation dialog, and confirming it made the entire settings panel disappear — which read as "I just lost everything I typed."

## Root cause of the disappearing settings panel

`SettingsPanel.tsx:186-210` closes the panel on any `pointerdown` outside the panel and its widget, unless the target matches `[data-settings-exclude]`. `DialogContainer` portals its overlay onto `document.body` and did **not** carry that marker — so a click on any button inside any confirm/alert/prompt registered as a click-outside.

This was never Schedule-specific: **every** widget that opens a dialog from its settings had the same bug (including Schedule's own "Delete Schedule" confirm). Fixed at the source rather than worked around in the widget.

## Changes

**Event add placement** (`Schedule/Settings.tsx`)
- Controls split by what they act on: `Sort` and `Import` (whole-list operations) stay above the list; `Add Event` and `Today Only` moved directly under the last row, where the new row appears.
- A freshly added row focuses its task field and scrolls into view, so you can type immediately.

**"New Schedule" disambiguation**
- Consolidated to a single `FolderPlus` button in the schedule-name row, beside the existing delete-schedule button — the one row that is about the schedule rather than its events.
- The bottom "New Schedule" button is removed, so nothing below the event list operates on schedules.

**Single-event delete**
- No confirmation dialog. The row goes immediately, with an **Undo** toast as the safety net (10s).
- Undo re-inserts at the original index into the **live** list (read through a ref), so edits made during the toast window aren't clobbered by a stale snapshot.
- Blank, never-filled rows delete silently — an Undo offer for an empty row is noise.
- Deleting a **whole schedule** keeps its confirm; that destroys every event in it.

**Panel-close bug** (`DialogContainer.tsx`, `DashboardView.tsx`)
- The dialog overlay now carries `data-settings-exclude`.
- The toast stack carries it too, so clicking "Undo" doesn't close the panel either. Side effect: dismissing any toast by clicking it no longer closes an open settings panel.

## Testing

- New regression coverage on **both sides** of the contract: `SettingsPanel` honouring the marker, and `DialogContainer` actually emitting it — so the fix can't silently rot from either end.
- New Schedule settings tests: add-button DOM position relative to the last row, "New Schedule" living in the header row, instant delete with an Undo toast, undo-restores-at-original-index, and silent blank-row delete.
- `pnpm run type-check` clean; ESLint clean at `--max-warnings 0`; Prettier clean.

## Notes for review

Two judgement calls worth a second opinion:
- Blank rows delete with no toast at all (rather than always offering Undo).
- The toast-stack exclusion is deliberately whole-stack rather than action-toasts-only, which changes click-to-dismiss behavior slightly beyond the reported bug.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JGUy1WRaYxhkrnGc3VAcho)_

### DIFF
--- a/components/common/DialogContainer.tsx
+++ b/components/common/DialogContainer.tsx
@@ -487,6 +487,12 @@ export const DialogContainer: React.FC = () => {
 
   return createPortal(
     <div
+      // Dialogs render in a portal on document.body, so a pointerdown on this
+      // overlay reads as a "click outside" to any open widget SettingsPanel and
+      // would close it out from under the dialog that it opened. Marking the
+      // overlay excludes it from that check, so confirming/cancelling a dialog
+      // leaves the settings panel underneath exactly as it was.
+      data-settings-exclude
       className="fixed inset-0 z-dialog flex items-center justify-center p-4 bg-slate-900/60 backdrop-blur-sm animate-in fade-in duration-150"
       onClick={() => {
         // Clicking the backdrop dismisses alerts, cancels confirms/prompts
--- a/components/layout/DashboardView.tsx
+++ b/components/layout/DashboardView.tsx
@@ -104,6 +104,10 @@ const ToastContainer: React.FC = () => {
   // unpredictably or drop entirely.
   return (
     <div
+      // Toasts sit outside every widget's DOM subtree, so clicking one (e.g. an
+      // "Undo" action) would otherwise register as a click outside an open
+      // SettingsPanel and close it. Exclude the whole stack from that check.
+      data-settings-exclude
       className="fixed z-toast space-y-3 pointer-events-none"
       style={{
         top: 'calc(1.5rem + env(safe-area-inset-top, 0px))',
--- a/components/widgets/Schedule/Settings.tsx
+++ b/components/widgets/Schedule/Settings.tsx
@@ -1,4 +1,10 @@
-import React, { useState, useMemo, useEffect, useCallback } from 'react';
+import React, {
+  useState,
+  useMemo,
+  useEffect,
+  useCallback,
+  useRef,
+} from 'react';
 import { useDashboard } from '@/context/useDashboard';
 import { useDialog } from '@/context/useDialog';
 import { useFeaturePermissions } from '@/hooks/useFeaturePermissions';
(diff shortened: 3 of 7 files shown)


OPS-PIvers:
Confirmed and fixed in `dd910d6` — thanks, this was a real bug.

I verified it against the code first: five call sites move the selection (`Settings.tsx` lines 167, 213, 235, 282, 732 pre-fix), and rows do unmount on selection change, so returning to a schedule re-mounted the last-added row with `autoFocus` still true.

**I took a different approach than suggested.** Adding `setAutoFocusItemId(null)` at each site is correct in effect, but it needs the line in five places and silently regresses the moment a sixth is added. Instead the autofocus target now carries the schedule it belongs to, and is dropped during render when the effective selection differs:

```ts
const [autoFocusItem, setAutoFocusItem] = useState<{
  scheduleId: string;
  itemId: string;
} | null>(null);

// ...after effectiveSelectedId is computed:
if (autoFocusItem && autoFocusItem.scheduleId !== effectiveSelectedId) {
  setAutoFocusItem(null);
}
```

One place, catches every path including future ones. It's React's "adjusting state while rendering" pattern — the setter resolves before paint, so a stale target is never handed to a child — and it's what `CLAUDE.md` prescribes for reset-on-change over an effect.

Two notes on the test:

- I verified the regression test **fails with the reset removed** and passes with it, so it isn't vacuous.
- It surfaced that jsdom implements neither `scrollTo` nor `scrollIntoView`; the file already stubbed the former, so I extended that precedent rather than adding a `typeof` guard to production code.

Full suite green at 588 files / 7128 tests, plus type-check, ESLint at `--max-warnings 0`, and Prettier.

The two judgement calls flagged in the PR description are still open for a second opinion: blank rows deleting with no toast at all, and the toast exclusion being whole-stack rather than action-toasts-only.

---

PR #2329, opened 2026-08-02
docs: multilingual pronunciation engine — exploration map

OPS-PIvers:
Adds `docs/multilingual-pronunciation-engine.md` — a decision map for adding spoken-response items to the quiz widget, grounded against the current codebase rather than the spec alone.

**Docs only. No feature code, no dependency changes.**

## Resolved

- **Gate granularity** → `GlobalFeature` + `global_permissions` with `accessLevel: 'beta'` + `betaUsers[]`. Widget-level `FeaturePermission` is the wrong tool — it's keyed by `WidgetType` and would hide the entire quiz widget from everyone not on the list. `video-activity-audio-transcription` is the closest precedent.
- **Graded vs ungraded** → teacher-selectable per assignment, as an additive field on `QuizAssignmentSettings` (`types.ts:3980`), alongside `attemptLimit`/`dueAt`. Note `assignment-modes` is a different concept: `'submissions' | 'view-only'`, admin-set org-wide.

## Blocking (D1)

Whether an LLM told the target text biases its phoneme transcription toward the expected answer. If it does, the server-side path is disqualified — a biased scorer silently erases the exact errors the feature exists to catch, and tells nearly every student "great job."

A four-condition probe is designed and built: hold the audio constant, vary only the *claimed* target. Ground truth becomes irrelevant — if reported phonemes move when only the prompt moves, that shift is the bias.

It could not be run here. There's no `GEMINI_API_KEY` in the container; the key lives in Secret Manager via `defineSecret` and only materializes inside deployed functions. Running it needs the variable set in environment settings and a fresh session. Harness details and regeneration commands are in §8 of the doc.

## G2P findings (espeak-ng evaluated directly)

It reproduces the spec's own cited examples for Spanish, German, and English — including **both** English cases (`rough` → `ɹˈʌf`, `through` → `θɹˈuː`) without CMUDict or a fallback neural G2P, a subsystem the spec budgets for. Mandarin is broken (`马` → `(en)ma5θɹˈiː5(cmn)`).

Two findings change the spec:

1. **Dialect must be a selectable field, not a constant.** `cielo` → `θjˈelo` under `es` but `sjˈelo` under `es-419`. One default means marking Spanish-speaking students wrong for pronouncing correctly.
2. **espeak-ng is GPL-3.** Needs a legal read before any espeak-derived code ships — and it cuts against the client-side path D1 may otherwise force.

## Still open

D2 (voice data privacy/retention — largely collapses if D1 lands client-side), D4 (`'Speak'` question type vs separate activity), D7 (30 concurrent recorders).

§9 lists what's safe to build before D1 resolves — the alignment engine is a pure zero-dependency function, identical under either outcome.

---
_Generated by [Claude Code](https://claude.ai/code/session_0137yLR1cq9YwPBy8PdzZTvJ)_


OPS-PIvers:
Both spike notes taken; the `isWrittenQuestionType` suggestion I've recorded as the opposite, because following it would introduce a bug. Pushed in 38d41e9.

**Statistical power — fixed.** You're right that n=5 admits a 2/5 vs 4/5 split, which is Fisher p≈0.5 and not evidence of anything. Default `--runs` is now 10, the verdict prints raw counts (`3/10`) rather than just percentages, and any run below 10 emits an explicit under-powered warning.

**Binary WAVs — agreed, no action.** 87KB total with documented regeneration; worth revisiting in a cleanup pass once D1 resolves.

**`isWrittenQuestionType()` — recorded as a trap rather than a template.** The framing "it's the existing precedent for *this type skips the standard auto-grader*" is the part that doesn't hold. Its actual contract (`types.ts:3009`) is *"requires manual teacher grading — there is no auto-grader for student responses"*, and callers depend on that literally:

- `QuizStudentApp.tsx:3235` — `autoGradedQuestionIds = publicQuestions.filter((q) => !isWrittenQuestionType(q.type))`. The negation **is** the definition of auto-graded.
- `QuizStudentApp.tsx:3430` — the same predicate decides whether to read a teacher-entered grade out of `myResponse.grading[q.id]`.

`'Speak'` *is* auto-graded — by the pronunciation engine rather than by string comparison. Adding it to that predicate would drop spoken items out of the auto-graded count and route them to a manual grading UI where no teacher entry ever arrives. It's a third category the boolean can't express, and D6 (teacher-selected ungraded) adds a fourth.

So D4 now carries the sub-decision explicitly: add a parallel boolean, or replace it with a grading-strategy lookup keyed by question type (`'auto-string' | 'auto-phonetic' | 'manual' | 'ungraded'`). The lookup is the better shape but touches every call site, so it's a cost to weigh rather than a free refactor.

Worth flagging that this is exactly the drift `config/featureDefaults.ts` was written to stop — parallel per-type lookups that fall out of sync one feature at a time.

---

OPS-PIvers:
Both taken, in `0f414b7`.

**`VITE_GEMINI_API_KEY` fallback** — correct diagnosis, and I went one step further than either suggestion. Dropping the fallback or just commenting it both leave the underlying trap: `.env.local` is this repo's documented home for the key (CLAUDE.md), so the developer most likely to run this spike — one whose local app already works — is precisely the one for whom it silently does nothing.

The script now reads `.env.local` directly. Resolution order is shell `GEMINI_API_KEY` → shell `VITE_GEMINI_API_KEY` → each from `.env.local`, and it prints which source it used:

```
key source: VITE_GEMINI_API_KEY (.env.local)
model: gemini-2.5-flash   runs: 10
```

Name only, never the value, so run output is safe to paste into a PR comment. A missing key now lists every location checked rather than naming two. Parser is deliberately minimal — `KEY=VALUE`, strips `export` and quotes, not a general dotenv. All three paths verified including shell-wins-over-file precedence.

**0.4 threshold** — fair, and the fix is worth more than a one-liner. It's now a named `BIAS_DELTA` with the reasoning attached, calibrated against two worked splits so a cold reader can judge it without trusting me:

- 9/10 vs 5/10 clears the bar but is Fisher p≈0.14 — suggestive only.
- 10/10 vs 2/10 is p<0.01 — decisive.

So a bare threshold pass now reads as "investigate further," not "D1 is settled." That matters given what hangs off this: a false BIASED sends us down the client-side path and a ~100MB per-device model download.

Also fixed stale `default 5` references in the header left from the earlier `--runs` change.

One correction to the closing line — this isn't "ready to merge once the spike is run." The spike's result belongs in §4 of the exploration doc, and *that* is a follow-up commit. This PR is the map and the harness; merging it doesn't depend on D1 being answered.

---

OPS-PIvers:
Both taken, in `1116f37`. The warning-ordering one is the best catch of this review.

**Warning fired after the spend.** Correct, and it made the warning close to useless: by the time it printed, all `4 × RUNS` calls were gone and the only choice left was whether to distrust a result already paid for. It now prints before the condition loop, names the cost, and says to abort:

```
key source: GEMINI_API_KEY (shell)
model: gemini-2.5-flash   runs: 2

! UNDER-POWERED RUN — about to spend 8 API calls on a result
  that cannot separate bias from noise. At --runs 2, the verdict
  threshold can fire on an ordinary sampling split. Ctrl-C now and
  re-run with --runs 10 (the default) unless you are just smoke-testing.
```

The verdict-time notice stays as a shorter reminder, so a scrolled-back result still carries the caveat.

**"5×" in the doc.** Also correct — mine, introduced when I raised the default and didn't sweep the doc. Two stale spots, not one: §4 described each condition as running 5×, and §8's run command passed `--runs 5`, which would have immediately tripped the very warning the doc introduces. §8 now relies on the default.

One thing worth flagging since it's the kind of error this review process should catch: while fixing §4 I first wrote that the script *"refuses to start below 10 without an explicit acknowledgement."* It doesn't — it warns and continues. I'd described a stricter design than I built. Corrected in the same commit, but the near-miss is the point: a doc claiming a guardrail that isn't there is worse than no doc, because the next person trusts it.

Resolved the two threads from the previous round (`0f414b7`) — key resolution now reads `.env.local`, and `BIAS_DELTA` is named and calibrated.

---

OPS-PIvers:
All three taken, in `ce54430`. All five threads now resolved.

**Response guard — taken, and it matters more than "low-severity".** The failure mode isn't the crash, it's the silence: `parsed.phonemes ?? []` turned a malformed response into `[]`, which `rhoticOf()` tallies as `'?'` — indistinguishable from *"the model reported no rhotic."* A non-answer could sit in the counts looking like a measurement, in a script whose entire output is a measurement. It now throws into the caller's catch and tallies as `'ERR'`, which is visibly not data.

Verified all five paths rather than just the reported one:

| response | result |
|---|---|
| `null` | throws → `ERR` |
| `{"phonemes":["p","e"]}` | ok |
| `{}` | throws → `ERR` |
| `{"phonemes":"pe"}` | throws → `ERR` |
| `not json` | throws → `ERR` (already did) |

**Both line anchors — correct, and I checked each against the file.**

- `AuthContext.tsx:2464` pointed into the middle of a function body. Now cites the public entry `canAccessFeature` (`:2493`) *and* the shared `resolvePermissionAccess` helper (`:2457`) that makes the access-level decision, since a reader wants both.
- `QuizStudentApp.tsx:3430` → `:3429`. Your reading is right about which line is more useful: `:3430` is the ternary that *consumes* the predicate, `:3429` is the `isWrittenQuestionType` call itself, which is what someone greps for.

One correction to the report: the parse site is at line **167–168**, not 913 — the file is 265 lines. The finding was real and the suggested fix was right, so this changes nothing about the outcome, but flagging it because a wrong anchor in a review costs the next reader the same navigation time the anchor fixes above were about.

The doc's D4 analysis you singled out is the part I'd most want scrutinized before implementation too — it's the one finding here that changes the shape of the code rather than the shape of the docs.

---

OPS-PIvers:
Fixed in `107c5e1`. This one was a genuine correctness bug in the measurement, not a nit — good find.

**The failure direction is what makes it serious.** An `ERR` slot contains no observation, but dividing by `RUNS` counted it as *"did not report the tap."* That deflates whichever condition lost more calls, and it fails toward a **false clean bill of health** — exactly the outcome that would send us down the server-side path on a model that is in fact biased.

I built the case table rather than reasoning about it, because my first attempt at a demonstration didn't actually reproduce the bug (with B at 0/10 both versions fire, which proves nothing):

| scenario | old delta | verdict | new delta | verdict |
|---|---|---|---|---|
| no errors, biased model | 0.70 | BIASED | 0.70 | BIASED |
| **D loses 5 calls to errors** | **0.30** | **missed** | **0.80** | **BIASED** |
| **D loses 6 calls to errors** | **0.20** | **missed** | **0.80** | **BIASED** |
| B loses 5 (skew reversed) | 0.80 | BIASED | 0.70 | BIASED |

Rows 2–3 are a fully biased model that the old code scores as clean.

**One deliberate divergence from the suggested patch.** It returns `0` when there are no valid samples; I return `NaN` and exit non-zero. Returning `0` reintroduces the same bug class one level up — it makes *"no data"* look like *"never reported the tap,"* which is the precise conflation being fixed. Verified against a dead key:

```
B (primed "perro") reported the true tap: 0/0 (n/a)  [1 error excluded]
D (unprimed)       reported the true tap: 0/0 (n/a)  [1 error excluded]

NO RESULT. At least one condition produced zero usable samples — every
call errored. Fix the API errors above and re-run; there is nothing to
interpret here.
```

Also took the lopsided-ERR warning, keyed off the smallest compared condition's usable count rather than a fixed ERR threshold, and per-condition output now prints the real denominator plus the excluded error count.

**Implementation notes — all three folded into §9.** One amendment on the unit tests: I've noted they should be written *before* deciding on fixes, because three of the four flagged issues are arguably behaviour changes rather than bugs (tie-break ordering in particular is a defensible choice that just needs to be a deliberate one). Your sequencing point on GPL-3 is the sharper one and I'd missed it — §9.2 is independent of D1 but **not** of the licensing read, so rule-table work committed before that read may need redoing from independent sources. That's now explicit in the doc.

---

OPS-PIvers:
Declining the deletion, and documenting why in a new `scripts/spikes/pronunciation-bias-probe/README.md` (`e19557d`) — which is where you suggested the note should live anyway.

**The .wav files are the stimulus of a controlled experiment, not build output.** The probe's entire design is *hold the audio constant, vary only the prompt*. If the bytes are regenerated per-run, a changed verdict can no longer be attributed to the model rather than to the audio — which is the one thing the experiment exists to isolate.

I measured the regeneration risk rather than asserting it. On espeak-ng 1.51 the documented commands do reproduce the committed files byte-for-byte, but the output is sharply parameter-sensitive:

| command | md5 (12) | size |
|---|---|---|
| `-v es-419 -s 130` (committed) | `20edab1676c3` | 41772 B |
| `-v es-419 -s 131` | `ff4521508724` | 40926 B |
| `-v es -s 130` | `eaceaaa1650a` | 38556 B |

A one-unit speed typo yields different audio. More seriously, dropping `-419` yields **Castilian instead of Latin American Spanish** — a different dialect, which per §5 of the doc *changes what counts as a correct pronunciation*. That failure is silent: the probe would run cleanly and hand back a confident verdict about the wrong stimulus. Cross-version espeak-ng drift is a further unknown I can't measure from one version.

Committing the bytes also removes espeak-ng — and root to `apt-get install` it — from the requirements for anyone re-running this.

**On cost:** 92 KB against a 13 MB `.git`, about 0.7%, written once and never modified. The repo already tracks 9 other binaries, including a PDF under `docs/assets/`. On the LFS alternative: there's no `.gitattributes` and no LFS anywhere in this repo, so adopting it here would impose a clone-time dependency on every contributor to save 92 KB — heavier than the problem.

Worth noting the automated reviews disagree with each other here. An earlier round on this same PR called the committed WAVs *"the right call for reproducibility."* That's a fair signal this is a judgement call rather than a defect, which is why I've written the reasoning down instead of just re-litigating it per round.

The README also covers running the probe, that its output is safe to paste publicly (no key material), how to regenerate deliberately if the fixtures ever need to change, and that the whole directory is deletable once D1's result is recorded in the doc — the doc being the durable artifact, not the spike.

Everything else in the review matches my read, including the §7 D4 assessment.

---

Issue #2335, opened 2026-08-02
Define correct alignment &amp; scoring behaviour


OPS-PIvers:
## Question

What is the *correct* behaviour of `PhonemeAlignmentEngine` — as opposed to the behaviour the reference implementation happens to have?

Four review notes were recorded against the reference implementation during transcription (source spec §4). **Three of the four are arguably behaviour changes, not bugs**, which is why this is a decision ticket and not an implementation task. They are ticketed together because they interact: fixing insertions changes what PER should divide by, which changes what the thresholds mean.

### The four notes

1. **Insertions are counted but never appear in `alignment[]`.** The final `else` branch increments `insertions` and decrements `j` without pushing an entry. A student who inserts an extra sound sees no badge for it — so the results UI *cannot render insertions at all*, which is why the teacher-results UI is still fog on the map.
2. **PER is normalized by `n` only.** `PER = (subs + omissions + insertions) / N`. With enough insertions `totalEdits` can exceed `n`, making `per > 1`. `Math.max(0, …)` clamps the score at 0, so this is not a crash — but PER is then no longer a rate in [0, 1], which matters if it is ever surfaced to teachers.
3. **`THRESHOLDS[matchLevel] || 80` silently falls back to Close** on a typo'd `matchLevel`. The only unambiguous one: prefer an explicit validation error.
4. **Backtrace tie-breaking is order-dependent.** The `else if` chain prefers substitution over deletion when both are optimal. Defensible but arbitrary — and it changes *which diagnostic the student sees*, so it should be a documented decision rather than an accident of statement order.

### Method

Per source doc §9.1: **write these as tests before deciding how to fix them.** Each note is a behavioural claim with a concrete input, so each converts directly into a test case. Deciding the assertion *is* the decision.

The canonical first test is the spec's own worked example, `El perro` (source spec §6.2) — it is exactly the tap/trill contrast the bias probe exercises.

### Also settle here

The `diagnostic` string in the spec's example output — `"Expected trilled /r/, detected English retroflex /ɹ/"` — is richer than what the reference implementation generates (`"Expected /r/, detected /ɹ/"`). Human-readable phoneme names are an additional lookup table that does not exist yet. Decide whether it is in scope for the spec, and if so where it lives and how it interacts with the three in-scope languages.

## Blocked by

Nothing — this is on the frontier. It is fully independent of D1: the alignment engine is a pure function with zero dependencies, identical whether phonemes arrive from a CTC model or an API.

[...]

Issue #2333, opened 2026-08-02
D1 — Where does acoustic inference run?


OPS-PIvers:
## Question

Does audio → phoneme inference run **client-side (CTC model)**, **server-side (Gemini)**, or **hybrid**?

This is the map's structural decision. It un-fogs student recording UX, concurrency (D7), voice-data privacy (D2), and the Firestore response shape.

**This ticket is now unblocked.** Both acoustic spikes are resolved; the evidence is summarised below.

### Why it is contested

A pronunciation scorer must report what the student **actually said**. An LLM that is told the target text knows what the word is *supposed* to sound like and may normalize toward it — silently erasing the exact errors the feature exists to catch. The failure mode is invisible in casual testing: a biased scorer tells nearly every student "great job."

The codebase's instinct and the spec point opposite ways. `transcribeVideoWithGemini` (`functions/src/aiGeneration.ts:1762`) is already the exact shape of "gated, quota-limited, server-side Gemini audio feature" — `onCall`, 1GiB/300s, gated on `global_permissions/video-activity-audio-transcription`, daily limit in the permission doc's `config` blob, disabled by default. That is the path of least resistance. The spec instead calls for a client-side CTC phoneme model, chosen *precisely because* it is a dumb listener: frame-by-frame acoustics, no notion of the expected answer.

### What the spikes established

All on `gemini-3.5-flash-lite`, all-human fixtures, one speaker. Tickets: [bias probe](https://github.com/OPS-PIvers/SpartBoard/issues/2332), [error direction](https://github.com/OPS-PIvers/SpartBoard/issues/2344).

| Question | Result |
| --- | --- |
| Does the stated target move the report? | **No** — clean both directions |
| Does the report follow the audio? | **Yes** — trill clip 24/30 trill, tap clip 0/30 |
| Does it invent the trill on untrilled audio? | Spanish tap: **never (0/20)**. Anglo `ɹ`: **3/20 (15%)** |
| Does it report the English retroflex `ɹ`? | **Never (0/40)** under any Spanish-framed prompt |

**The original argument for client-side does not survive contact with evidence.** "An LLM told the target will normalize toward it" is not observed. Two earlier alarms were both withdrawn — one was a deprecated model, one was a broken espeak fixture.

### But the fourth row is a new argument, and it is not about bias

[...]

Issue #2332, opened 2026-08-02
Run the D1 bias probe and record the verdict


OPS-PIvers:
## Question

Does Gemini's audio phoneme reporting bias toward the target text it is told to expect?

This is a **task**, not a decision — it produces the fact that D1 waits on. Nothing to design; run the harness and read the output.

The probe is built, syntax-verified, and committed at `scripts/spikes/pronunciation-bias-probe/`. It has never been run: the container had no `GEMINI_API_KEY`. **It does now** (verified present, 39 chars), which is what puts this ticket on the frontier.

```
cd scripts/spikes/pronunciation-bias-probe
node bias-probe.mjs
```

Run from inside the repo after `pnpm run install:all` — the script resolves `@google/genai` from the repo's own `node_modules`.

### The design, and why it needs no phonetician

Audio is held **constant**; only the **claimed target** varies. Ground truth is irrelevant — if reported phonemes move when only the prompt moves, that shift *is* the bias.

|   | Audio | Told target | Honest answer |
|---|---|---|---|
| A | trill (`perro`) | "perro" | `r` |
| **B** | **tap (`pero`)** | **"perro"** | `ɾ` — reporting `r` proves bias |
| C | tap (`pero`) | "pero" | `ɾ` |
| D | tap (`pero`) | *nothing* | `ɾ` |

**Verdict rule:** compare **B against D** — identical bytes, one primed to expect a trill, one unprimed. If B reports `/r/` where D reports `/ɾ/`, the model is answering from the prompt rather than the waveform, and the server-side path in D1 is disqualified.

Each condition runs 10× by default. Do not reduce this — a 2/5 vs 4/5 split clears the threshold at Fisher p ≈ 0.5, i.e. pure noise. The script warns before spending any API calls so an under-powered run can be aborted.

### Load-bearing caveat — record this with the verdict

The clips are espeak-ng synthesis, not real learner speech, and the tap/trill contrast in them is unambiguously clean. That makes this a **conservative** test: a model that fails here certainly fails on messy classroom audio, but **a pass is necessary and not sufficient**. A passing verdict must be recorded as provisional, with re-testing on real learner recordings — plus German final-devoicing and Mandarin tone contrasts — noted as a precondition before the server-side path is actually trusted.

### Resolution should record

- The raw B-vs-D counts and the script's verdict.
- Whether the verdict is clean, or noisy enough to need a re-run.
- If it passes: an explicit note that this is provisional per the caveat above.

[...]

Issue #2331, opened 2026-08-02
Wayfinder Map: Multilingual Pronunciation Engine


OPS-PIvers:
## Destination

An **accepted, implementation-ready specification** for spoken-response (`'Speak'`) quiz items in SpartBoard, covering **Spanish, German, and English** — complete enough that a building agent can execute it without further decisions.

The spec must cover four surfaces: the engine/scoring layer, the type + permission + data surface, the student recording UX, and the teacher authoring & results UX. **No feature code ships from this map** — the deliverable is the spec.

## Notes

**Domain.** SpartBoard quiz widget. React 19 + TypeScript + Vite, flat file structure (no `src/`), `@/` aliases the repo root. Read `CLAUDE.md` before any session — especially the container-query scaling rules and the WCAG AA baseline.

**Source documents** (all on `dev-paul`):
- [`docs/multilingual-pronunciation-engine.md`](https://github.com/OPS-PIvers/SpartBoard/blob/dev-paul/docs/multilingual-pronunciation-engine.md) — the exploration map this wayfinder map supersedes. Decisions D1–D7, codebase grounding (§2), espeak-ng findings (§5).
- [`docs/multilingual-pronunciation-engine-spec.md`](https://github.com/OPS-PIvers/SpartBoard/blob/dev-paul/docs/multilingual-pronunciation-engine-spec.md) — the original spec as authored. **Source material, not accepted design**; contradictions flagged inline.
- [`scripts/spikes/pronunciation-bias-probe/RESULTS.md`](https://github.com/OPS-PIvers/SpartBoard/blob/dev-paul/scripts/spikes/pronunciation-bias-probe/RESULTS.md) — the acoustic evidence behind the closed spikes, and the human clips later tickets reuse as stimuli.
- [`scripts/spikes/english-g2p-probe/RESULTS.md`](https://github.com/OPS-PIvers/SpartBoard/blob/dev-paul/scripts/spikes/english-g2p-probe/RESULTS.md) — the espeak-vs-CMUDict measurement behind the English G2P decision, with a runnable harness. Merged in [PR #2357](https://github.com/OPS-PIvers/SpartBoard/pull/2357).
[...]

PR #2343, opened 2026-08-02
docs(pronunciation): wayfinder map artifacts + D1 probe record and fixes

OPS-PIvers:
**This branch was the collection point for the [Multilingual Pronunciation Engine wayfinder map](https://github.com/OPS-PIvers/SpartBoard/issues/2331).** Parallel sessions working the map's tickets pushed here, so this PR accumulated 18 commits from several sessions. No application code — the only executable touched is the throwaway spike under `scripts/spikes/`.

> **Description corrected after merge.** An earlier version of this body said *"The bias question itself remains open, and #2344 exists to settle it at adequate power. D1 is still blocked."* **All three were false by the time this merged.** The bias question resolved, [#2344](https://github.com/OPS-PIvers/SpartBoard/issues/2344) is closed, and [D1](https://github.com/OPS-PIvers/SpartBoard/issues/2333) is unblocked. The body is rewritten below; nothing in the diff changed.

## What this settled

Two map tickets resolved, both recorded in [`scripts/spikes/pronunciation-bias-probe/RESULTS.md`](https://github.com/OPS-PIvers/SpartBoard/blob/dev-paul/scripts/spikes/pronunciation-bias-probe/RESULTS.md):

| Question | Answer |
| --- | --- |
| Does the stated target bias the reported phonemes? | **No** — clean in both directions |
| Does the report follow the audio? | **Yes** — trill clip 24/30 trill, tap clip 0/30 |
| Does it invent the trill on untrilled audio? | Spanish tap: **never (0/20)**. Anglo `ɹ`: **3/20 (15%)** |
| Does it report the English retroflex `ɹ`? | **Never (0/40)** under any Spanish-framed prompt |

**[#2332 Run the D1 bias probe](https://github.com/OPS-PIvers/SpartBoard/issues/2332)** — provisional pass on bias.
**[#2344 Does the model catch the untrilled error on real audio?](https://github.com/OPS-PIvers/SpartBoard/issues/2344)** — yes on the tap; the Anglo `ɹ` exposed a 15% false-pass rate.
**[#2333 D1](https://github.com/OPS-PIvers/SpartBoard/issues/2333)** is now unblocked and on the frontier.

## Three findings worth not losing

**1. A stale model default silently invalidated a run.** The probe's committed default was `gemini-2.5-flash`. On it, the probe returned `NO RESULT` — the unprimed control transcribed the Spanish clips as English word-shapes (*bella, bane, rainbow, male*), so the comparison had no measurement in it. That was an artifact of the model, not of the audio or the design. Default moved to `gemini-3.5-flash-lite`.

**2. A synthesized stimulus produced a confident, alarming, false reading.** espeak-ng's Spanish trill drew tap reports 36/40, which read as *"this model cannot hear the tap/trill contrast"* — a capability disqualifier for the server-side path. A human recording of the same word scored 82/90. **The fixture was broken, not the model.** `perro_trill.wav` deleted; human clips committed. espeak's *tap* was faithful (40/40 synthetic vs 20/20 human), so the lesson is "verify against human audio", not "synthesis is always wrong".

**3. The proxy error was not the real error.** The probe used a Spanish tap as its stand-in for "failed to trill" for its entire life, and the model reports that tap honestly every time. The actual L1-English classroom error is the retroflex `ɹ` — and on that the model reports a trill ~15% of the time, passing a student who did not trill. Condition **E** added so a run cannot miss this again.

Separately: **the model never reports `ɹ` at all once the prompt names Spanish** (0/40, vs 4/10 unframed). Pass/fail survives — `ɾ` still mismatches target `r` — but per-phoneme diagnostic feedback does not, and the spec's own worked example (expected `/r/`, detected `/ɹ/`) is not reproducible server-side. Flagged onto [#2335](https://github.com/OPS-PIvers/SpartBoard/issues/2335).

## Commits

| | |
|---|---|
| `1d14268` | docs(pronunciation): point the exploration map at its wayfinder tickets |
| `b35f7f9` | Record the D1 bias probe run |
| `84a6ae1` | Re-run the D1 bias probe on a current model, correct the record |
| `98e1a15` | Replace the synthetic trill fixture with a human recording |
| `2af9e39` | fix(spike): name the real cause when the bias probe returns NO RESULT |
| `432611a` | Close the false-positive cell with human audio; add condition E |
| `dd2c162` | fix(spike): report both causes when a NO RESULT collapse is mixed |
| `8a156be` | fix(spike): label condition E's residual bucket for what it counts |
| `af2348e` | Match the results record to the classifier's actual buckets |
| `6d0f329` | fix(spike): say "no samples" instead of 0/0 when condition E collapses |
| `226dbf0` | docs(spike): name the two bare thresholds; record fixture provenance gap |
| `1f59755` | docs(spike): make the audio-provenance note self-enforcing |
| `074cf90` | fix(spike): derive the pre-spend cost estimate from CONDITIONS.length |
| `8cdf9f8` | docs(spike): record the provenance call as decided, not pending |
| `808630d` | docs(spike): calibrate DISCRIMINATION_DELTA; own the judgement calls |
| `ee5509f` | docs(spike): explain why `dead` hardcodes B and D |
| `67bad4c` | docs(spike): record that the model alias cannot be pinned |
| `d0cef41` | docs(spike): fix stale heading claiming the tap is synthetic |

### Docs (`1d14268`)

`docs/multilingual-pronunciation-engine.md` was charted into tracked decision tickets, so its §10 "Next actions" list is no longer the work queue — a reader working from it standalone would pick up already-ticketed work.

- Header callout marking it **superseded as a work queue**, while keeping it as the *grounding reference* the tickets cite. §2 (codebase anchors), §4 (bias-probe design), §5 (espeak-ng findings) remain current.
- §10 replaced with a decision-to-ticket mapping table, plus what stayed deliberately **unticketed** and what moved **out of scope** (Mandarin).
- §8 corrected: the spike was blocked on a missing `GEMINI_API_KEY`, now present.

## Voice fixtures

Six `.wav` files are committed, five of them **recordings of a real person's voice**, contributed deliberately as experiment stimuli after the synthetic ones proved untrustworthy. They are single Spanish words (`perro`, `pero`) with no personal content. Committing them was an explicit decision, not an oversight: the tap/trill contrast cannot be synthesized reliably, and a spike whose stimulus can't be reproduced can't be audited.

Worth flagging against [D2 (voice privacy)](https://github.com/OPS-PIvers/SpartBoard/issues/2331), which is still an open question on this map — the standard for *student* audio is not the standard applied here.

## Notes

- Base is `dev-paul` rather than `main`, since that is where the source docs live.
- Merged as a **merge commit** rather than squashed, deliberately: this PR is a multi-session collection point whose commit chain is the audit trail for a sequence of self-corrections, several of which carry reasoning that exists nowhere else. The two preceding PRs on this spike (#2329, #2330) were squashed; they were single-session and small.
- One reviewer suggestion was declined — removing the `--model` default — with reasoning [in this comment](https://github.com/OPS-PIvers/SpartBoard/pull/2343#issuecomment-5159850221).
- **CI caveat:** the last run confirmed green was `226dbf0`, five commits before the merged head. Runs on the later commits were still in flight or superseded when the merge landed. Those five are comment and documentation changes to spike code, so the risk is low, but `dev-paul` carries content that did not get a completed validation pass on this PR.


OPS-PIvers:
Both reviews land on the same blocking item. Fixed in `2af9e39`; second point declined with reasoning below.

## Fixed — the `NO RESULT` message (`bias-probe.mjs`)

Verified before changing it: the branch asserted `"every call errored. Fix the API errors above"` unconditionally, and in the one run this probe has had there were **zero** API errors — condition D collapsed entirely on `none`.

What made this clearly worth fixing rather than deferring is that the script already knows the distinction matters. The comment above `line()` says exactly this:

> `"3 excluded"` invites the reader to assume API trouble when the real cause may be the model reporting no rhotic — a different problem with a different fix.

`line()` follows that rule; the verdict block didn't. So this wasn't a new opinion, just the file not obeying its own stated principle.

It now counts `ERR` and `none` separately across the collapsed conditions and prescribes a different remedy for each — a `none` collapse points at the stimulus/prompt, an `ERR` collapse at the wiring.

Exercised against stubbed counts for three shapes, since triggering it live would mean burning API calls on a deliberately broken run:

| Scenario | Output |
|---|---|
| Recorded run (D all `none`) | `0 API errors, 10 with no rhotic reported` → stimulus/prompt remedy |
| Genuine outage (D all `ERR`) | `10 API errors, 0 with no rhotic reported` → "fix the API errors" |
| Both dead, one error | `Conditions B and D …` `1 API error` → singular/plural correct |

`RESULTS.md` marked fixed. The `samples[0]` print limitation is left open and labelled as such — that changes what the script *outputs*, not what it *claims*, so it's a real change rather than a message correction.

## Declined — removing the `--model` default

The suggestion is to drop the default and require `--model` explicitly, since a stale default silently invalidated a run. I'd rather not, for two reasons:

1. **`node bias-probe.mjs` is the documented invocation** in [#2332](https://github.com/OPS-PIvers/SpartBoard/issues/2332), [#2344](https://github.com/OPS-PIvers/SpartBoard/issues/2344), and the spike's own README. Requiring a flag breaks the copy-pasteable command in three places, and the next runner hits an error before they hit the experiment.
2. **The failure wasn't actually silent.** The script prints `model: ${MODEL}` on every run and `RESULTS.md` records it per-run — which is precisely how the stale default *was* caught, in `84a6ae1`. The detection mechanism worked.

The "Keep this current" comment is a real maintenance promise that will drift, though, and that part of the critique stands. Happy to change it if @OPS-PIvers wants — flagging rather than deciding unilaterally, since it's someone else's spike ergonomics.

## Note on CI

The PR body previously said no checks would register, because `pr-validation.yml` has `paths-ignore: ['**/*.md', 'docs/**']`. **That is no longer true** — commits from parallel sessions added `bias-probe.mjs` and `.wav` fixtures, and `paths-ignore` only skips when *every* changed file matches. The full 6-job suite is running now. Body corrected.

---

OPS-PIvers:
The `errs === nones` edge case is real and it's mine. Fixed in `dd2c162`.

It was slightly worse than the tie: keying off whichever count was *larger* meant any mixed collapse named one cause and hid the other. `errs=7, nones=3` printed "Fix the API errors above and re-run" with no mention of the three missing rhotics — so the reader fixes the wiring, re-runs, and the condition collapses again.

That's the same confidently-wrong-cause failure the block was added to stop, just narrower. So rather than special-case the tie, I changed the shape: a single cause is named **only** when the other count is genuinely zero; otherwise both are reported, with the note that fixing one won't restore the condition.

| `errs` | `nones` | Message |
|---|---|---|
| 0 | 10 | stimulus/prompt remedy |
| 10 | 0 | "Fix the API errors above and re-run." |
| **5** | **5** | both causes, both remedies |
| **7** | **3** | both causes, both remedies |

Verified against all four. Rows 3 and 4 were wrong before this commit.

**Not taking** the `> 0.1` false-pass threshold suggestion. Agreed it's unlabelled, but that constant belongs to the condition-E logic from `432611a`, not to anything I've touched — naming it means asserting what its author intended it to mean, and I'd be guessing. Worth a follow-up by whoever wrote it, or a note on [#2344](https://github.com/OPS-PIvers/SpartBoard/issues/2344) if the probe outlives this run.

---

OPS-PIvers:
Took the `countOf('E', 'other')` label point — `8a156be`. Noting why, since I declined a similarly-sized suggestion one comment ago and the two look alike from outside.

The classifier docs in this same file define the bucket as:

> `'other'` another rhotic — **chiefly** the English retroflex ɹ

"Chiefly", not "only". So `E reported the retroflex ɹ honestly` was counting a bucket wider than its label — a uvular ʁ would land there and be reported as the retroflex heard honestly. Now reads `E reported a non-trill rhotic, usually ɹ`, with a comment saying why. Changes no recorded number; it's ɹ in every run so far.

What separates this from the `> 0.1` threshold I passed on: this one needed no guess about intent. The file already documents what the bucket holds, so the label was checkable against the code's own stated contract and simply didn't match. The threshold has no stated rationale anywhere, so naming it would have meant inventing one.

Worth noting this is the third message in this block asserting more than the measurement supported — after the `NO RESULT` cause and the mixed-collapse case. Individually each is small; as a pattern it's the thing to watch in a script whose entire output is claims about what a model heard.

---

OPS-PIvers:
Took the `eValid === 0` point — `6d0f329`.

Agreed on the diagnosis, but I'd put the severity higher than "mildly confusing". `E reported a trill (a WRONG pass): 0/0` doesn't read as absent data to someone scanning output — it reads as **zero wrong passes**, the most reassuring result the probe can produce. On the one check that exists to catch the costliest classroom error (a student told they rolled the r when they didn't), a collapse that looks like a clean bill of health is the wrong failure direction.

Now mirrors the DISCRIMINATION block:

```
No usable samples in E — no false-pass rate can be computed. This is not
a clean bill of health; the costliest error is simply untested here.
```

Your framing made it easy to justify: DISCRIMINATION already had exactly this guard two blocks up, so applying it to E introduces no new convention — it just makes the file consistent with itself.

That's the fourth message in this block that stated more than the measurement supported, and the second one that failed *reassuringly*. Worth naming as a pattern rather than four separate nits: in a script whose whole output is claims about what a model heard, "no data" and "a good result" have to be visually distinguishable, or the spike can talk you into shipping the thing it was built to interrogate.

Rebased over `af2348e` (concurrent push from another session on this shared branch).

---

OPS-PIvers:
Done — `RESULTS.md` now carries an explicit **"If you are here to collect more audio, stop and read this"** section rather than a bare pointer.

You're right that the previous version failed at its own job. It recorded the gap but not its weight, and a note nobody registers isn't a note — it's how "we committed undocumented voice recordings once" quietly becomes the working norm.

The line I drew, since it matters for what the file now claims: **the constraint is stated as binding, but labelled a standing recommendation awaiting @OPS-PIvers' decision, not ratified policy.** The decision to keep these fixtures was the repo owner's and it stands — I'm not reversing it through a docs edit. But "no learner-audio collection before D2 is answered" is a project policy, and writing it into the repo as settled would be me manufacturing authority I don't have. The note says to treat it as binding until told otherwise, on the grounds that the failure is irreversible and waiting costs nothing.

The substantive argument the section makes, which is the part worth reviewing rather than the wording: **every reason this gap was tolerable stops applying the moment the recording is a student's.** One consenting-by-assumption adult is a different object from a minor's voice, which engages COPPA, district agreements, and parental consent — none of which can be arranged after the file is in a public repo. That's why the recommended gate is *before the first recording*, not before ship.

Worth noting three independent reviews converged on this. That's a stronger signal than any one of them, and it's what moved me from "flag it and let the owner decide" to "flag it, state it as binding, and let the owner overrule."

Also carried to [#2344](https://github.com/OPS-PIvers/SpartBoard/issues/2344#issuecomment-5159975941) — which matters, because that ticket already names real learner recordings as a precondition for its re-run. The gate and the ticket point at the same moment.

---

OPS-PIvers:
Confirmed and fixed. Also corrected the stale `all 4 x RUNS calls` in the comment directly above it, which had desynced the same way.

Derived from `CONDITIONS.length` rather than bumped to `5`, so adding a condition F can't reintroduce it. `CONDITIONS` is declared at line 180, well before the use at ~273.

```
--runs 5 -> quotes 25 calls (was 20)
--runs 9 -> quotes 45 calls (was 36)
```

Worth stating why this one isn't a nit despite being a single character of real change: it is the **only** message in the script whose entire job is letting the operator abort *before* spending money, and it was under-quoting by 20%. A wrong number there is worse than no number — someone reads "20 calls", accepts, and gets billed for 25.

That is now the **fifth** finding in this file of the same shape: output asserting something the code doesn't support. The others were the `NO RESULT` cause, the mixed-collapse case, the `ɹ` bucket label, and `0/0` reading as a clean false-pass result. Three of the five failed in the reassuring direction — cheaper, safer, or cleaner than reality.

I'd put weight on that pattern rather than on any individual fix. This is a probe whose only product is claims about what a model heard, and its self-reporting has now been wrong five times in one review cycle. That is not a reason to distrust the recorded numbers — those were checked against raw output — but it is a reason to treat [#2344](https://github.com/OPS-PIvers/SpartBoard/issues/2344) as a genuine re-run from scratch rather than a confirmation pass, and to read the D1 verdict as provisional in the way `RESULTS.md` already says it is.

---

OPS-PIvers:
You're right that binding-until-told-otherwise needs a told-otherwise moment. Here it is: **@OPS-PIvers has decided, knowingly, to keep the files.** `RESULTS.md` now records that as a dated decision with an owner rather than pointing at one that hadn't happened. That was a real defect in what I wrote and your framing fixed it.

**But the removal remedy rests on a premise I checked, and it doesn't hold.**

> The five WAV files are permanently in the public repo history the moment this merges.

They already are. `98e1a15` and `432611a` were pushed to this branch on a public repository well before the objection was raised:

```
$ git log --oneline --diff-filter=A -- 'audio/*human*'
432611a Close the false-positive cell with human audio; add condition E
98e1a15 Replace the synthetic trill fixture with a human recording
$ git branch -r --contains 98e1a15
  origin/claude/multilingual-pronunciation-plan-x1jcrv
```

Merging doesn't create the exposure — it propagates it into `dev-paul`. So option (b), moving the audio to a `.gitignore`'d directory, does **not** deliver "zero provenance debt": a later commit deleting the files leaves them fully fetchable from this branch's pushed history. It would have signalled intent without changing who can reach the bytes, while reading in the diff like a fix. Actually purging them meant rewriting a branch that several sessions were pushing to concurrently — and unreachable objects can still be served by SHA for a period afterward.

Given a choice between a remedy that looks like a fix and isn't, and keeping the files with the gap labelled loudly, the owner chose the latter. The reasoning is recorded in `RESULTS.md` so it reads as considered rather than skipped.

**What I did tighten, because your underlying concern is right:** the note now separates two calls that were blurring together. Accepting an already-public adult fixture is cheap. Recording a **minor** without settled consent cannot be walked back at all. The learner-audio gate is stated flatly — no collection before D2 is answered, before the first recording, not before ship — with an instruction that relaxing it must be an explicit decision on [#2344](https://github.com/OPS-PIvers/SpartBoard/issues/2344), so it can't become a step nobody noticed.

Four reviewers converged on this and it materially improved the artifact. Thank you — genuinely. Treating it as closed now.

---

OPS-PIvers:
Both fair. Fixed — and I took the option you offered rather than the fallback.

**On the phrasing.** "Whoever set it did not write down why 0.5" was evasive about a value chosen inside this PR. Gone; it now says plainly that it was a judgement call with no derivation behind it.

**On the calibration** — you wrote *"If you do have a calibration argument for 0.5 … add it — that's what makes the threshold defensible rather than arbitrary."* Agreed, so I computed it instead of shipping the honest-but-empty annotation:

```
two-tailed Fisher exact, n=10 per condition, splits that exactly clear 0.5
  10/10 vs 5/10   p ~= 0.033
   9/10 vs 4/10   p ~= 0.057
   8/10 vs 3/10   p ~= 0.070   <- weakest qualifying splits are mid-range
   7/10 vs 2/10   p ~= 0.070
   5/10 vs 0/10   p ~= 0.033
```

A bare `DISCRIMINATION_DELTA` pass therefore lands around **p ≈ 0.03–0.07** — meaningfully stronger than a bare `BIAS_DELTA` pass (p ≈ 0.14), still short of decisive. "Probably real, worth confirming."

Two things I checked before trusting my own numbers:

1. **Validated the implementation against `BIAS_DELTA`'s published figures first.** It reproduces 9/10 vs 5/10 as `p = 0.1409` (documented "~0.14") and 10/10 vs 2/10 as `p = 0.00071` (documented "< 0.01"). Same code then produced the table above.
2. **Nothing on record turns on the cutoff.** The recorded run cleared it at 24/30 vs 0/30 — gap 0.80, `p ≈ 3e-11`. Also noted in the comment, so a future reader doesn't relitigate 0.5 thinking the verdict hangs on it.

**`FALSE_PASS_WARN_RATE` deliberately does not get the same treatment**, and now says why rather than leaving it looking like an oversight: it gates a single rate, not a comparison between conditions, so there is no two-sample test to calibrate against. What counts as an intolerable rate of passing a student who did not roll the r is a **pedagogical** judgement, not a statistical one. 0.1 is a placeholder for a teacher's answer, not a substitute for it — and inventing a p-value there would have been exactly the false rigour your comment is pushing against.

---

OPS-PIvers:
Good catch on the `['B','D']` / `CONDITIONS` contrast — read side by side, the hardcoded one does look like the one that got missed. Comment added explaining why it isn't: the pre-spend estimate counts every condition because every condition costs API calls, whereas this list is the two conditions the **bias verdict** is computed from, which is a property of the experiment's design rather than of the `CONDITIONS` array. A, C and E can each collapse without making the bias verdict uncomputable — they have their own guards — so widening it would abort runs that still had an answer to give.

**On the WAV files: the decision stands, and it was taken with the purge option explicitly on the table** — force-push was offered as one of four choices, alongside documenting provenance, deleting from HEAD, and keeping as-is. @OPS-PIvers chose to keep.

**One premise in the recommendation is wrong, though, and it's the load-bearing one:**

> The bytes are not load-bearing for anyone re-running the probe (they'll generate fresh API calls anyway)

Fresh API calls still need audio to send. These clips **cannot be regenerated** — per `README.md` they were cut from a `recording.MOV` that is not in the repository:

```
ffmpeg -i recording.MOV -map 0:1 -ac 1 -ar 22050 -c:a pcm_s16le full.wav
ffmpeg -i full.wav -ss <start> -t <dur> -c:a pcm_s16le audio/perro_trill_human_N.wav
```

That's the difference between these and the espeak fixtures they replaced: a synthetic clip is a command line, a human recording is not. Dropping the bytes would mean the recorded results can never be reproduced against the same stimulus — and it would also destroy the evidence for the one claim keeping `perro_tap.wav` credible (40/40 synthetic vs 20/20 human agreement). The md5s in `RESULTS.md` identify the files; they don't substitute for them.

So the trade isn't "documentation vs. bytes" — it's provenance debt against permanent loss of reproducibility for the only human-audio evidence this spike has.

Your point about **merge extending this into `dev-paul`'s mainline history** is the strongest version of the objection anyone has raised, and I've put it to @OPS-PIvers rather than deciding it here. Everything else — the labelling, the D2 gate — is unchanged and already says loudly that this precedent isn't the project's answer.

---

OPS-PIvers:
Right concern, but I checked the remedy against the live API before documenting it and **it doesn't exist** — which makes the problem worse than described, not better.

```
models.list matching 3.5-flash-lite:
  models/gemini-3.5-flash-lite          <- the only identifier offered

gemini-3.5-flash-lite-07-2026  -> REJECTED: 404 not found for API version v1beta
gemini-3.5-flash-lite          -> ACCEPTED (modelVersion = "gemini-3.5-flash-lite")
```

Three things follow. No dated variant is exposed. Requesting one 404s. And the response's own `modelVersion` echoes the bare alias, so a run **cannot capture which checkpoint answered it** even retrospectively — the `3.5-flash-lite-07-2026` string in `RESULTS.md` is not an identifier the API will accept or currently report back.

So `--model gemini-3.5-flash-lite-07-2026` would have failed on first use for whoever followed the advice. Thanks for raising it — checking it turned a footnote into a real constraint.

**Why this is sharper than a reproducibility nit:** this exact failure already cost the spike one full run. The first execution used a stale `gemini-2.5-flash` default and returned a `NO RESULT` that was an artifact of the model rather than the audio, which is why the script now warns that a stale default "silently invalidates the whole experiment." That warning assumed the fix was keeping the default current. It isn't sufficient — a *correct, current* alias still drifts underneath you, and there's no way to hold it still.

Documented in `README.md` under *Running it*, with the consequence recorded on [#2344](https://github.com/OPS-PIvers/SpartBoard/issues/2344#issuecomment-5160101600): re-run every condition in one sitting, never compare fresh cells against the numbers recorded here, and don't read a moved rate as a finding about the audio — "the model changed" is at least as likely and isn't distinguishable.

---

OPS-PIvers:
Taking one of these, declining the other.

**Fixed — the stale heading.** You're right, and it contradicted its own section. The script loads `pero_tap_human.wav` as `TAP` (`bias-probe.mjs:166`); `perro_tap.wav` is retired and kept only as evidence. Now reads *"All active fixtures are human recordings; the synthetic tap is retired"*, per your suggestion. This one mattered more than a typo — the README is the re-run guide for [#2344](https://github.com/OPS-PIvers/SpartBoard/issues/2344), and someone scanning headings would have concluded the tap condition was synthetic and drawn the wrong inference from a moved result.

**Not changing the `RESULTS.md` model line.** @OPS-PIvers has confirmed the model directly, and it is not in question — the run was `gemini-3.5-flash-lite`. That line stays as-is.

To close the loop on what I actually verified, since it's the narrower claim: the README's sentence is about what the API *reports back*, not about which model ran. Dumping the full response payload:

```
top-level keys: sdkHttpResponse, candidates, modelVersion, responseId, usageMetadata
modelVersion = "gemini-3.5-flash-lite"
dated substrings anywhere in payload: NONE
```

So `modelVersion` does carry an identifier — the bare alias — and no dated string appears anywhere in the response. The README sentence is accurate as written: you cannot use what the API returns to request the same checkpoint later. That is a statement about *pinning*, not about model identity, and both it and the `RESULTS.md` line can be true at once.

Your closing point stands either way, and it's the one that matters: runs are not comparable across time, and date-stamping is the mitigation. That's already recorded on #2344 as a constraint on how the re-run must be executed.

---

Issue #2356, opened 2026-08-03
Validate the G2P accuracy numbers against real teacher vocabulary


OPS-PIvers:
## Question

Nothing to decide — this closes the weakest link in [Does English G2P need CMUDict + a neural fallback, or do rules suffice?](https://github.com/OPS-PIvers/SpartBoard/issues/2336).

### Why it exists

That ticket resolved on a first-party measurement: espeak-ng scores **94.1% segments / 97.1% under vowel-reduction tolerance** on a K-12 EL practice corpus, which is what makes "rules are enough, drop the neural fallback" defensible.

**That corpus was constructed by the resolving agent** — 341 words drawn from standard beginner-ESL topical domains — **not sampled from real SpartBoard authoring data.** The same measurement on full CMUDict scores 61.5%, because that corpus is surname-heavy. The gap between 94% and 61% is entirely a claim about *what teachers actually type*, and that claim is currently an assumption.

Per the map's Notes rule 4: a property nobody measured is an assumption, however plausible. This map has already been burned twice by exactly this — a stale model default and a synthetic audio fixture — and both times the artifact looked fine until it was checked.

### What to do

1. Assemble **200–500 real or realistically-drafted Speak target phrases** in es/de/en. Sources, in preference order: existing SpartBoard quiz content with language-practice vocabulary; a world-languages or EL teacher drafting what they would actually assign; published beginner curriculum word lists for the district's courses.
2. Run them through the committed harness — `scripts/spikes/english-g2p-probe/teacher.py`, which already implements the CMUDict cross-check and the flag-rate calculation.
3. Report: segments / +reduction accuracy, the CMUDict-disagreement flag rate, and a breakdown of the residual by the failure taxonomy in `RESULTS.md`.

### What would change the verdict

- **Proper nouns or loanwords are a large share of real input.** Both are the categories where rules degrade *and* where CMUDict does not rescue them (*piñata*, *jalapeño*, *quinoa*, *edamame* are all absent from CMUDict). This would not resurrect the neural fallback — published neural G2P fails on the same class — but it would raise the flag rate and make the teacher-facing affordance load-bearing rather than a nicety.
- **Flag rate lands far above ~6%.** At that point the review queue stops being a safety net and starts being data entry, and the authoring UX in [#2341](https://github.com/OPS-PIvers/SpartBoard/issues/2341) needs rethinking.
[...]

Issue #2339, opened 2026-08-02
Spanish and German G2P rule tables — design and source


OPS-PIvers:
## Question

What shape do the Spanish and German G2P rule tables take, and where do the rules come from?

> **Updated by [D1](https://github.com/OPS-PIvers/SpartBoard/issues/2333), [#2349](https://github.com/OPS-PIvers/SpartBoard/issues/2349) and [#2336](https://github.com/OPS-PIvers/SpartBoard/issues/2336).** Three sub-questions below are now answered: the tables run **server-side only**, invoked at authoring time to precompute the expected phoneme string onto the `QuizQuestion` doc; the output alphabet is **fixed to espeak's 392-symbol set** by the acoustic model choice; and **English joins this ticket** rather than needing its own subsystem. Two new requirements arrived: **rule-table versioning** and a **CMUDict cross-check**.

### What the rules must cover

Verified working in espeak-ng 1.51 (source doc §5), so these are known-tractable, not speculative:

**Spanish** — digraphs (`rr`, `ch`, `ll`, `qu`), silent `h`, soft/hard `c` and `g`, and the tap /ɾ/ vs trilled /r/ contrast that the whole feature's canonical example rests on (`pero` → `pˈeɾo` vs `perro` → `pˈero`). Plus the dialect split: `cielo` → `θjˈelo` (es) vs `sjˈelo` (es-419).

**German** — umlauts (`ä`, `ö`, `ü`), diphthongs (`ei`, `eu`, `au`), digraphs and trigraphs (`ch`, `sch`, `sp`, `st`), and word-final consonant devoicing (`Tag` → `tˈɑːk`). The ach/ich split (`Bach` → `bˈax` vs `ich` → `ˈɪç`) is what the "Exact" strictness tier needs to be meaningful at all.

[...]

PR #2357, opened 2026-08-03
spike(g2p): English G2P probe harness and results

### LINKED ISSUE #2336: Does English G2P need CMUDict + a neural fallback, or do rules suffice?
## Question

The source spec budgets an entire subsystem for English grapheme-to-phoneme — dictionary lookup (CMUDict mapped to IPA) **plus** a fallback neural G2P parser — to resolve non-phonemic spelling. Is that subsystem actually necessary?

### Why it is in doubt

Source doc §5 tested espeak-ng 1.51 against the spec's own two cited examples and got both right, rule-based, with **no dictionary and no neural fallback**:

- `rough` → `ɹˈʌf`
- `through` → `θɹˈuː`

Those are precisely the pair the spec offers as proof that rules are insufficient. If a rule engine handles them, the spec is budgeting for a subsystem that may not be needed — and English is only in scope for this map at all because dropping that subsystem makes it cheap.

### What to establish

- How far a rule-based approach actually gets on English, beyond a two-example anecdote. Where does it break — and are those breakages ones a classroom EL use case will hit?
- Whether the relevant corpus for this feature (short, teacher-authored target phrases, not arbitrary text) changes the answer. The spec assumes general-purpose G2P; the actual input is narrow.
- What CMUDict costs if it *is* needed: size, license, IPA mapping fidelity, and whether it ships client-side or server-side — which interacts with D1.
- Whether any viable rule source is license-clean, given the GPL-3 finding on espeak-ng (see the licensing ticket). A rules-suffice answer is worth much less if the only rules available are copyleft.

### Note

Findings should be captured on a throwaway `research/english-g2p` branch with a pointer back to this ticket, not pasted into the issue body.

## Blocked by

Nothing — this is on the frontier, and it is AFK: resolvable by a `/research` subagent without human input.

OPS-PIvers:
Adds `scripts/spikes/english-g2p-probe/` — the measurement harness and findings that resolved [#2336](https://github.com/OPS-PIvers/SpartBoard/issues/2336) on the [Multilingual Pronunciation Engine map](https://github.com/OPS-PIvers/SpartBoard/issues/2331).

**No application code.** Nine standalone Python scripts and a `RESULTS.md`, landing beside the sibling `scripts/spikes/pronunciation-bias-probe/` on `dev-paul`. Nothing imports from it and nothing in the app imports it.

## What it measures

espeak-ng 1.51 (`en-us`) against CMUDict across all 117,493 alphabetic headwords plus frequency-banded and practice-vocabulary subsets, to answer whether English grapheme-to-phoneme needs a dictionary plus a neural fallback or whether rules suffice.

| Corpus | n | literal | segments | +reduction |
| --- | --- | --- | --- | --- |
| All CMUDict (surname-heavy) | 117,493 | 56.8% | 61.5% | 68.6% |
| Top 1,000 frequency | 1,000 | 88.5% | 93.1% | 97.1% |
| K-12 EL practice corpus | 340 | — | 94.1% | 97.1% |

**Verdict: rules + a teacher-facing confidence affordance.** Neural fallback dropped — published state-of-the-art runs 19.9–25.1% WER on held-out CMUDict, failing on exactly the loanwords and proper nouns a dictionary also misses, which is a silent 1-in-5 error in a flow where a human is already present. CMUDict stays as a cross-check oracle at a 5.9% flag rate.

The harness validates against published work: the assumption-free score on full CMUDict (56.8%) reproduces Black, Lenzo & Pagel (1998) at 57.80% to within one point.

It also corrected the ticket's own premise — espeak resolves `through` from its bundled exception dictionary rather than from rules, so espeak English is already the dictionary-plus-rules architecture the spec proposed.

## Files

- `RESULTS.md` — findings, failure taxonomy, license survey, reproduction steps, and a documented list of the research's own limits
- `prep.py`, `run_espeak.py` — build the corpora
- `score.py`–`score4.py`, `split.py`, `src.py` — scoring passes. `score2.py` holds the shared ARPAbet/IPA helpers (`esp_phones`, `cmu_phones`, `VOW`, `ed`) which the others **import**; its own analysis sits behind an `if __name__ == "__main__":` guard so importing it needs no data files. `score.py` and `score3.py` are earlier iterations kept for provenance.
- `teacher.py` — practice-vocabulary corpus and CMUDict cross-check flag rate
- `.gitignore` — excludes the ~19 MB of downloaded and generated data; `RESULTS.md` documents how to regenerate

## Verification

`prettier --check` passes on the added files. No TypeScript, ESLint, or test surface is touched — the directory is Python-only and outside the build.

Beyond CI, which cannot exercise this code:

- Every script run against a 20-word stub fixture — no name, import, or syntax errors.
- `from score2 import …` in a directory with no `cmu.json`/`espeak.json`, confirming the `__main__` guard does what it claims.
- `run_espeak.py`'s failure guard exercised against both a healthy espeak-ng and a deliberately broken invocation.
- An empty data directory produces a diagnostic from all six evaluators rather than a traceback.
- The full pipeline re-run end-to-end against freshly downloaded CMUDict and frequency data, reproducing the headline numbers independently of the session that first produced them — and re-verified again after the rebase onto `dev-paul`.

## Review history

Eight review rounds; every item applied or declined with reasoning in the thread. Three defects surfaced that no round flagged, each found while verifying a suggestion rather than applying it:

1. A `random.sample` `ValueError` that the suggested zero-division guard would have left in place.
2. `teacher.py` incrementing its denominator before the `None` check, silently deflating every rate under partial data.
3. `src.py` classifying empty espeak output as `'other'`, skewing the dictionary-vs-rules split.

Two suggested fixes targeted the wrong mechanism and were replaced with measured alternatives: guarding on `returncode` when espeak-ng exits 0 on failure, and batching `-X` output that isn't safely batchable.

## Note

**The weakest claim is flagged, not buried.** The 341-word practice corpus was constructed by the resolving agent rather than sampled from real SpartBoard content, and the gap between 94% and full-CMUDict's 61% rests entirely on that. Tracked as [#2356](https://github.com/OPS-PIvers/SpartBoard/issues/2356), and called out in `RESULTS.md` under "Known limits."

https://claude.ai/code/session_01TMMRk7U9iz4VpseMHqCHfW

### DIFF
--- a/scripts/spikes/english-g2p-probe/.gitignore
+++ b/scripts/spikes/english-g2p-probe/.gitignore
@@ -1,0 +1,8 @@
+# Generated / downloaded data — regenerate with the steps in RESULTS.md (~19 MB)
+cmudict.dict
+g10k.txt
+words.txt
+*.json
+*.pdf
+*.txt
+!RESULTS.md
--- a/scripts/spikes/english-g2p-probe/.gitignore
+++ b/scripts/spikes/english-g2p-probe/.gitignore
@@ -5,4 +5,3 @@ words.txt
 *.json
 *.pdf
 *.txt
-!RESULTS.md
--- a/scripts/spikes/english-g2p-probe/.gitignore
+++ b/scripts/spikes/english-g2p-probe/.gitignore
@@ -1,7 +1,9 @@
-# Generated / downloaded data — regenerate with the steps in RESULTS.md (~19 MB)
+# Pipeline data — downloaded or generated. Regenerate with the steps in RESULTS.md (~19 MB).
 cmudict.dict
-g10k.txt
 words.txt
+g10k.txt
 *.json
-*.pdf
-*.txt
+
+# Ad-hoc reference material pulled while researching (papers, wordlists) is deliberately
+# NOT ignored — it shows up as untracked. Don't commit it, but better visible than silently
+# swallowed, which is what a blanket *.txt / *.pdf rule did here before.
--- a/scripts/spikes/english-g2p-probe/RESULTS.md
+++ b/scripts/spikes/english-g2p-probe/RESULTS.md
@@ -246,9 +246,11 @@ python3 split.py        # accuracy broken down by resolution path
 python3 teacher.py      # K-12 practice corpus + CMUDict cross-check flag rate
 ```
 
-`score2.py` holds the shared ARPAbet/IPA helpers (`esp_phones`, `cmu_phones`, `ed`, `VOW`);
-the other scorers `exec` its header. `score.py` and `score3.py` are earlier iterations kept
-for provenance.
+`score2.py` holds the shared ARPAbet/IPA helpers (`esp_phones`, `cmu_phones`, `ed`, `VOW`)
+and the other scorers import them directly; its own analysis is behind an
+`if __name__ == "__main__":` guard, so importing it does not require `cmu.json` or
+`espeak.json` to exist. `score.py` and `score3.py` are earlier iterations kept for
+provenance.
 
 ## Sources
 
--- a/scripts/spikes/english-g2p-probe/RESULTS.md
+++ b/scripts/spikes/english-g2p-probe/RESULTS.md
@@ -68,6 +68,20 @@ years apart, agreeing to within a point is good evidence the mapping and scoring
 Spot-checks confirm it: _island, choir, yacht, colonel, one, said, friend, subtle, women,
 busy, laughter, daughter, through, though, thought_ all come out correct.
 
+### Independent reproduction
+
+The whole pipeline was later re-run from scratch — fresh `cmudict.dict` and
+`google-10000-english` downloads, corpora rebuilt, espeak re-run over all 117k words — in a
+different session from the one that produced the table above. Every rate reproduced
+identically, including the practice corpus (94.1% / 97.1% / 91.5%, 340/341 CMUDict coverage
+with _piñata_ the sole miss) and the rules-vs-dictionary split (77.1%/22.9% at top-1000).
+
+Two cells drifted, from re-downloading upstream data rather than from the harness: the
+top-10k band matched **9,425** words against CMUDict rather than 9,428 (and its stress
+figure 87.7% rather than 87.8%), and the complement band 108,068 rather than 108,076. All
+accuracy rates were unchanged to 0.1pp. Worth knowing that these corpora are not pinned —
+anyone re-running should expect single-digit membership drift and record the download date.
+
 ### Rules are not the weak part
 
 Split by espeak's own resolution path:
--- a/scripts/spikes/english-g2p-probe/RESULTS.md
+++ b/scripts/spikes/english-g2p-probe/RESULTS.md
@@ -1,0 +1,267 @@
+# English G2P probe — does English need CMUDict + a neural fallback?
+
+Resolves [#2336](https://github.com/OPS-PIvers/SpartBoard/issues/2336) on the
+[Multilingual Pronunciation Engine map](https://github.com/OPS-PIvers/SpartBoard/issues/2331).
+
+**Verdict: rules + a teacher-facing confidence affordance. Drop the neural fallback.
+Keep CMUDict as a cross-check oracle, not as the primary path.**
+
+Measured with **espeak-ng 1.51** (`en-us`, Ubuntu 24.04 package) against **CMUDict**
+(`cmusphinx/cmudict` master, BSD-2-Clause), over all 117,493 alphabetic headwords plus
+frequency-banded and hand-built teacher-corpus subsets. Current espeak-ng upstream is
+1.52.0; numbers may be marginally better there.
+
+---
+
+## First: a correction to the ticket's premise
+
+The ticket claimed espeak-ng got both of the spec's cited examples right "with no dictionary
+and no neural fallback." That is half wrong. Tracing with `espeak-ng -v en-us -q -X`:
+
+| Word                | Resolved by                                                                                                 |
+| ------------------- | ----------------------------------------------------------------------------------------------------------- |
+| `rough` → `ɹˈʌf`    | **Letter-to-sound rule** (`r) ough → [Vf]`, priority 99). Genuine rule win.                                 |
+| `through` → `θɹˈuː` | **`Found: 'through' [Tru:]`** — an entry in espeak's bundled exception dictionary (`en_list`, 5,794 lines). |
+
+So espeak-ng English is **not a pure rule engine**. It is already the architecture the spec
+proposed — exception dictionary plus fallback rules — in a 166 KB compiled `en_dict`.
+
+Share of lookups resolved by the exception dictionary, by frequency band:
+
+| Band              | Dictionary | Rules |
+| ----------------- | ---------- | ----- |
+| Top 1,000         | 22.9%      | 77.1% |
+| Rank 1,000–3,000  | 12.3%      | 87.7% |
+| Rank 5,000–10,000 | 12.1%      | 87.9% |
+
+The real question is therefore not _"rules vs. dictionary"_ but **"is a 166 KB curated
+exception list enough, or do you need a 3.6 MB dictionary plus a neural net on top?"**
+
+---
+
+## Accuracy
+
+| Corpus                                 | n       | literal ARPAbet | segments¹ | + vowel reduction² | stress pattern | PER (norm.) |
+| -------------------------------------- | ------- | --------------- | --------- | ------------------ | -------------- | ----------- |
+| All CMUDict headwords (surname-heavy)  | 117,493 | 56.8%           | 61.5%     | 68.6%              | 72.1%          | 9.7%        |
+| Top 1,000 frequency                    | 1,000   | 88.5%           | 93.1%     | 97.1%              | 93.5%          | 1.9%        |
+| Top 3,000 frequency                    | 3,000   | 83.6%           | 88.2%     | 95.3%              | 91.1%          | 2.7%        |
+| Top 5,000 frequency                    | 5,000   | 81.2%           | 86.3%     | 94.4%              | 89.7%          | 3.0%        |
+| Top ~10,000 frequency                  | 9,428   | 78.3%           | 83.8%     | 92.7%              | 87.8%          | 3.5%        |
+| CMUDict minus top-10k                  | 108,076 | 54.9%           | 59.6%     | 66.5%              | 70.7%          | 10.2%       |
+| **Hand-built K-12 EL practice corpus** | 340     | —               | **94.1%** | **97.1%**          | 91.5%          | —           |
+
+¹ Phone-sequence match, stress ignored, after normalising five _systematic notation_
+differences that are not pronunciation errors: velar-nasal assimilation (`IH NG K` vs
+`IH N K` in _including_ — espeak is phonetically correct), T/D flapping, the cot–caught
+merger, the north–force merger (`F OW R`/`F AO R` for _four_), and `ER`+`R` vs `ER`.
+
+² Additionally treating AH/IH/IY/UH/ER as one reduced class — i.e. not penalising
+unstressed-vowel-quality disagreement, which is not a contrast a learner can be scored on.
+
+### Harness validation
+
+The assumption-free **literal** column on full CMUDict (56.8%) lands within one point of
+Black, Lenzo & Pagel (1998), who report **57.80% words correct / 91.99% letters correct**
+for CART letter-to-sound rules on the same dictionary. Two independent rule systems, 26
+years apart, agreeing to within a point is good evidence the mapping and scoring are sound.
+Spot-checks confirm it: _island, choir, yacht, colonel, one, said, friend, subtle, women,
+busy, laughter, daughter, through, though, thought_ all come out correct.
+
+### Rules are not the weak part
+
+Split by espeak's own resolution path:
+
+| Resolved by           | segments | + reduction |
+| --------------------- | -------- | ----------- |
+| Letter-to-sound rules | 84.0%    | 93.0%       |
+| Exception dictionary  | 82.9%    | 90.0%       |
+
+Dictionary entries score _lower_ because they are the residue of irregular words. The rules
+are carrying the system competently.
+
+---
+
+## Why the neural fallback is not justified
+
+State-of-the-art neural G2P on **held-out CMUDict** (r-G2P, arXiv:2202.11194 Table 1):
+
+| Model                              | PER   | WER        |
+| ---------------------------------- | ----- | ---------- |
+| CNN-encoder + BiLSTM               | 4.81% | 25.13%     |
+| CNN + NSGD                         | 5.58% | 24.10%     |
+| Encoder-decoder + global attention | 5.04% | 21.69%     |
+| Transformer 4×4                    | 5.23% | 22.10%     |
+| Best r-G2P (adversarial)           | 4.84% | **19.85%** |
+
+A neural fallback exists to handle words a dictionary misses. On the actual corpus those
+misses are ~0.3–2.3% and are dominated by abbreviations and loanwords — precisely the class
+where published neural G2P _also_ fails (LatPhon names "proper nouns & foreign words" in its
+own limitations).
+
+So the fallback would be **wrong roughly one time in five, silently, on exactly the hard
+words.** At authoring time, with a human present, a wrong-but-confident phoneme string is
+strictly worse than a flag reading _"I'm unsure about this word — here's my guess, confirm
+or edit."_ A neural model costs a model artifact, cold-start weight in a Cloud Function, a
+training/eval pipeline, and a re-derivation story — to buy a 20%-error guess in place of a
+zero-error human confirmation. That trade is negative.
+
+> **Caution on comparing numbers.** The 1.9–3.5% PER measured here on realistic frequency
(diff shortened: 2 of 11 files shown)


OPS-PIvers:
Took the `exec`-split note — pushed in `ace2a60`.

You flagged it as "not a blocker for a provenance spike," but it's a bit more than provenance: [#2356](https://github.com/OPS-PIvers/SpartBoard/issues/2356) explicitly directs a future session to re-run `teacher.py` against real teacher vocabulary, to validate the one number this spike's verdict rests on. So a silent break in the shared-helper path would land exactly where it does the most damage.

Applied as suggested: `score2.py`'s analysis is now behind `if __name__ == "__main__":`, and the four dependents import directly —

```python
from score2 import esp_phones, cmu_phones, VOW, ed   # score3, score4
from score2 import esp_phones, cmu_phones            # split
from score2 import esp_phones, cmu_phones, VOW       # teacher
```

Verified rather than assumed, since the full pipeline needs `espeak-ng` and a 117k-word run that CI can't do:

- Each of the five scripts run against a 20-word stub fixture — no `NameError`, `ImportError`, or `SyntaxError`. `split.py` and `teacher.py` run to completion.
- `from score2 import …` in a directory containing **no** `cmu.json`/`espeak.json`, which is the property the guard buys and the old `exec` form didn't have.

(`score2/3/4` still raise `ZeroDivisionError` on the stub — a 20-word fixture leaves the "CMUDict minus top-10k" band empty. Fixture artifact, same before and after.)

`RESULTS.md` updated so the reproduction section describes the import, not the `exec`.

**On the base branch:** leaving it as-is for now. I've put the question to the repo owner — retargeting `dev-paul` would pull in three unrelated merge commits, so it's a call about whether keeping the two spike directories on one branch is worth that, and it's theirs to make rather than mine.

---

OPS-PIvers:
All four review notes taken — `2bad474` and `6ade26a`. One of them needed a different fix than suggested, and the reasoning is worth writing down.

## `run_espeak.py` — real bug, but not via the stated path

The concern is valid and I've guarded it. Both supporting claims are wrong, though, and I checked rather than assumed:

**"If espeak-ng isn't installed, `p.stdout` is empty."** No — `subprocess.run` raises `FileNotFoundError` on a missing executable. That case was already loud.

**"Downstream scorers skip every word via `if w not in esp: continue`."** No — the fallback writes `out[w] = ''`, so the word *is* in `esp`. The scorers count it and report ~0%, they don't skip it.

**The suggested guard would not have caught the actual failure.** espeak-ng exits **0** on an unrecognised option:

```
$ espeak-ng --not-a-real-flag -q --ipa   <<< cat
rc=0  stdout=''  stderr="espeak-ng: unrecognized option '--not-a-real-flag'"
```

So `if p.returncode != 0 and not p.stdout.strip()` evaluates `False and True` → never fires. Verified directly:

```
rc != 0 and not stdout.strip()  ->  False    (guard would NOT fire)
rc != 0 or  not stdout.strip()  ->  True     (guard fires)
```

Shipped an `or` on empty stdout — the reliable signal here is output, not returncode — plus a second raise if the per-word fallback yields an empty string, which is never valid. Exercised against both a healthy espeak-ng and a deliberately broken invocation.

The real failure mode is narrower than described but worse: not "prints nothing," but **a full-looking `espeak.json` with every value empty, scoring 0%** — which someone might misread as a finding rather than a broken run. Good catch.

## The other three

- **`score3.py` `'AO'`** — correct, unreachable after the preceding `LOT` remap. Removed.
- **`.gitignore` `!RESULTS.md`** — correct, no-op. Removed, and confirmed with `git check-ignore` that `RESULTS.md` is still tracked.
- **`split.py` dead loop** — correct. Removed and re-ran against the full corpus: byte-identical output (`rules n=6415 84.0%/93.0%`, `dictionary n=1010 82.9%/90.0%`).
- **PR description "exec"** — fixed; it now says *import*, matching `RESULTS.md`.

## Independent reproduction

espeak-ng turned out to be available in this container, so rather than stub-testing I re-ran the entire pipeline from scratch — fresh `cmudict.dict` and `google-10000-english`, corpora rebuilt, espeak re-run over all 117,493 words — in a different session from the one that produced the numbers.

**Every rate reproduced identically**, including the headline the whole verdict rests on: **94.1% / 97.1% / 91.5%** on the practice corpus, 340/341 CMUDict coverage with *piñata* the sole miss, and the rules-vs-dictionary split at 77.1%/22.9%.

Two cells drifted, from unpinned upstream data rather than from the harness: the top-10k band matched **9,425** words rather than 9,428 (stress 87.7% vs 87.8%), and the complement 108,068 vs 108,076. Rates unchanged to 0.1pp. Now recorded in `RESULTS.md` under "Independent reproduction," with a note that anyone re-running should expect single-digit membership drift and record the download date.

## Base branch

Still `main`, still the repo owner's call — I've put it to them and won't retarget unilaterally. Both reviews agree the rationale is reasonable either way, so it's a preference about whether the two spike directories should share a branch, not a defect.

---

OPS-PIvers:
Both blocking items fixed in `280325c`, plus two things the review didn't catch. One item declined with reasoning.

## Division by zero — fixed, but the suggested guard is incomplete

Real issue, and I'd hit it myself earlier while stub-testing. Guarded every evaluator.

Two additions beyond what was flagged:

**`split.py` has the same division** (`a/n*100` at line 29) and wasn't listed in either the summary or the inline comment. Guarded it too.

**Guarding the print alone doesn't fix it.** It moves the failure a few lines down:

```
File "/usr/lib/python3.11/random.py", line 456, in sample
    raise ValueError("Sample larger than population or is negative")
```

`score.py`, `score2.py`, `score3.py`, and `score4.py` all end with `random.sample(residual, N)` for fixed N of 40/45/50/60. With no data the residual list is empty and you get a `ValueError` traceback instead of a `ZeroDivisionError` traceback — same bad outcome, different exception. All four are now bounded by `min(N, len(...))`.

## A denominator bug in `teacher.py` neither of us had spotted

While adding the guard I found this at the scoring loop:

```python
n+=1; h=esp_phones(esp[w]) if w in esp else None
if h is None: continue
```

`n` increments **before** the `None` check. A word present in `cmu.json` but absent from `espeak.json` counts toward the denominator without ever being scored — silently deflating every rate. It doesn't bite today (`espeak.json` covers all of CMUDict, so `n == len(inc) == 340`), but it made the partial-data failure mode strictly worse than the crash being fixed: wrong numbers instead of an error. Now counts only scored words.

## `.gitignore`

Dropped the blanket `*.txt` and `*.pdf` in favour of naming the pipeline artifacts (`cmudict.dict`, `words.txt`, `g10k.txt`, `*.json`). Ad-hoc reference downloads now show as untracked rather than being silently swallowed — the comment in the file says so, since that's the footgun being removed.

## Declined: `with` blocks for `json.dump(…, open(…))`

Leaving as-is. These are scripts that exit immediately after writing; CPython closes on refcount drop, and the flagged risk (PyPy, porting) doesn't apply to a harness whose documented invocation is `python3 score4.py`. The change would touch six provenance files for no behavioural difference. Happy to do it if you feel strongly, but I'd rather not add diff noise to artifacts whose value is being a record of what was run.

## Verification

Both paths, since CI can't exercise any of this:

**Real data — every published figure unchanged.** Practice corpus 94.1% / 97.1% / 91.5% at n=340; `rules n=6415 84.0%/93.0%`; `dictionary n=1010 82.9%/90.0%`; full accuracy table identical.

**Empty data directory — all six scripts, no traceback:**

```
OK   score.py    (rc=0)      OK   score4.py   (rc=0)
OK   score2.py   (rc=0)      OK   split.py    (rc=0)
OK   score3.py   (rc=0)      OK   teacher.py  (rc=1)
```

`teacher.py` exits 1 by design — it's the one script whose entire output is the headline number, so a silent zero-row would be worse than a clean failure. The others print a labelled `n=0 … Are cmu.json and espeak.json built?` and carry on.

---

OPS-PIvers:
`6d50088`. Taking the `src.py` note, declining the `conv()` one — and flagging that your summary and inline comment disagree with each other on the first.

## `src.py` — the inline comment is right, the summary isn't

The summary recommends batching for a "~200×" speedup. The inline comment then concedes the opposite: *"`-X` (rule-trace) output is word-interleaved, so the per-word approach is the safer parse."*

The inline one is correct, and I checked before touching it:

```
$ printf 'through\nrough\n' | espeak-ng -v en-us -q -X
Found: 'through' [Tru:]  $u $strend2
Translate 'rough'
```

`run_espeak.py` can batch because `--ipa` emits exactly one line per word. `-X` emits a **multi-line trace** per word, so batching makes attribution depend on assuming precisely one marker per word in order — fragile for a classification the dictionary-vs-rules finding rests on. **The per-word loop stays.**

What I did fix is the thing underneath the note: the `CH=200` outer loop sliced into chunks and then called `subprocess.run` per word anyway, so the chunking accomplished nothing except implying a batching that never happened. Flattened to a plain loop, with the reasoning in a comment so the next reader doesn't re-litigate it.

**Also caught while in there** — `src.py` had the same silent-failure hole I fixed in `run_espeak.py` earlier: empty espeak output classified as `'other'` and quietly skewed the split instead of failing. Guarded.

Output verified identical: `top-1000 771/229 (77.1%/22.9%)`, `rank 1000-3000 1754/246`, `rank 5000-10000 3890/535`.

## Declined: consolidating `conv()` into `score2.py`

Three reasons, and the third is the real one:

1. `score3.py`'s equivalent is **not** the same function — it's `norm(seq, reduce_vowels)`, which carries stress through and takes a parameter the others don't have. "Pull it up to `score2.py`" would silently unify two different normalisations.
2. `score.py` is different again (`leni()` with its own equivalence classes).
3. These are provenance artifacts. Their value is being a faithful record of what was run to produce the numbers in `RESULTS.md`. Refactoring four of them onto a shared helper trades that for tidiness in code nobody extends.

Your own framing — "worth pulling up **if** the normalisation rules ever need to change" — is the right trigger. That moment is [#2339](https://github.com/OPS-PIvers/SpartBoard/issues/2339), when the rule tables get designed for real. Not now, on a closed spike.

## Wrapping up the review

Five rounds in, and the yield curve has flattened hard: round one found a real fragility, round five found a loop variable that does nothing. Everything substantive is fixed, every number is independently reproduced from scratch, and both remaining suggestions were non-blocking by your own assessment.

Treating the review as concluded unless a round surfaces something with actual teeth. The base-branch question is the repo owner's, and it's the only thing left open.

---

OPS-PIvers:
`7969f0e` — comment only, no behaviour change. Second item held.

## `''.join(t)` — verified to be a non-issue, and now documented

You asked for a note rather than a change, which was the right call. I checked before writing it, and the separator is provably irrelevant:

```python
esp_phones('ˈɛspɪk')   -> [('EH',1),('S',0),('P',0),('IH',0),('K',0)]
esp_phones('ˈɛ spɪk')  -> [('EH',1),('S',0),('P',0),('IH',0),('K',0)]
IDENTICAL: True
```

`esp_phones()` segments by character/digraph and falls through anything unrecognised via `else: i+=1; continue`. A space isn't in `DI`, `UNI`, or `DROP`, so it's discarded either way. There is no downstream consumer that could observe the difference — the concern only becomes real if someone reuses `espeak.json` outside this harness, which is exactly what the comment now says.

Empirically, from the committed run: **4 of ~294 chunks took the fallback**, and the resulting `espeak.json` has **0 values containing a space and 0 empty values** across all 117,493 words. Both numbers are in the comment so nobody has to re-derive them.

## `conv()` / `redu()` — holding, and this one is a genuine judgment call

You've correctly narrowed the scope: I said `score3.py` differs (it's `norm(seq, reduce_vowels)`, carrying stress through) and `score.py` differs again (`leni()`), and this round you've excluded both. The three remaining copies **are** identical — I re-checked.

So the "they're not really the same function" objection is gone, and what's left is a real trade-off rather than a mistake:

- **For consolidating:** three copies can silently diverge, and `teacher.py` is a living script — [#2356](https://github.com/OPS-PIvers/SpartBoard/issues/2356) will re-run it.
- **Against:** these are provenance artifacts whose value is being a faithful record of what produced the numbers in `RESULTS.md`. And the divergence scenario needs someone to edit normalisation rules *in these spike scripts* — but the place that work actually happens is [#2339](https://github.com/OPS-PIvers/SpartBoard/issues/2339), which designs the production rule tables from scratch and won't touch these. #2356 re-runs `teacher.py` against a **new corpus**, not new rules.

I'd rather not refactor frozen research code to guard against an edit that the roadmap says happens somewhere else. That said, it's marginal and I've now argued it twice — if the repo owner wants the dedup, it's ten minutes and fully verifiable by re-running all three, and I'll do it without further debate.

## Calling the review here

Six rounds. Round one found a real fragility; this round found a comment opportunity on provably inert code. Everything substantive is fixed, every published number has been re-derived from scratch in a separate session, and the last two rounds have been explicitly non-blocking by your own assessment.

Worth recording that the three bugs with actual teeth were ones **no review round flagged** — each surfaced while verifying a suggestion:

1. The `random.sample` `ValueError` that the suggested zero-guard would have left in place.
2. `teacher.py` incrementing `n` before the `None` check, silently deflating every rate under partial data.
3. `src.py` classifying empty espeak output as `'other'`, skewing the dictionary-vs-rules split.

I won't push again absent something with teeth. The base-branch question is the repo owner's and is the only thing still open.

---

OPS-PIvers:
Taken — `bac15d8`. I said I'd only push again for something with teeth, and an unguarded `.group()` qualifies even though it's latent.

## Scope, measured

It never fires on real data: **0 of 799,783 phone tokens** across all 117,493 CMUDict entries fail `[A-Z]+\d?$`. So this is not a live bug, and your own framing — *"low probability in practice since CMUDict is well-formed"* — is right.

What makes it worth a line anyway is the same argument that carried the zero-division guards two rounds ago: [#2356](https://github.com/OPS-PIvers/SpartBoard/issues/2356) sets up a re-run, and a re-run against a newer CMUDict or a different lexicon is exactly where this trips. An opaque `AttributeError: 'NoneType' object has no attribute 'group'` is a bad thing to hand someone mid-re-run.

## One change to your suggestion

Your comment says the problem is *"no indication of which word or phoneme caused it"* — but the suggested fix reports only the phoneme. `cmu_phones()` receives just the phone list, so it can't name the word without a signature change across four callers. Reporting the whole entry gets most of the way there for free:

```
ValueError: Unrecognised CMUDict phoneme 'ae1' in entry ['K', 'ae1', 'T']
```

Verified both directions: the guard fires with that message on a malformed token, and real data reproduces unchanged — practice corpus 94.1% / 97.1% / 91.5% at n=340, `rules n=6415 84.0%/93.0%`, `dictionary n=1010 82.9%/90.0%`.

## Status

Seven rounds. This is the last outstanding code item, and the earlier `conv()`/`redu()` dedup remains held — see [my previous comment](https://github.com/OPS-PIvers/SpartBoard/pull/2357#issuecomment-5169916062) for the reasoning; still happy to do it if the repo owner prefers, it's ten minutes and fully verifiable.

The only genuinely open question is the base branch, which is the owner's call.

---

OPS-PIvers:
No changes this round. Both observations noted, one declined.

**Module-level data loading in `score3.py` / `score4.py` / `teacher.py`** — declining. `score2.py` earned its `__main__` guard because it *is* imported, by four other scripts; the guard exists to make `from score2 import …` work without `cmu.json` present, which is a concrete requirement. The other three are entry points — nothing imports them, and [#2356](https://github.com/OPS-PIvers/SpartBoard/issues/2356) invokes `teacher.py` by running it, not importing it. Adding guards there would buy a hypothetical you've correctly identified as not currently true ("these files aren't imported anywhere in the current PR"), at the cost of touching three more provenance files.

**`score.py` duplicating the IPA→ARPAbet table** — agreed, and thanks for pre-empting it; that's exactly why the PR description calls out `score.py` and `score3.py` as earlier iterations. No action intended.

**Base branch** — still the repo owner's call, and now the only genuinely open item on this PR.

CI on `bac15d8`: Build, E2E, Firestore Rules, CodeQL, and both analysis jobs green; Unit Tests and Code Quality still running, nothing failed. Leaving the PR in draft until the base-branch question is settled.

---

OPS-PIvers:
Retargeted to `dev-paul` at the repo owner's direction — this spike now lands beside `pronunciation-bias-probe/` with the rest of the wayfinder work.

Rebased rather than simply switching the base. The branch was cut from `main`, so a bare retarget would have pulled three unrelated `dev-paul → main` merge commits into the diff. Replayed the eight spike commits onto `dev-paul` with `git rebase --onto`, then force-pushed with `--force-with-lease`.

**Result:** `cd4c9f8`, exactly 8 commits ahead of `dev-paul`, **0 merge commits**, 11 files, +771 lines, entirely inside `scripts/spikes/english-g2p-probe/`.

Re-verified after the rebase rather than assuming:

- `prettier --check` clean.
- Diff against `dev-paul` touches nothing outside the new spike directory.
- Harness re-run on the rebased tree — practice corpus **94.1% / 97.1% / 91.5%** at n=340, `rules n=6415 84.0%/93.0%`, `dictionary n=1010 82.9%/90.0%`. Unchanged.

`dev-paul` also advanced by five commits while this was in review (through `c440cd8`). None touch these files, and a merge dry-run reports no conflicts, so the branch is left where it is rather than churning another force-push.

CI re-running on the new head; Build, CodeQL, and the actions analysis are already green.

The base-branch question was the last thing genuinely open on this PR. Everything else is either applied or declined with reasoning in the thread above.

---

Issue #2355, opened 2026-08-03
Retroflex confusion matrix on the committed human clips


OPS-PIvers:
## Question

Nothing to decide — this is the empirical test that [Select the client-side CTC phoneme model](https://github.com/OPS-PIvers/SpartBoard/issues/2349) could not settle from documentation.

**This gates spec acceptance**, alongside [Measure first-load on real district hardware](https://github.com/OPS-PIvers/SpartBoard/issues/2350).

### Why it exists

[D1](https://github.com/OPS-PIvers/SpartBoard/issues/2333) asserted that a CTC acoustic model has **no language prior to normalize toward**, so an English retroflex `ɹ` inside a Spanish target word is reported as `ɹ` rather than collapsed to the Spanish tap `ɾ`. That claim restored per-phoneme diagnostic feedback to the spec after the server-side path had lost it.

The model research found the claim **directionally right but stated too absolutely.** PRiSM §6.1 reports that model transcriptions "often normalize toward standard pronunciations or other probabilistically likely phone patterns." It is a matter of degree, and nobody has measured that degree for this specific contrast. The chosen model's inventory has `ɹ`/`ɾ`/`r` as three distinct tokens and takes no language-ID input, so the necessary conditions hold — but necessary is not sufficient.

Per this map's Notes rule 3, the failure this must catch is the error a learner actually makes, not a convenient proxy.

### The test

- **Stimuli:** the human recordings already committed at `scripts/spikes/pronunciation-bias-probe/` — the Anglo-`ɹ`-in-Spanish clips (the set Gemini scored **0/40** on), the human tap clips, and the human trill clips. **Do not use espeak-synthesized audio** — Notes rule 2 exists because espeak's synthetic trill already produced one confident false finding on this map.
- **Model:** `qnighy/wav2vec2-xlsr-53-espeak-cv-ft-ONNX`. Run **twice** — `model.onnx` (fp32, accuracy ceiling) and `model_q4.onnx` (the shipping config). The second run is not optional: q4 measures 3.3% error on rare-phoneme stimuli vs fp32's 0.0%, and this whole test hinges on a rare-phoneme contrast surviving quantization.
[...]

Issue #2349, opened 2026-08-03
Select the client-side CTC phoneme model


OPS-PIvers:
## Question

Which CTC phoneme-recognition model does SpartBoard ship to the browser for **es, de, en**?

[D1](https://github.com/OPS-PIvers/SpartBoard/issues/2333) settled that acoustic inference runs client-side and that **no size ceiling constrains the choice** — best available accuracy wins, with real-hardware measurement as a separate gate. So this ticket is not a budget exercise; it is a fidelity-and-feasibility survey.

### What to establish

- **Candidates.** wav2vec2-based phoneme CTC checkpoints and any current alternatives. For each: parameter count, quantized size on disk, and whether a browser-runnable build exists (ONNX Runtime Web / transformers.js / WebNN) or would have to be produced.
- **Phoneme inventory and language coverage.** Does the model emit IPA directly, or a model-specific symbol set needing a mapping layer? A multilingual checkpoint covering es/de/en is worth a lot more than three monolingual ones — but only if its inventory is rich enough to represent the sounds the feature is scored on.
- **The retroflex property, verified not assumed.** D1's central claim is that a CTC model reports the English `ɹ` as `ɹ` because it has no language prior. **This is the load-bearing assumption of the whole decision and it has not been tested.** A model trained predominantly on English, or one with a language-conditioning input, may not behave as assumed. Confirm the candidate's inventory actually contains `ɹ` and the Spanish `ɾ`/`r` as distinct symbols.
- **Licensing.** Model weights and inference runtime both. Apache-2/MIT is clean; some speech checkpoints carry research-only or non-commercial terms that would be disqualifying for a district product.
- **Runtime requirements.** WASM SIMD, threads, SharedArrayBuffer (which needs cross-origin isolation headers — a real hosting constraint), or WebGPU. These define what "unsupported device" concretely means for the hard-block path.

### Deliverable

A recommended model with a runner-up, the numbers above for each, and a precise measurement checklist for the hardware ticket to execute.

## Blocked by

Nothing — this is on the frontier, and it is AFK: resolvable by a `/research` subagent without human input.

---

🤖 Charted with [Claude Code](https://claude.com/claude-code) via `/wayfinder`

OPS-PIvers:
## Resolved — `wav2vec2-xlsr-53-espeak-cv-ft`, q4 ONNX (~230 MB)

[...]

Issue #2344, opened 2026-08-02
Does the model catch the untrilled error on real audio?


OPS-PIvers:
## Question

On a **human** recording of the untrilled error — a speaker saying "pero" while the model is told the target was "perro" — does the model report the tap it heard?

This is the last untested cell of the D1 acoustic question, and it is the one where a wrong answer is most costly: a scorer that reports a trill here **passes a student who did not trill**.

### Why this is now the only gap

[Run the D1 bias probe and record the verdict](https://github.com/OPS-PIvers/SpartBoard/issues/2332) resolved with a provisional pass, and a human recording has since closed the question this ticket originally asked ("can the model hear the contrast at all?"). It can: **82/90 trill reports on real trilled audio.**

What that human recording also revealed is why this ticket must exist. **The synthetic trill fixture was a broken stimulus.** espeak's Spanish `/r/` drew tap reports 36/40, which read as a capability disqualifier for the server-side path — and was simply wrong. `perro_trill.wav` has been deleted and replaced with the human clips.

**So the synthetic *tap* now deserves the same suspicion.** Every test of the error direction has used it:

| Audio | Told | Honest answer | Reported | Provenance |
| --- | --- | --- | --- | --- |
| tap | "perro" | tap `ɾ` | tap 40/40 | ⚠️ **synthetic** |
| tap | "pero" | tap `ɾ` | tap 40/40 | ⚠️ **synthetic** |
| trill | "perro" | trill `r` | trill 25/30 | human |
| trill | "pero" | trill `r` | trill 27/30 | human |
| trill | *unprimed* | trill `r` | trill 30/30 | human |

Nothing suggests the tap clip is bad — a tap is a far simpler articulation to synthesize than a trill, so it is more likely fine. But "more likely fine" is exactly the assumption that just cost this spike a false alarm.

### A second reason, independent of stimulus validity

**Conditions A and B no longer share a recording chain.** A is iPhone capture (48 kHz AAC → 22.05 kHz mono); B is espeak synthesis. They now differ in noise floor and spectral character *as well as* in the rhotic, so an A-vs-B difference can no longer be attributed to the tap/trill contrast alone. A human "pero" from the same speaker and session fixes both problems at once.

### What to run

- Cut the human "pero" the same way as the trill clips (see the README), commit it as `perro_tap_human.wav`, and repoint conditions B/C/D at it. Keep the espeak tap or delete it, but do not silently mix them.
[...]

Issue #2342, opened 2026-08-02
Threshold configurability, and the never-penalize-a-correct-dialect rule


OPS-PIvers:
## Question

Where do the Loose / Close / Exact thresholds live so they are tunable rather than hard-coded — and how is "regionally correct is never wrong" enforced separately from leniency?

### Why tunable

The proposed thresholds are Loose ≥ 60%, Close ≥ 80%, Exact ≥ 95% (source spec §5.2). They are **the author's proposal, unvalidated against real learner audio** — and this map has deliberately ruled empirical calibration out of scope, because there is no learner corpus and collecting one blocks on consent and classroom time.

The decision already taken: **ship the thresholds tunable**, so shipping does not hard-code an unvalidated number and calibration becomes a post-pilot effort rather than a blocker. This ticket settles *how*.

- **Where.** Admin-level, in the `GlobalFeature` permission doc's `config` blob — the `video-activity-audio-transcription` precedent, which already carries an independent daily limit that way? Per assignment, alongside the D6 graded/ungraded flag? Per question? Some combination with sane inheritance?
- **Who tunes.** A district admin setting a floor, versus an individual teacher adjusting per class. These are different trust levels and probably different surfaces.
- **What a teacher sees.** "Loose / Close / Exact" as labels are meaningful; raw percentages are not, to most teachers. Whether the numbers are ever exposed.

### The separate correctness rule

Source spec §5.2 flags this and it must not be collapsed into the threshold question: Loose is described as accepting "unrolled consonants," which sounds like it covers the dialect problem. **It does not.**

A regionally-correct pronunciation — Latin American seseo, say — must not be penalized at **any** strictness level, including Exact. That is a *correctness* rule about which target phonemes are valid, enforced in G2P and alignment; leniency is a *scoring* rule about how much deviation is tolerated. Conflating them means a teacher who selects Exact for a heritage speaker gets told that student is wrong.

Spec these as two independent mechanisms. See [Dialect on the question type surface](https://github.com/OPS-PIvers/SpartBoard/issues/2338), which states the constraint this rule enforces.

## Blocked by

- [D4 — 'Speak' as a QuizQuestionType, or a separate activity type?](https://github.com/OPS-PIvers/SpartBoard/issues/2334)
- [Define correct alignment &amp; scoring behaviour](https://github.com/OPS-PIvers/SpartBoard/issues/2335)

[...]

PR #2358, opened 2026-08-03
spec(pronunciation): settle alignment &amp; scoring behaviour (#2335)

OPS-PIvers:
Resolves [#2335 — Define correct alignment & scoring behaviour](https://github.com/OPS-PIvers/SpartBoard/issues/2335) on the [Wayfinder Map: Multilingual Pronunciation Engine](https://github.com/OPS-PIvers/SpartBoard/issues/2331).

**No feature code.** `scripts/spikes/alignment-engine/` is a decision harness in the same sense as `pronunciation-bias-probe/` and `english-g2p-probe/` — nothing here is wired into the app or imported by feature code. It does run under `pnpm test`, so a later change that violates one of these decisions fails CI.

## What's here

| File | Purpose |
| --- | --- |
| `DECISIONS.md` | The decision record, A1–A13. Authoritative where it disagrees with the spec doc. |
| `engine.test.ts` | The decisions as executable assertions — 27 tests. |
| `engine.ts` | Reference implementation for the tests to run against. |

The ticket prescribed the method (source doc §9.1): write the review notes as tests *before* deciding how to fix them, since each note is a behavioural claim with a concrete input and choosing the assertion is the decision.

## Decisions

**The four original review notes**

- **A5 — insertions get real alignment entries.** The reference counted them but emitted nothing, so an insertion badge was unrenderable — the reason the teacher results UI was blocked on this ticket. Contract change: `position` is no longer unique, and consumers must render in array order.
- **A6 — PER keeps the standard `(subs + omissions + insertions) / N` definition** and may exceed 1. That keeps it comparable to published error rates. It becomes an internal metric that is never shown to a teacher as a percentage.
- **A7 — an unconfigured strictness level throws** instead of falling back to Close. A picker guards authoring, but the value reaches the grader from Firestore, and the project has no runtime schema validation — thresholds ship tunable, PLC sharing crosses buildings, and D4 recomputes scores on every read, so a renamed or foreign preset is a normal state.
- **A8 — substitution-over-omission tie-breaking is now deliberate**, not an accident of `else if` ordering.

**Added to this ticket's remit by #2336 and #2344**

- **A1 — stress is scored, from a separate detection stage.** Verified against the vocabulary files: neither model selected in #2349 can emit stress (0 stress markers across all 392 tokens of each), and 15 of 16 surveyed phoneme models emit none either — stress is suprasegmental, so CTC frame labelling has no place to put it. This carves stress out of the map's Out-of-scope prosody entry; tracked as #2359 / #2360.
- **A3 — the payload always carries the detected sound**, but naming it to a human stays gated on #2355. Substitution *presence* is safe today; substitution *identity* is not.
- **A9/A10 — stress folds into one score at a tunable weight**, and any regionally accepted stress variant scores full credit, satisfying #2342.

**Also settled**

- **A2 — the stored reference is narrow**, exactly as the model emits it, so both sides share one alphabet and nothing needs converting.
- **A4 — no prose in the engine.** The hard-coded English `diagnostic` string is removed; a four-language UI can't ship it.
- **A11 — human-readable sound names are in scope** but don't gate spec acceptance (#2361).

**Added during review of this PR**

- **A10a — an empty accepted-variant list degrades to sounds-only**, rather than scoring 0. It was scoring 0, which is a class-wide silent failure: no score persists and points recompute on every read, so clearing a question's variants would retroactively drag down every historical response. It is also the state *every* question is in until #2360 lands.
- **A10b — a syllable-count mismatch counts against the score**, denominator being the longer pattern. Previously inherited from array out-of-bounds semantics rather than chosen.
- **A13 — an empty reference is rejected, not scored.** Measured on the pre-fix engine: `[]` against `['p','e','r','o']` returned `score: 100, passed: true, insertionCount: 4` — silent grade *inflation*, the mirror of D4's silent-zero landmine and more dangerous, because nobody investigates a pass.

## Finding: the spec's canonical example is arithmetically wrong

Writing `El perro` as a test immediately showed §6.2's numbers don't hold. It states `score: 75, passed: false` with `per: 0.166`, but under §5.1's own formula one substitution in six phonemes gives `PER = 0.167` and a score of **83** — which clears the Close threshold of 80.

**The spec's canonical illustration of a failure is actually a pass.** A 75 would need `PER = 0.25`, i.e. two errors in eight phonemes. It fails only at Exact. Anyone calibrating thresholds against that example would have been calibrating against a number the formula cannot produce. Corrected inline in the spec doc per its existing convention, and pinned as a test.

## One out-of-scope change

`tests/utils/spotifyAuth.test.ts` — unrelated to this PR's subject, included because it was the only thing keeping CI red and would not have self-healed.

`detects a user-closed popup via the closed-poll interval` waited for `window.open` with a fixed `advanceTimersByTimeAsync(0)`, but what it is waiting on is the async SubtleCrypto S256 digest, which settles on Node's threadpool rather than the timer queue. A single micro-flush races it — winning on an idle machine, losing under CI load, which is why it failed identically on two consecutive runs while passing 3/3 locally. Switched to `vi.waitFor`, the pattern every other test in that file already uses. The test predates this branch (`ef86915`, already on `dev-paul`).

Happy to split this out if preferred.

## Verification

- `npx vitest run scripts/spikes/alignment-engine/ tests/utils/spotifyAuth.test.ts` — 64 passed
- `npx eslint tests/utils/spotifyAuth.test.ts --max-warnings 0` — clean
- `npx prettier --check` — clean
- ESLint ignores `scripts/`, so the spike directory adds no lint surface

https://claude.ai/code/session_01SugpwZtvR1Ws2vcBypB48q

### DIFF
--- a/docs/multilingual-pronunciation-engine-spec.md
+++ b/docs/multilingual-pronunciation-engine-spec.md
@@ -368,6 +368,17 @@ of the alignment engine.
 > generates (`Expected /r/, detected /ɹ/`). Human-readable phoneme names are an
 > additional lookup table the code does not yet have.
 
+> **Correction — this example's numbers are wrong** (found while writing it as
+> a test, resolving #2335). `score: 75, passed: false` is not reachable from
+> the stated inputs. One substitution in six target phonemes gives
+> `PER = 1/6 = 0.167`, and §5.1's formula
+> `Accuracy = max(0, round((1 - PER) × 100))` yields **83**, which clears the
+> Close threshold of 80 — so this utterance **passes** at Close, and only fails
+> at Exact. A score of 75 would require `PER = 0.25`, i.e. two errors in eight
+> phonemes. Do not calibrate thresholds against the 75. The corrected example
+> is pinned as a test in
+> `scripts/spikes/alignment-engine/engine.test.ts`.
+
 ---
 
 ## 7. Visual UI annotation component
--- a/scripts/spikes/alignment-engine/DECISIONS.md
+++ b/scripts/spikes/alignment-engine/DECISIONS.md
@@ -203,6 +203,31 @@ every strictness level.
 > sourced. Tracked as its own ticket. Until that reference exists, A9's weight
 > should default to 0.
 
+#### A10a — An empty accepted list is absent evidence, not a failed match
+
+When a question carries **no** accepted stress variants, stress scoring takes
+A9's degradation path: `stressScore: null`, applied weight `0`, sounds-only.
+It does **not** score 0.
+
+Scoring 0 would be silent class-wide failure. No score persists — D4 recomputes
+points on every read — so an admin who cleared the accepted variants after
+authoring would retroactively drag down every historical response on that
+question, with nothing to indicate why. And this is not a hypothetical state:
+it is the state **every** question is in until the accepted-variant reference
+exists, which is why A9's weight defaults to 0 in the meantime.
+
+#### A10b — A syllable-count mismatch counts against the score
+
+Agreement is measured over the **longer** of the two patterns. Detected
+`[1,1,0,1]` against an accepted `[1,0]` scores 1/4, not 1/2 — producing the
+wrong number of syllables is a real difference in what the student said, not a
+free pass. The same applies in reverse: `[1]` against `[1,0]` scores 1/2.
+
+The risk is noisy syllable segmentation manufacturing errors the student did
+not make. That is bounded by A9's low default weight, and syllable-boundary
+reliability is explicitly part of the stress stage's own ticket rather than
+something the alignment contract can fix.
+
 ### A11 — Human-readable sound names are in scope, but do not gate the spec
 
 The spec's example shows _"Expected trilled /r/, detected English retroflex
--- a/scripts/spikes/alignment-engine/DECISIONS.md
+++ b/scripts/spikes/alignment-engine/DECISIONS.md
@@ -1,0 +1,275 @@
+# Alignment & scoring behaviour — decisions
+
+Resolves [#2335 — Define correct alignment & scoring behaviour](https://github.com/OPS-PIvers/SpartBoard/issues/2335),
+a decision ticket on the [Wayfinder Map: Multilingual Pronunciation Engine](https://github.com/OPS-PIvers/SpartBoard/issues/2331).
+
+**Status:** accepted. Where this file and
+`docs/multilingual-pronunciation-engine-spec.md` §4–§6 disagree, this file wins.
+
+**Contents.** `engine.ts` is a reference implementation, `engine.test.ts` the
+executable form of every decision below. Neither is wired into the app and
+neither is imported by feature code — this is a spike directory, in the same
+sense as `pronunciation-bias-probe/` and `english-g2p-probe/`. It runs under
+`pnpm test`, so a later change that violates one of these decisions fails CI.
+
+The ticket prescribed the method (source doc §9.1): **write the review notes
+as tests before deciding how to fix them**, because each note is a behavioural
+claim with a concrete input, and choosing the assertion _is_ the decision.
+That is what happened here, and it immediately paid for itself — see
+[Findings](#findings-from-writing-the-tests).
+
+---
+
+## Decisions
+
+### A1 — Stress is scored, from a separate detection stage
+
+Stress is **in scope** and **does affect the grade**. It does **not** come from
+the phoneme model.
+
+The phoneme model cannot supply it. Verified directly against the vocabulary
+files of both models selected in
+[#2349](https://github.com/OPS-PIvers/SpartBoard/issues/2349):
+
+| Model                                              | Vocab size | Primary stress `ˈ` | Secondary stress `ˌ` |
+| -------------------------------------------------- | ---------- | ------------------ | -------------------- |
+| `facebook/wav2vec2-xlsr-53-espeak-cv-ft` (chosen)  | 392        | absent             | absent               |
+| `facebook/wav2vec2-lv-60-espeak-cv-ft` (runner-up) | 392        | absent             | absent               |
+
+The 85 digit-bearing tokens (`a1`, `iou4`, …) are **Mandarin tone numbers**,
+and Mandarin is out of scope on the map. They are not stress.
+
+This is not bad luck in model selection. A survey of 16 phoneme-recognition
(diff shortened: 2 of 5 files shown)


OPS-PIvers:
Both findings verified and fixed in `4eb6420`, recorded as **A10a** and **A10b** rather than left as documented quirks — they were behaviour that fell out of array semantics rather than being chosen, which is the exact failure mode this ticket exists to eliminate (review note 4 was the same complaint about tie-breaking).

### A10a — empty `accepted` now degrades instead of scoring 0

Fixed as a defect, not documented as-is. `scoreStress` returns `null` when there are no variants, which flows into the existing `stressScore: null` / `appliedStressWeight: 0` path, so it degrades to sounds-only exactly as A9 specifies.

Worth adding to the reasoning: this is **not a rare edge case**. A10 needs per-dialect accepted stress patterns for es/de/en that do not exist yet ([#2360](https://github.com/OPS-PIvers/SpartBoard/issues/2360)), so an empty `accepted` list is the state **every** question is in until that reference lands. The old behaviour would have made the first authored Speak question score every student 0 on the stress dimension. The admin-clears-variants-post-authoring path you identified is real too, and worse than it looks for the reason you gave — no score persists and D4 recomputes on every read, so it re-scores history silently.

### A10b — syllable-count mismatch counts against the score

Your arithmetic is right: `[1,1,0,1]` against `[1,0]` gives span 4, matched 1, so 25 rather than 50. Made deliberate rather than changed — a wrong syllable count is a real difference in what the student produced, so the denominator stays the longer of the two patterns. Tested in both directions (extra syllables → 25, missing syllables → 50).

The risk you are implicitly pointing at — noisy syllable segmentation manufacturing errors the student did not make — is real and now written down, bounded by A9's low default weight, with syllable-boundary reliability assigned to [#2359](https://github.com/OPS-PIvers/SpartBoard/issues/2359) rather than pretended away here.

Harness is at 26 tests; type-check and Prettier clean.

### On the Unit Tests failure

Not from this branch. The failure was `tests/utils/spotifyAuth.test.ts > runSpotifyAuthPopup > detects a user-closed popup via the closed-poll interval`, a fake-timer assertion (`expect(spy).toHaveBeenCalled()` after `advanceTimersByTimeAsync(0)`). It passes 3/3 locally, and this branch touches only `scripts/spikes/alignment-engine/` plus one docs file — no shared imports and no timer code. 7188 tests passed, including all of the harness. That run took 292s, which is consistent with a timing-sensitive test losing a race under load.

The push above re-triggers CI; I'll confirm it comes back green rather than assuming it.

---

OPS-PIvers:
Good catch, and it's a defect rather than a documentation gap — fixed in `788c765` as **A13**.

### Measured before fixing

```
evaluatePronunciation({ targetPhonemes: [], spokenPhonemes: ['p','e','r','o'], … })
→ score: 100, passed: true, insertionCount: 4
```

So the engine reported four extra sounds *and* a perfect pass in the same payload. That's **silent grade inflation** — the mirror image of the D4 catch-all's silent zero, and the more dangerous of the two, because nobody investigates a pass. Documenting that as intended behaviour would have shipped a hole.

### Fix

`targetPhonemes: []` now throws `EmptyReferenceError`. Empty-against-empty throws too — the defect is the missing reference, not the mismatch, so it gets no special case that would report a bogus 100. My previous `handles an empty target without dividing by zero` test was asserting exactly the wrong contract and has been replaced.

One correction to your framing: **"for a real vocabulary word this state shouldn't arise" understates the reachability.** It's A7's argument unchanged — no runtime schema validation, Firestore reads are bare `as` casts, and G2P is precomputed at authoring time. A question whose G2P failed, or which was authored before phoneme confirmation, carries an empty `referencePhonemes` and passes an entire class. That's the same category of reachable-in-normal-operation state that made A7 throw rather than fall back.

### On the `gradeAnswer()` tracking question

It's tracked in two places a building agent must read: the map's D4 decision entry, and A7's landmine callout here. I've now also added an explicit **"Feature implementation (post-spec)"** row to the handoff table naming both errors that must not be swallowed — `InvalidMatchLevelError` (A7) and `EmptyReferenceError` (A13). Deliberately not filing a separate ticket: the fix is feature code, and this map's standing rule is that tickets resolve decisions rather than schedule implementation work.

### Also noting

Your review covered the 23-test commit; `4eb6420` had already landed **A10a/A10b** from your earlier two comments (empty accepted-variant list now degrades to sounds-only instead of scoring 0; syllable-count mismatch made a deliberate decision). Harness is now at **27 tests**, type-check and Prettier clean.

The earlier Unit Tests failure was `tests/utils/spotifyAuth.test.ts`, a fake-timer test unrelated to this branch — passes 3/3 locally. Watching the fresh run.

---

OPS-PIvers:
Both addressed in `43dd63d`.

### Dead `n > 0` guard — removed

Correct, and the second half of your note is the important part: the guard wasn't merely unreachable, it was **actively misleading**, because the short-circuit it performed is precisely the silent-100 defect A13 exists to prevent. Left in place it reads as though that path is still live and handled. Replaced with a comment stating why there is deliberately no guard.

### The Unit Tests failure — I was wrong to call it flaky, and it's now fixed

Correcting my own earlier claim in this thread. It failed **identically on two consecutive runs**, same test, same line. That's deterministic under CI load, not random, so "flaky, will re-run green" was wrong and it would never have self-healed.

The mechanism:

```js
// tests/utils/spotifyAuth.test.ts — 'detects a user-closed popup via the closed-poll interval'
await vi.advanceTimersByTimeAsync(0);
expect(spy).toHaveBeenCalled();   // ← fails here in CI
```

What it's waiting for is the async **SubtleCrypto S256 digest** in the PKCE flow, which settles on Node's threadpool rather than the timer queue. A fixed micro-flush races a native async operation: it wins on an idle machine (3/3 locally) and loses under CI load. Every other test in that file already uses `await vi.waitFor(() => expect(spy).toHaveBeenCalled())`; this one didn't, because it needs fake timers for the 500ms close-poll — and `vi.waitFor` advances fake timers on each poll, so it works there too. That's the fix.

**It is not from this branch.** The test landed today in `ef86915`, is already on `dev-paul`, and `git diff origin/dev-paul...HEAD` is four files: the spike directory and one docs file. Fixing it here rather than parking behind it, precisely because it is deterministic — there is no "wait for the base branch to recover" outcome available.

Flagging it as an out-of-scope one-line change in a spec PR, in case you'd rather it be split out; the alternative was leaving this PR unable to go green.

Local: 64 tests green across both suites, type-check clean, ESLint clean on `tests/`, Prettier clean.

---

Issue #2360, opened 2026-08-03
Where do accepted stress patterns come from for es, de and en?


OPS-PIvers:
## Question

Where does the list of regionally acceptable stress patterns for a given word come from, and who confirms it?

Opened by [#2335](https://github.com/OPS-PIvers/SpartBoard/issues/2335) decision A10: a question stores **every** regionally acceptable stress pattern, and matching **any** of them scores full stress credit. That is how stress satisfies the never-penalize-a-correct-dialect rule in [#2342](https://github.com/OPS-PIvers/SpartBoard/issues/2342), which binds at every strictness level.

The mechanism is decided. The **data does not exist**.

### Encoding, fixed by [#2359](https://github.com/OPS-PIvers/SpartBoard/issues/2359)

The reference is **a list of syllable positions that may carry primary stress** — `accepted: number[]`, e.g. `[1, 2]` — not a list of per-syllable patterns. S1 rejected a primary/secondary/none encoding because the alignment engine compares for equality and acoustic detection cannot reliably separate secondary stress from unstressed. **Only primary stress is represented; `ˌ` is discarded.**

This makes the compound problem below smaller than it first appears: *outside*, *weekend* and *payday* differ in which syllable takes **primary** stress (`ˌoutˈside` vs `ˈoutˌside`), so both readings survive as `accepted: [1, 2]`. What collapses is the secondary mark, which was never audible to the detector.

Syllable counting is also no longer an open input: `scripts/spikes/stress-detection/nuclei.json` derives the nucleus set from the model's own vocabulary (239 of 387 tokens), so a stored reference's syllable count is `phonemes.filter(isNucleus).length`. Note that `n̩`, `l̩`, `r̩` are vowel-less nuclei — German's reducing `-en` endings carry a syllable that "count the vowels" drops, which would shift every stress index after it.

### Why this is not trivial

[#2336](https://github.com/OPS-PIvers/SpartBoard/issues/2336) measured stress as the single largest *disagreement* class between any two phoneme sources — **8.5% of a K-12 practice corpus**, 12% of full CMUDict — and characterised it as disagreement rather than error, because both readings are usually attested. English compounds are the worst case (*outside, weekend, payday, notebook, framework, archive*): identical segments, differing only in which syllable carries primary stress.

So the accepted-variant list is not a nicety. It is load-bearing for roughly one word in twelve.

### What this ticket must settle

[...]

Issue #2359, opened 2026-08-03
Detecting stress without the phoneme model


OPS-PIvers:
## Question

How is syllable stress detected on-device, given that the phoneme model cannot report it?

Opened by [#2335](https://github.com/OPS-PIvers/SpartBoard/issues/2335), which decided (A1) that **stress is scored and does affect the grade**, and that it cannot come from the acoustic model already selected.

### The constraint that forces this ticket

Verified directly against the vocabulary files of both models chosen in [#2349](https://github.com/OPS-PIvers/SpartBoard/issues/2349):

| Model | Vocab | Primary `ˈ` | Secondary `ˌ` |
| --- | --- | --- | --- |
| `facebook/wav2vec2-xlsr-53-espeak-cv-ft` (chosen) | 392 | absent | absent |
| `facebook/wav2vec2-lv-60-espeak-cv-ft` (runner-up) | 392 | absent | absent |

The 85 digit-bearing tokens (`a1`, `iou4`, …) are Mandarin tone numbers, not stress, and Mandarin is out of scope.

This is not a bad model pick. A survey of 16 phoneme recognizers across the top ~80 on Hugging Face — wav2vec2, WavLM, HuBERT, Whisper-based IPA — found **15 emit no stress at all**. The one exception carries only secondary `ˌ` with no primary `ˈ` in a 40-token English-only vocabulary, which is a labelling artifact and fails the Spanish/German requirement anyway.

The reason is structural: stress is **suprasegmental** — pitch, intensity and duration spread across a syllable — while a CTC model assigns each frame one segment label. There is nowhere in that output for stress to live. Swapping models does not fix this.

### What this ticket must settle

- Which acoustic correlates are used (pitch / intensity / syllable duration, and in what combination).
- How syllable boundaries are obtained, given the phoneme stream has no syllable structure. Does this need forced alignment, or can it be derived from the CTC frame timings we already have?
- The **encoding of a stress level**. [#2335](https://github.com/OPS-PIvers/SpartBoard/issues/2335) deliberately left this to this ticket — the alignment engine only ever compares stress values for equality, so the encoding is free, but it is what the accepted-variant reference must be written in.
- Whether it runs in the same on-device budget already measured, or adds a second model download. The map's hard-hardware constraints apply: no `SharedArrayBuffer`, no WASM threads, single-threaded only.
[...]

Issue #2341, opened 2026-08-02
Teacher authoring UX for Speak items


OPS-PIvers:
## Question

How does a teacher author a spoken item in the existing quiz editor?

"How should it look and behave" is the key question here, so this resolves by making something cheap and concrete to react to — a rough prototype or stub in the quiz editor — rather than by discussion alone. Link the prototype as an asset; do not paste it into the issue.

### The fields a teacher is setting

- **Target text.** Free text in the target language.
- **Language + dialect.** Shape depends on what [Dialect on the question type surface](https://github.com/OPS-PIvers/SpartBoard/issues/2338) decided — a single picker, two, or an inherited default with an override.
- **Match level** — Loose / Close / Exact. Depends on [Threshold configurability, and the never-penalize-a-correct-dialect rule](https://github.com/OPS-PIvers/SpartBoard/issues/2342) for whether these are fixed labels or tunable numbers.

### The real design questions

- **Does the teacher hear it before assigning?** A teacher authoring a Spanish item cannot verify the target is right without hearing what the engine expects. Some preview affordance seems necessary, but it interacts with [D1](https://github.com/OPS-PIvers/SpartBoard/issues/2333) and with [the GPL-3 read](https://github.com/OPS-PIvers/SpartBoard/issues/2337) (a TTS preview is a *second* espeak-ng-shaped problem).
- **Is IPA ever shown to the teacher?** Language teachers may want it; EL teachers may not. Showing the expected phoneme string is the most direct way to make the target verifiable, and the most direct way to make the editor intimidating.
- **How does authoring fail loudly?** A target phrase containing a character the G2P tables do not handle should surface at authoring time, not when 30 students have already recorded.
- **Discoverability under the gate.** The feature is behind a `GlobalFeature` with `missingDocPublic: false`, so most teachers never see it. What the gated-in teacher's entry point looks like in the editor.

### Constraints

`CLAUDE.md` governs: container-query units for any widget front-face content, WCAG AA, and the quiz editor's existing patterns. Settings-panel surfaces do not need container-query scaling.

## Blocked by

- [D4 — 'Speak' as a QuizQuestionType, or a separate activity type?](https://github.com/OPS-PIvers/SpartBoard/issues/2334)
- [Dialect on the question type surface — es vs es-419 and beyond](https://github.com/OPS-PIvers/SpartBoard/issues/2338)

There is nothing to author until the question shape and its dialect field exist.

OPS-PIvers:
[...]

PR #2365, opened 2026-08-03
spike(pronunciation): where accepted stress patterns come from (#2360)

OPS-PIvers:
Resolves [#2360](https://github.com/OPS-PIvers/SpartBoard/issues/2360) on the [Wayfinder Map: Multilingual Pronunciation Engine](https://github.com/OPS-PIvers/SpartBoard/issues/2331). Full reasoning in the [resolution comment](https://github.com/OPS-PIvers/SpartBoard/issues/2360#issuecomment-5172338153).

**No feature code.** A spike directory like its four siblings — not imported by the app, but it runs under `pnpm test`, so a later change that violates one of these decisions fails CI.

## What's here

- `scripts/spikes/stress-reference/DECISIONS.md` — five decisions (R1–R5) plus three sub-decisions (R1a, R1b, R2a). **This file is authoritative** for the decisions and the test count; this description deliberately doesn't restate the count, since it drifted three times during review.
- `reference.ts` — reference implementation of the derivation and the grading gate
- `reference.test.ts` — the executable form of every decision. Every espeak string is real output from the runs in `measure/`, never hand-written to make a test pass
- `measure/` — six harnesses behind every number quoted, plus the tuned German variant that shows the ceiling. Data files are downloaded, not committed (~10 MB); `DECISIONS.md` has the commands

## The decisions

| | |
| --- | --- |
| **R1** | Two sources per language where one exists. Disagreement stores the **union** (so #2342 holds unconditionally) **and** raises a flag (~half of disagreements are espeak being wrong, not dialect). Flag rates: en 5.1%, es 0.35%, de 0%. |
| **R1a** | A syllable-**count** mismatch yields no opinion rather than a manufactured disagreement. *Decided without asking, flagged for review.* |
| **R1b** | An espeak `?` placeholder yields **no reference and a flag**, with `syllableCount: null` — unknown, not zero. *Surfaced in review.* |
| **R2** | An unconfirmed flag **degrades to unscored** (A10a's existing path), surfaced **inline on the question**. |
| **R2a** | A confirmation must name at least one syllable, else it silences the prompt permanently while scoring `null` forever. *Surfaced in review.* |
| **R3** | Spanish cross-checks against its own orthography — ~30 lines, no data file, **99.65%** agreement. |
| **R4** | German has **one source**. An explicit, recorded weakening of never-penalize-a-correct-dialect for one of three languages. A fact about available data, deliberately **not** a special case in the code. |
| **R5** | Two primary marks from one source: accepted silently for Spanish `-mente`, flagged for en/de initialisms. |

## Premises that died on contact with the artifacts

1. **The problem is half the size the ticket assumed.** #2336's 8.5% compared *full* patterns including secondary. Primary only: **5.1%**.
2. **"Sources disagree" ≠ "both attested."** CMUDict lists two primary positions for only **0.82%** of headwords, and only **48.8%** of disagreements are readings it lists at all.
3. **wikipron has zero stress marks** in 333,536 entries. The obvious uniform source doesn't contain what this ticket needs.
4. **German stress isn't recoverable from word shape**, and fixing every named failure class made it **worse** (86.24% → 81.15%).
5. **espeak emits literal `?` for unmappable phonemes** — **0.97% of German**, 0 in es/en. Found while measuring a review claim that a related bug was academic.
6. **espeak never abstains** — a primary mark on every polysyllabic word it renders, **0 exceptions in 33,974**. That measurement decided what the no-mark branch should do.

## Also recorded

Syllable indices live in the **model's nucleus space**, not a human's: English `aɪə` and `aɪɚ` are each one token (*lion*, *tired*), Spanish `ue` is two (*luego*, *prueba*). A stress picker must render from the phoneme stream, never spelling — and for R1b words must supply the count itself.

## Verification

- `npx prettier --check` — clean
- Full suite at first push: **592 files, 7,250 tests passed**
- Every number `DECISIONS.md` quotes re-verified after each harness change — all reproduce exactly
- `scripts/` is eslint-ignored, same as the sibling spikes

https://claude.ai/code/session_012jNeZMi2FCH1aq2FzKNiqr

### DIFF
--- a/scripts/spikes/stress-reference/DECISIONS.md
+++ b/scripts/spikes/stress-reference/DECISIONS.md
@@ -7,7 +7,7 @@ a decision ticket on the [Wayfinder Map: Multilingual Pronunciation Engine](http
 (S1–S5) and supplies the reference that lets **A9's stress weight leave 0**.
 
 **Contents.** `reference.ts` is a reference implementation, `reference.test.ts`
-the executable form of every decision below (35 tests), `measure/` the
+the executable form of every decision below (41 tests), `measure/` the
 harnesses that produced every number quoted here. Neither is wired into the
 app nor imported by feature code — a spike directory, like its four
 siblings. It runs under `pnpm test`, so a later change that violates one of
@@ -137,6 +137,53 @@ nuclei.
 
 _Decided without asking_ — it follows from R1 but was not raised in the
 session. Flagged for review, as S3a and S3b were.
+
+#### R1b — An unrenderable espeak symbol yields no reference at all
+
+**espeak writes a literal `?` where one of its internal phonemes has no entry
+in its IPA translation table.** Measured across the same corpora as everything
+else above:
+
+|        | words whose espeak IPA contains `?` |
+| ------ | ----------------------------------- |
+| **de** | **117 / 12,000 — 0.97%**            |
+| es     | 0 / 12,000                          |
+| en     | 0 / 9,974                           |
+
+It is the `UR` phoneme — `durch` is `d'URC` in espeak's own scheme and
+**`dˈ??ç`** in IPA — and it hits a common word class: _wurde, durch, kurz,
+geburt, sturm, urteil, ursache, geburtstag, verurteilt_.
+
+When this happens **both the syllable count and every index derived from it
+are untrustworthy**, so the derivation returns nothing and flags. Dropping
+just the bad index is not enough: `geburt → ɡəbˈ??t` scans as **one** nucleus,
+which would take the monosyllable path and assert stress on syllable 1 of a
+word whose vowel was never rendered.
+
+It **flags** rather than silently degrading, so R2's affordance surfaces it.
+Note this is **the only flag German can ever raise**, since R4 leaves it
+without a cross-check — and it is a German-only defect, which is a second
+independent reason German is the weakest of the three languages here.
+
+> **Does this contaminate R4's 86.24%?** No. Only **1 of the 9,365** scored
+> German words contains a `?`: losing a vowel changes the syllable count, so
+> R1a's count guard already excluded 46 of the 47 that reached scoring range.
+> Maximum swing **0.01pp**, against a 13-point gap. Worth checking rather than
+> assuming — the defect was found _after_ the decision was made.
+
+**Related invariant, same cause:** `parseEspeak` encodes a mark as "the next
+nucleus is primary", so a trailing bare `ˈ` would yield an index one past the
+end. Out-of-range is the worst value available — it can never match a detected
+syllable, so it scores every student 0 while `accepted` stays non-empty and
+therefore never degrades. `confirmReference` already rejected such indices on
+the teacher path; **the derivation path must not be able to produce what
+confirmation forbids**, so out-of-range marks are dropped and a test asserts
+the invariant across every fixture in the suite.
+
+_Surfaced in review of [PR #2365](https://github.com/OPS-PIvers/SpartBoard/pull/2365),
+where the reported risk was a trailing bare mark and was judged academic
+because "espeak never does this." Measuring it found a different, real cause
+producing the same bad value — see the map's rule 4._
 
 ### R2 — An unconfirmed flag degrades, and says so on the item
 
--- a/scripts/spikes/stress-reference/DECISIONS.md
+++ b/scripts/spikes/stress-reference/DECISIONS.md
@@ -7,7 +7,7 @@ a decision ticket on the [Wayfinder Map: Multilingual Pronunciation Engine](http
 (S1–S5) and supplies the reference that lets **A9's stress weight leave 0**.
 
 **Contents.** `reference.ts` is a reference implementation, `reference.test.ts`
-the executable form of every decision below (31 tests), `measure/` the
+the executable form of every decision below (33 tests), `measure/` the
 harnesses that produced every number quoted here. Neither is wired into the
 app nor imported by feature code — a spike directory, like its three
 siblings. It runs under `pnpm test`, so a later change that violates one of
@@ -161,6 +161,26 @@ items altogether).
 [#2341](https://github.com/OPS-PIvers/SpartBoard/issues/2341):** an inline,
 per-question stress affordance showing the syllables and which are accepted —
 not a badge, and not a queue entry.
+
+#### R2a — A confirmation must name at least one syllable
+
+`confirmReference(ref, [])` throws. An empty confirmation sets
+`confirmed: true, flagged: false`, which turns `needsAuthoringPrompt()` off
+and **removes the affordance R2 depends on, permanently** — the question then
+reads as settled while scoring `null` for stress forever, indistinguishable
+from a question nobody flagged. This is A13's failure mode exactly: nobody
+investigates a question that looks finished. It is strictly worse than the
+out-of-range index the reference already rejected, because it does not look
+wrong.
+
+There is deliberately **no "confirm that stress should not be scored here"
+path.** If teachers want one it is an authoring decision for #2341, and it
+needs a state of its own rather than an empty list that mimics absent
+evidence — the same collision S5 closed between an empty `accepted` and an
+empty `detected`.
+
+_Surfaced in review of [PR #2365](https://github.com/OPS-PIvers/SpartBoard/pull/2365)
+as reachable but unguarded._
 
 ### R3 — Spanish cross-checks against its own orthography, with no data file
 
--- a/scripts/spikes/stress-reference/DECISIONS.md
+++ b/scripts/spikes/stress-reference/DECISIONS.md
@@ -7,7 +7,7 @@ a decision ticket on the [Wayfinder Map: Multilingual Pronunciation Engine](http
 (S1–S5) and supplies the reference that lets **A9's stress weight leave 0**.
 
 **Contents.** `reference.ts` is a reference implementation, `reference.test.ts`
-the executable form of every decision below (46 tests), `measure/` the
+the executable form of every decision below (47 tests), `measure/` the
 harnesses that produced every number quoted here. Neither is wired into the
 app nor imported by feature code — a spike directory, like its four
 siblings. It runs under `pnpm test`, so a later change that violates one of
--- a/scripts/spikes/stress-reference/DECISIONS.md
+++ b/scripts/spikes/stress-reference/DECISIONS.md
@@ -7,7 +7,7 @@ a decision ticket on the [Wayfinder Map: Multilingual Pronunciation Engine](http
 (S1–S5) and supplies the reference that lets **A9's stress weight leave 0**.
 
 **Contents.** `reference.ts` is a reference implementation, `reference.test.ts`
-the executable form of every decision below (49 tests), `measure/` the
+the executable form of every decision below (50 tests), `measure/` the
 harnesses that produced every number quoted here. Neither is wired into the
 app nor imported by feature code — a spike directory, like its four
 siblings. It runs under `pnpm test`, so a later change that violates one of
@@ -181,7 +181,13 @@ That is the same category error S5 closed: a value meaning "we do not know"
 must not be spelled as a value that means something else.
 
 Consequently `confirmReference` **requires the true count as an argument** when
-the stored one is `null`. The obligation is in the type rather than left as an
+the stored one is `null` — and **throws if one is supplied when the stored
+count is already known.** A known count is not overridable: it is in the
+model's nucleus space, so "correcting" it toward a human's count would store
+an index the detector can never emit, scoring 0 forever against a reference
+that looks right. `camera → kˈæmɹə` genuinely has two nuclei there. Overriding
+is not a repair, it is a way to make the reference unreachable, so it fails
+loudly at authoring time rather than silently in a grade. The obligation is in the type rather than left as an
 implicit contract, because [#2341](https://github.com/OPS-PIvers/SpartBoard/issues/2341)
 is the code that has to satisfy it: **rescuing an unrenderable word means the
 authoring UI supplying the syllable count from outside this reference.**
--- a/scripts/spikes/stress-reference/DECISIONS.md
+++ b/scripts/spikes/stress-reference/DECISIONS.md
@@ -7,7 +7,7 @@ a decision ticket on the [Wayfinder Map: Multilingual Pronunciation Engine](http
 (S1–S5) and supplies the reference that lets **A9's stress weight leave 0**.
 
 **Contents.** `reference.ts` is a reference implementation, `reference.test.ts`
-the executable form of every decision below (41 tests), `measure/` the
+the executable form of every decision below (44 tests), `measure/` the
 harnesses that produced every number quoted here. Neither is wired into the
 app nor imported by feature code — a spike directory, like its four
 siblings. It runs under `pnpm test`, so a later change that violates one of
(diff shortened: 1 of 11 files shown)


OPS-PIvers:
All three confirmed and fixed in 3f60089.

The dead-code one was the real find, and the cause is worth recording: `german_tuned.py` was generated by a non-greedy regex over `german.py`, and `rule_stress` has **two** `return 1, len(nuc)` statements — one inside the separable-prefix branch, one as the default. The regex stopped at the first, so the original steps 3–4 were stranded past the new function's final return.

Since DECISIONS.md cites that script's output as the basis for R4, the thing that needed checking was not whether the code parses but whether the **numbers still reproduce**. Re-run from `measure/` on the relative path, they do, exactly:

| | before | after |
| --- | --- | --- |
| German morphological rule | 86.24% | **86.24%** |
| German tuned | 81.15% | **81.15%** |
| Spanish orthographic rule | 99.645% | **99.645%** |
| espeak vs CMUDict | 94.9% / 5.1% | **94.9% / 5.1%** |
| …disagreement split | 48.8% / 51.2% | **48.8% / 51.2%** |

So the dead block was genuinely unreachable and the decision record stands unchanged. Duplicate `'ös'`/`'ent'` dropped too — no effect, as you noted, since the scan is longest-first and the strings are identical.

---

OPS-PIvers:
Real gap, fixed in 02e6fb9 and recorded as **R2a**.

I probed it rather than reading the branch, and the behaviour was exactly as described:

```
confirmReference(flagged, [])  →  { confirmed: true, flagged: false, accepted: [] }
effectiveAccepted(c)           →  []
needsAuthoringPrompt(c)        →  false      ← the affordance is gone, permanently
scoreStress(c, 1)              →  null
```

The `needsAuthoringPrompt → false` is the part that makes this worth a decision rather than a guard. R2's entire design is that an unscored stress dimension stays *visible* on the question; an empty confirmation removes the prompt and leaves a question that reads as settled while never scoring stress again. That's A13's failure mode — nobody investigates something that looks finished — and it's why this is worse than the out-of-range case even though it fails "safely."

Now throws `RangeError`. Two tests added (31 → 33): the throw, plus the invariant it exists to protect — **a confirmed reference always has something for the grader to compare against**, so "confirmed" and "unassessable" can't be the same state wearing different labels.

One thing I deliberately did *not* do: add a "confirm that stress should not be scored here" path. It's a plausible teacher intent, but if it's wanted it needs a state of its own — an empty list would mimic absent evidence, which is the exact collision S5 closed between an empty `accepted` and an empty `detected`. Noted in R2a as an authoring decision for #2341.

On the other notes: agreed and no action. The one-subprocess-per-word in `stress_source.py` is deliberate for the 341-word corpus and documented inline; `german_tuned.py` duplicating `german.py` is the point of it, since the two numbers only mean something side by side.

---

OPS-PIvers:
Both correct. Fixed in d44b08b.

**The `aɪɚ` nit.** Verified rather than taken on trust — they are genuinely two vocabulary entries, and the difference is one codepoint that renders almost identically:

```
'aɪə'  in nuclei=True   U+0061 U+026A U+0259 (LATIN SMALL LETTER SCHWA)
'aɪɚ'  in nuclei=True   U+0061 U+026A U+025A (SCHWA WITH HOOK)

lion   'lˈaɪən'  → longest nucleus present: aɪə
tired  'tˈaɪɚd'  → longest nucleus present: aɪɚ
```

Split into two `it` blocks with the codepoints in the comments, so dropping either token fails with an obvious message rather than leaving one word's syllable count unexplained. This is worth more than a comment fix: `ɚ` is one of the four vowels #2359's rule-8 audit found silently missing from the nucleus set, and a test that only pins `aɪə` while using it to explain both words is exactly the shape of assertion that lets a regression through.

**Test count.** PR description updated — it now says 34 and lists R1a/R2a, which it never mentioned.

34 tests, type-check and prettier clean.

---

OPS-PIvers:
All three verified and fixed in 4706531. 35 tests.

**Empty cross-check path.** Reachable, and I went looking for how often before writing the test: **55 polysyllabic CMUDict headwords carry secondary stress but no primary in any variant.**

I used a real one rather than the suggested `weekend` fixture, because the fixture choice decides whether the test actually reaches the branch:

```
accredit   cmu = AH0 K R EH2 D AH0 T   (3 syllables, no primary anywhere)
           espeak = ɐkɹˈɛdɪt           (3 nuclei, primary on 2)
```

The counts **agree**, so R1a's mismatch guard doesn't fire first and the empty-primary branch is genuinely exercised. A fixture whose counts disagreed would have passed for the wrong reason and shadowed the very path being tested.

**Sibling count.** `scripts/spikes/` holds five directories, so four siblings — DECISIONS.md was the wrong one, PR body was right. Fixed.

**`confirmReference` JSDoc.** Fair, and it matters more than a doc nit: R4 leaves German with **no cross-check and therefore no flag**, so an unflagged override is the *only* way a German reference is ever corrected. If a #2341 implementer read the old wording and gated the affordance on `flagged`, German would have no correction path at all. The doc now says so explicitly.

---

OPS-PIvers:
## The bug is real. The stated reason it was academic is not.

Fixed in 8170e0b as **R1b**, and this one changed the decision record rather than just the code.

The note said a trailing bare `ˈ` is the only way to get an out-of-range index, and that it's academic because "espeak never does this." I measured that premise instead of accepting it — map rule 4 — and found **a different cause producing the same bad value, in real corpus data**:

**espeak writes a literal `?` where one of its internal phonemes has no entry in its IPA translation table.**

```
durch    -x: d'URC       --ipa=3: dˈ??ç
geburt   -x: g@b'URt     --ipa=3: ɡəbˈ??t
```

| | words whose espeak IPA contains `?` |
| --- | --- |
| **de** | **117 / 12,000 — 0.97%** |
| es | 0 / 12,000 |
| en | 0 / 9,974 |

It's the `UR` phoneme, and it hits a common word class: *wurde, durch, kurz, geburt, sturm, urteil, ursache, geburtstag, verurteilt*. Trailing bare marks: **0** across all 34k words. So the guard was needed for a reason neither of us had.

**And the suggested one-line filter is not sufficient on its own.** `geburt → ɡəbˈ??t` scans as **one** nucleus, so filtering the bad index would drop it into the monosyllable path and assert stress on syllable 1 of a word whose vowel espeak never rendered — trading a loud wrong answer for a quiet one. R1b therefore returns no reference at all and **flags**, which for German is the only flag that can ever be raised, since R4 leaves it without a cross-check.

Your filter is kept as a defensive invariant, with a test asserting **no derived reference contains an out-of-range index** across every fixture in the suite. That's the right framing for it: `confirmReference` already forbade out-of-range indices on the teacher path, so derivation must not be able to produce what confirmation rejects.

**Does this contaminate R4's 86.24%?** No — checked, not assumed, since the defect was found *after* the decision. Only **1 of the 9,365** scored German words contains a `?`: losing a vowel changes the syllable count, so R1a's count guard had already excluded 46 of the 47 that reached scoring range. Maximum swing **0.01pp** against a 13-point gap.

41 tests, type-check and prettier clean. Also worth flagging for the map: this is a concrete instance of the "server-side symbol normalization" fog, and a nastier one than that fog anticipated — it assumed espeak emits *valid IPA outside the model's set*, not *invalid placeholder characters*.

---

OPS-PIvers:
Fixed in ffbcca7. The first note was right, and probing it showed the problem is **worse than reported** — and you were right that my own test documented the workaround instead of the contract, which is the part I should have caught.

The count isn't reliably `0`. It's reliably **wrong**:

```
durch   -> dˈ??ç      syllableCount=0   confirmReference(r,[1]) => RangeError
geburt  -> ɡəbˈ??t    syllableCount=1   confirmReference(r,[1]) => OK
```

`geburt` is **ge-burt** — two syllables, stress on the second. Storing `1` means a teacher who correctly wants syllable 2 is rejected as out of range, **against a reference that looks perfectly valid**. The zero case at least fails loudly; this one fails plausibly. Same category error S5 closed: a value meaning "we do not know" spelled as a value that means something else.

So rather than documenting the workaround, `syllableCount` is now **`number | null`** and `confirmReference` takes the true count as an argument when the stored one is `null`:

```ts
const r = deriveReference({ lang: 'de', espeakIpa: 'ɡəbˈ??t' });
confirmReference(r, [2]);      // RangeError — count unknown
confirmReference(r, [2], 2);   // OK, and bounds-checked against the supplied 2
```

That puts the obligation in the type instead of an implicit contract, which was your actual point. #2341 is the code that has to satisfy it, and it's now stated there: **rescuing an unrenderable word means the authoring UI supplying the count from outside the reference.** Recorded in R1b.

**`SUF_NEUTRAL`** — correct, defined and never referenced in either harness. Removed. There was no dropped guard: the intent is already covered by `SUF_STRESS` being an allowlist rather than the rule being a denylist, so none of those suffixes could attract stress anyway. German numbers re-run after removal and unchanged — **86.24%** and **81.15%**.

**Test count** — `DECISIONS.md` was the accurate one; `grep -c "  it("` and the vitest runner both agreed at 41 before this change, so nothing was higher than either. The PR body was the stale one and is now updated. Both now say **44**.

44 tests, type-check and prettier clean.

---

OPS-PIvers:
Fix applied in 22bc45f. The gap is real — the stated impact isn't. Probed rather than reasoned about:

```
stored       {"syllableCount":2,"accepted":[2,3],"flagged":true,"source":"disagreement"}
effective    []
score(3)     null
confirm[2,3] RangeError
```

An out-of-range cross-check index can only enter via the **disagreement** path, and that path always sets `flagged: true` — so `effectiveAccepted` returns `[]` and the student scores `null`, not 100. Confirming it throws. So "would score a student 100" doesn't happen on any reachable path.

**What it does break is my own invariant test.** The fixture sweep I added last commit asserts no derived reference contains an out-of-range index — but it only ran fixtures with **no cross-check**, so this walked straight past it. That's a hole in the test, not just the code, and the more useful half of this note.

Fixed both:

- Out-of-range cross-check indices are dropped. And since a **malformed cross-check is not evidence of a second reading**, an empty result routes to R1a's no-opinion path rather than raising a flag — flagging would ask a teacher to adjudicate a disagreement that never existed.
- The sweep now runs every fixture against **no cross-check, an agreeing one, a disagreeing one, an empty one, and two malformed ones**.

Checked that the test earns its place rather than just restating the fix — reverting the one-liner fails **3 tests including the sweep**:

```
× drops an out-of-range CROSS-CHECK index instead of storing it
× treats a wholly out-of-range cross-check as no opinion, not disagreement
× holds across every fixture in this suite
Tests  3 failed | 43 passed (46)
```

46 tests, type-check and prettier clean.

Agreed on your framing that this is about promotion to feature code rather than a live defect — `deriveReference` will be called with caller-supplied data, and it should enforce the same bounds invariant `confirmReference` already does. Recorded alongside R1b.

---

OPS-PIvers:
All three addressed in 0c37fb3. You framed the first as "either interpretation is defensible" — measuring it made one of them clearly right.

**espeak emits a primary mark on every polysyllabic word it renders: 0 exceptions in 33,974 words** across es, de and en.

```
es (n=12000)  0 primary marks: 0 (0.00%)
de (n=12000)  0 primary marks: 0 (0.00%)
en (n= 9974)  0 primary marks: 0 (0.00%)
```

That settles it. Reaching that branch means **the input is no longer espeak output** — something upstream (the normalization step, a truncation) has mangled it. And once that's true, the syllable *count* is exactly as suspect as the missing mark. A cross-check index is only meaningful against a count you can trust, so deferring to it would assert a position in an index space we can no longer validate. That's R1b's reasoning, arriving by a different route.

So: **degrade regardless is intentional**, now named in a comment on the branch and pinned by a test that supplies a cross-check *with* a reading — the combination that was untested. It's a guard, not a live path, and the comment says so.

**Both Python notes correct**, both behaviour-neutral, both re-verified rather than assumed:

- `stress_source.py` opened `corpus.txt` without `encoding='utf-8'` while every other `open` in `measure/` passes it. Re-run: **97.4% / 96.2% / 3.8%** unchanged.
- `german_tuned.py` appended seven suffixes already in `SUF_STRESS` — `ismus age ade ur eur ent abel`. Only `ik` and `ee` were new, exactly as you read it. Re-run: **81.15%** unchanged.

47 tests, type-check and prettier clean.

---

A note on where this is, since this is the seventh round: the five decisions have been stable since the first review, and every round since has hardened the reference implementation against callers that don't exist yet — correct for CMUDict and the Spanish rule, defensive against hypothetical feature-code input. That's real promotion-readiness work and I've taken all of it, but it's well past what #2360 needed to answer. Unless something reaches an actual decision, I'd suggest merging on green.

---

OPS-PIvers:
9558648. **Declining the code change on the first one**, taking the documentation half. First note I've turned down across eight rounds, so here's the reasoning.

`if (!crossCheck || lang === 'de')` encodes a fact about **available data** as a **constraint in the function**. R4 says no usable second source exists for German *today* — and this file's own R4 section names re-scraping Wiktionary with stress preserved as the move if German stress proves a real classroom problem. At that point, passing a German cross-check is exactly the right call, and the guard would **silently discard it while appearing to work**. That's a worse failure than the one it prevents, and it would have to be ripped out to do the very thing R4 names as its own next step.

The prevented behaviour also isn't a misgrade. A flagged German word **degrades until confirmed** — the safe direction. Nothing scores wrong; a teacher gets asked a question they'd otherwise never be asked.

So the guarantee is now stated where it actually lives — *R4 holds today because there is nothing to pass, not because the function refuses it* — and **both halves are pinned**, including the one your suggested test would have asserted the opposite of:

```ts
it('is NOT special-cased: a German cross-check, if one ever exists, is used')
it('flags nothing today, because nothing is ever passed')
```

Happy to be overruled by Paul if he wants the guard — it's a one-token change either way.

**Test count** — you're right that `DECISIONS.md` is the file that ages into the codebase, so I've fixed the *class* rather than the instance: the PR description no longer states a count at all and points at `DECISIONS.md` instead. It had drifted three times (31 → 34 → 44) because I update the body and then reviews add tests. Now there's one place to be wrong.

49 tests, type-check and prettier clean.

---

OPS-PIvers:
Genuine finding — fixed in d472fbf. You offered two resolutions; probing it shows one of them would have introduced a worse bug than the one it fixes.

```
confirm([1], 3)  ->  silently kept syllableCount=2      the reported bug
confirm([3], 3)  ->  RangeError                          correct — and LOST if precedence flips
```

**Flipping to `syllableCount ?? ref.syllableCount` would be actively wrong.** The count is in the **model's nucleus space**, which is the space the detector ranks in — not a teacher's. `camera` really is `kˈæmɹə`, **two** nuclei. A caller "correcting" it toward a human's three isn't repairing anything; they're storing an index the detector can never emit, which scores 0 forever against a reference that looks perfectly right. That's the exact failure the index-space section of `DECISIONS.md` exists to warn about, and today's bounds check already catches it — flipping precedence would remove the catch.

So: **overriding a known count is not a repair, it is a way to make the reference unreachable.** It now throws, loudly at authoring time rather than silently in a grade. It throws even when the supplied count *agrees*, since accepting it would imply the parameter is meaningful there. Docstring covers the non-null case, as you asked.

Both minor notes taken:

- **`UNRENDERABLE_MARK`** now records that `?` is an espeak sentinel and **not a valid IPA glyph** — which is the actual reason substring-matching it is safe rather than a heuristic. Good catch that the constant didn't say why.
- **Sweep narrowness** — you're right, and right that it isn't a coverage gap. I've made the test say what it is: *a bounds invariant, not decision coverage; monosyllabic fixtures short-circuit before R1a, which the `camera` case covers instead.* A sweep that wide looks like it proves more than it does, and that's worth writing down.

50 tests, type-check and prettier clean.

---

OPS-PIvers:
Thanks — no action taken this round, and recording why so it's clear they were considered rather than missed.

- **One subprocess per word in `stress_source.py`** — agreed, and agreed on the trigger to revisit. It's deliberate for the 341-word K-12 corpus and documented inline; the batched path is what the 10k-word runs use.
- **Bare `open()` without context managers** — declining. These are Python, `scripts/` is eslint-ignored, and nothing lints them; the warning scenario requires a linted Python pipeline that doesn't exist. Touching all five harnesses for zero behavioural gain would mean re-verifying every number in `DECISIONS.md` again to prove I hadn't disturbed one, which is real risk against no benefit.
- **`lang` only consumed in the R5 branch** — correct read, and thank you for checking the R4 comment rather than flagging it as a redundant parameter. That's exactly the reasoning.

This is the ninth review round. The five decisions have been stable since the first, and the last several rounds have hardened the reference implementation against callers that don't exist yet — genuinely useful for #2341 copying the shape, and past what #2360 needed to answer. Unless something reaches an actual decision, this is ready to merge on green.

---

PR #2381, opened 2026-08-04
feat(schedule): show the current event's full title instead of truncating

OPS-PIvers:
## Problem

Long event titles were clipped with an ellipsis on every row of the Schedule widget. The only way to reveal more text was to resize the widget — which also rescaled the content, because `scheduleSize()`'s `cqw` term binds whenever the widget sits in the narrow band (content width between ~162px and 0.45 × height).

The obvious fixes each break something:

- **Shrink text to fit** → one long title shrinks every title.
- **Wrap every row** → rows become uneven in height, or uneven in text size.

## Approach

Only the **focused** row renders its title in full, wrapped. Font size is identical on every row — only the focused row's height differs, which reads as deliberate emphasis alongside the border and "Now" badge it already carries.

Focus is a **new display-only index**, kept deliberately separate from `activeIndex`:

1. The time-based active item, when it is not already checked off
2. Otherwise the first not-done item — covering both "teacher finished early and checked it off" and "no event is active yet" (before the first event, or in a gap)
3. `-1` when every item is done or the schedule is empty

`activeIndex` keeps its exact previous meaning, so the "Now" badge, auto-progress, and linked-widget auto-launch are untouched. That separation is the main regression-avoidance decision in this PR.

## Changes

- **`Schedule/utils.ts`** — new pure `computeFocusIndex()`, plus the `EXPANDED_ROW_MAX_HEIGHT_CQH` constant
- **`Schedule/components/ScheduleRow.tsx`** — new `isExpanded` prop swapping `truncate` for wrapping; capped at `max-height: 40cqh` so a pathological title scrolls inside the row instead of hiding every other event. Right padding is reserved on the active row **only while it is expanded**, so a wrapped first line cannot run under the "Now" badge — and rows render pixel-identically to before when the feature is off.
- **`Schedule/ScheduleWidget.tsx`** — derives `focusIndex`; auto-scroll now anchors on the focused row, still gated on the existing opt-in `autoScroll` setting
- **`Schedule/Settings.tsx`** — `Expand Current Event` toggle, default on
- **`types.ts`** — new `expandActiveItem?: boolean` on `ScheduleConfig`

`isExpanded` is wired into `ScheduleRow`'s hand-written `React.memo` comparator. That comparator enumerates props explicitly and returns `true` by default, so omitting it there would have silently stopped rows from re-rendering when focus moves.

**Intentionally unchanged:** title sizing (`scheduleSize(10, 36)`) and row chrome (check circle, gap, padding, timer button). Removing the `cqw` term would roughly *double* title size on tall-narrow widgets and make truncation worse, so it stays. Existing dashboards render identically apart from the focused row.

## Verification

- Full unit suite: **588 files, 7143 tests passed**
- `tsc --noEmit` clean; `pnpm run lint` exit 0; Prettier clean
- 8 new unit tests for `computeFocusIndex` (active / checked-off / nothing-active / all-done / empty / out-of-range index / missing `done` flag)
- 6 new widget tests covering wrapping vs. truncation, focus advancing on check-off, the pre-first-event fallback, the all-done case, the height cap, and the toggle off
- Measured in real Chromium at 230×400 using the actual `scheduleSize()` expressions: title font stays **18px on every row** in both states, the active row wraps to 2 lines and is no longer clipped, other rows still truncate, and `max-height` resolves to **160px** (40% of 400px)

### Known test-coverage gap

Font-size parity between expanded and truncated rows — the core promise of this feature — is **not** asserted in the jsdom suite. jsdom's CSS parser drops `clamp()`/`min()` values entirely, so every `style.fontSize` reads back as an empty string and such an assertion would pass vacuously. I removed the assertion rather than leave a fake-green test, and documented why in `ScheduleWidget.test.tsx`. Parity is guaranteed structurally instead (both branches read the same single `scheduleSize()` expression, which `isExpanded` does not participate in) and confirmed by the browser measurement above. The `paddingRight` reservation is untestable in jsdom for the same reason.

---
_Generated by [Claude Code](https://claude.ai/code/session_017x5iU1xdCaQvADDdZ36XtD)_


OPS-PIvers:
All 10 checks green. Thanks for the review — keeping the code as-is on all three points, with reasoning on the one that proposed a change.

**`overflow-y: auto` vs `hidden` on the capped title** — keeping `auto`. SpartBoard's schedule widget is used touch-first (that constraint is what ruled out a hover-to-reveal tooltip during design), so touch-scrolling is exactly the recovery path this needs, not an incidental side effect. `hidden` would clip the overflow with *no* way to read the rest of the title on any input method, which is strictly worse than a scroll region that's merely undiscoverable on a mouse. The cap only engages past 40cqh, so in practice this is a safety net for pathological titles rather than a path teachers hit — and when they do hit it, the honest failure mode is "you can scroll it" rather than "the text is gone."

**Auto-scroll shift when the active item is checked off** — correct reading, and intended. The expanded row is the one a teacher is actually reading, so anchoring the scroll on it (rather than on a row they just completed) is the point of the change. It only diverges from the old behavior once `activeIndex` and `focusIndex` disagree, which requires an explicit check-off; the untouched path is byte-identical.

**`widgetDefaults.ts`** — agreed, no action. The `= true` destructuring default is the same pattern `autoProgress`/`autoScroll`/`cardOpacity` already use, and adding it to the defaults object would change how existing saved boards hydrate for no benefit.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-05

### Summary
Adds an opt-in "Expand Current Event" toggle (default on) to the Schedule widget so the focused event renders its full title wrapped over multiple lines while every other row keeps its single-line ellipsis — font size stays identical across rows, only the focused row's height grows. Introduces a display-only `computeFocusIndex` util, deliberately decoupled from `activeIndex` (which still drives the "Now" badge, auto-progress, and linked-widget launch). Thorough unit + component tests.

### Code Quality
✓ Type-safe throughout. `ScheduleConfig` gains `expandActiveItem?`, the `ScheduleRow` memo comparator adds `isExpanded`, and auto-scroll now anchors on `focusIndex` (falling back to `activeIndex`) with a correctly updated dependency array. The focus/active separation is well-reasoned and documented.

### Widget Standards
✓ Font-size parity — the core promise — is preserved structurally: both the expanded and truncated branches read the same `scheduleSize(10 * textScale, 36 * textScale)` expression, which `isExpanded` doesn't participate in. The expanded row is capped at `40cqh` (`EXPANDED_ROW_MAX_HEIGHT_CQH`) with internal scroll so a pathological title can't push other events off-screen, and `paddingRight` reserves badge space only when a row is both active *and* expanded — keeping non-expanded rows pixel-identical. No hardcoded text-size classes introduced in front-face content.

### Admin Config Standards
✓ Not applicable — this is a per-widget user-level display preference, correctly placed in the widget's own Settings panel rather than admin config.

### Firestore & Security
✓ No issues found.

### Regression Risk
`types.ts` is touched, but only to extend `ScheduleConfig` — no new `WidgetType`, so the `WidgetConfig` union / `ConfigForWidget` helper are unaffected. The auto-scroll effect's anchor change preserves the original behavior when expansion is off or all items are done. Low risk.

### Missing Tests
`computeFocusIndex` is exhaustively unit-tested (active-not-done, checked-off fall-through, pre-day, all-done, empty, stale out-of-range index, missing `done` flag). Component tests cover wrap-vs-truncate, height cap, focus hand-off on check-off, pre-day expansion, all-done, and the toggle-off path. Font-size parity and `paddingRight` are documented as un-assertable in jsdom (it drops `clamp()`/`min()`); the structural guarantee is a reasonable substitute. Adequate.

### Overall
**Merge readiness:** Ready
Well-scoped, well-tested feature with careful attention to the scaling model and to not disturbing the existing `activeIndex`-driven behaviors.

---

---

PR #2392, opened 2026-08-05
fix(dashboard): DashboardView defers global Escape to an open Modal

OPS-PIvers:
## Update (post-review redesign)

The original approach in this PR (making `Modal.tsx` register its Escape handler in the capture phase + `stopPropagation()`, to out-race `DashboardView`'s global fallback) was reverted after an independent investigation — triggered by an automated review's regression-risk flag — found two concrete, live-reachable regressions. See the full writeup in the review thread below and the latest commit message. Summary:

1. **Nested confirm/prompt/alert dialogs** (`useDialog()`, rendered via `DialogContainer.tsx`) opened from inside a `Modal` (e.g. `PlcEditModal`'s "Remove member?" confirm) would *also* close the parent Modal on Escape — a same-phase (`window`, capture), same-target registration-order race: Modal's listener (registered when the Modal opened) always fires before the dialog's own capture listener (registered later, when the confirm was triggered), because plain `stopPropagation()` doesn't block sibling listeners on the same node — only `stopImmediatePropagation()` does.
2. **In-modal controls with their own Escape handling** not recognized by `isEscapeFromWidgetInput()` — e.g. `RestrictionsPicker.tsx`'s per-row popover (used inside `RosterEditorModal`, itself inside a `Modal`) — would never receive the Escape keydown at all, since a capture-phase `stopPropagation()` on `window` fires *before* the event ever reaches down to the target/bubble phases. Pressing Escape to close the popover would close the whole modal instead.

## Current fix

`Modal.tsx`, `utils/domHelpers.ts`, and `Modal.test.tsx` are reverted to their `dev-paul` state — the capture-phase escalation is fully undone. The original bug (Escape dismissing a Modal also minimized the topmost widget behind it) is instead fixed at its actual source: `DashboardView`'s global Escape handler now bails via `useHasOpenModal()` (Modal's own open-count tracker in `components/common/modalStore.ts`, already built for exactly this "suppress myself while any modal is on screen" purpose) before touching any widget.

This has **zero blast radius on nested modal content** — Modal, confirm/prompt/alert dialogs, and in-modal popovers all keep behaving exactly as they did before this PR ever existed. `DashboardView` simply steps out of the way whenever any Modal is open, rather than Modal trying to out-race `DashboardView` for event-phase priority.

## Root cause (unchanged from original)

`DashboardView.tsx`'s global Escape handler and `Modal.tsx`'s (pre-this-fix) Escape handler were both plain bubble-phase `window` `keydown` listeners. `DashboardView`'s mounts once at teacher-app start (always registered first); `Modal`'s mounts later, whenever a modal opens. Same-phase-same-target listeners fire in registration order, so `DashboardView`'s fallback always ran *before* Modal's own handler had a chance to call `onClose` — meaning every Escape-to-dismiss on any of Modal's 50+ call sites also minimized the topmost widget behind it.

## Regression test

`tests/components/layout/DashboardView.test.tsx` — new describe block `"global Escape defers to an open Modal"`, two tests: (a) no `widget-keyboard-action` dispatch while `useHasOpenModal()` is true, (b) dispatch still happens normally when no modal is open.

- **Fail-before** (verified by temporarily reverting only `DashboardView.tsx`): 27/28 tests pass, 1 fails (`expected "vi.fn()" to not be called at all, but actually been called 1 times`).
- **Pass-after**: 28/28.

## Evidence

- `pnpm run validate`: type-check, lint, format-check, full root suite (597 test files / 7284 tests), full functions suite (40 test files / 778 tests), `test:counts` guard — all green.
- `pnpm run build`: client + SSR bundle build succeeded, prerender of `/privacy`, `/terms`, `/support` succeeded.

## Risk

**contained** — the change is now scoped to `DashboardView`'s own fallback handler checking a pre-existing, purpose-built shared counter; it does not touch `Modal.tsx`'s event-handling model at all, eliminating the wide blast radius the original approach had.

### DIFF
--- a/components/common/Modal.tsx
+++ b/components/common/Modal.tsx
@@ -21,7 +21,6 @@ interface ModalProps {
   className?: string; // For additional styling on the content container
   contentClassName?: string; // For additional styling on the body/content wrapper
   footerClassName?: string; // For additional styling on the footer wrapper
-  captureEscape?: boolean; // Whether to use capture phase for Escape key
   ariaLabel?: string;
   ariaLabelledby?: string;
 }
@@ -39,7 +38,6 @@ export const Modal: React.FC<ModalProps> = ({
   contentClassName = 'px-6',
   footerClassName = 'p-6 pt-4 mt-auto shrink-0 border-t border-slate-100',
   variant = 'default',
-  captureEscape = false,
   ariaLabel,
   ariaLabelledby,
 }) => {
@@ -63,9 +61,26 @@ export const Modal: React.FC<ModalProps> = ({
     const handleEscape = (e: KeyboardEvent) => {
       if (e.key !== 'Escape') return;
       // Guard first: if focus is inside a widget portal, let the portal's
-      // own handler run (don't kill it with stopImmediatePropagation).
+      // own handler run (don't touch propagation — that Escape belongs to
+      // the widget, not this modal).
       if (isEscapeFromWidgetInput(e)) return;
-      if (captureEscape) e.stopImmediatePropagation();
+      // Capture phase + stopPropagation (not stopImmediatePropagation) is
+      // required here, unconditionally — not opt-in. Modal is portalled
+      // straight onto document.body via createPortal, so its content is
+      // never a descendant of `.widget`. DashboardView's global Escape
+      // handler (a window, bubble-phase listener mounted at teacher-app
+      // start, well before any Modal opens) falls back to acting on the
+      // dashboard's topmost widget whenever it can't resolve a `.widget`
+      // ancestor for the focused element — exactly the case here. Without
+      // capturing and stopping the event first, dismissing a Modal with
+      // Escape also silently minimizes an unrelated widget (e.g. a running
+      // Timer) behind it. stopPropagation (rather than
+      // stopImmediatePropagation) still lets a second, already-open Modal's
+      // own capture listener on window fire, preserving nested-modal
+      // Escape behavior. Mirrors the established pattern already used by
+      // DraggableWindow's maximized-menu Escape handler and
+      // DialogContainer's Alert/Confirm/Prompt dialogs.
+      e.stopPropagation();
       // Read from ref so we always call the current onClose even though
       // onClose is not in the effect deps array.
       onCloseRef.current();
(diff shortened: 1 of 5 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-05

### Summary
Removes the dead `captureEscape` prop from the shared `Modal` primitive (no production caller ever set it — only its own test did) and makes `Modal` always register its Escape handler in the **capture phase** with plain `stopPropagation()`, so a modal's Escape-to-dismiss no longer bubbles to `DashboardView`'s window-level widget-minimize fallback. The `domHelpers` doc comment and the test are updated to match. Net dead-code removal plus a real bug fix.

### Code Quality
✓ Clean removal of the dead prop and its two `? :` branches; effect deps correctly simplified to `[isOpen]`. Using `stopPropagation()` rather than `stopImmediatePropagation()` is the right call and is well-justified in the comment (preserves the existing stacked-Modal behavior where each open Modal responds to one Escape).

### Widget Standards
✓ Not applicable (shared common component, no widget front-face).

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found.

### Regression Risk
This is the **widest-blast-radius change in tonight's batch** — `Modal` backs 50+ call sites. Converting every modal from bubble-phase to capture-phase Escape has one behavior implication worth a human's eyes before merge: any control rendered *inside* a modal that relies on seeing Escape first via a bubble-phase handler — e.g. a nested dropdown / combobox / autocomplete that closes itself on Escape (via `stopPropagation`) *without* closing the modal — will no longer receive the event. The window capture handler now fires before the event ever reaches the inner control and closes the whole modal instead. `isEscapeFromWidgetInput()` only exempts widget-portal inputs, not ordinary in-modal controls. Under the previous bubble-phase default the inner control fired first and could keep the modal open.

Recommend a quick confirmation that no in-modal control across those call sites depends on that ordering (searchable selects / comboboxes are the usual suspects).

### Missing Tests
The new test proves Escape no longer reaches a pre-existing global `window` keydown listener — good coverage for the fix itself. The potential inner-control interaction above is not covered; a test rendering an in-modal control that `stopPropagation`s Escape would lock in the intended behavior either way.

### Overall
The fix is correct and removes dead code; given the 50+ call sites, a human should confirm the bubble→capture switch doesn't regress any in-modal Escape-handling control.

---

---

OPS-PIvers:
Following up on the automated review's "widest-blast-radius change" note about in-modal controls relying on bubble-phase Escape ordering: investigated this directly rather than dismissing it, and found it was a real, confirmed problem, not just theoretical.

Two concrete regressions from the capture-phase approach:
1. `useDialog()` confirm/prompt/alert dialogs opened from inside a Modal (e.g. `PlcEditModal`'s "Remove member?" confirm) would also close the parent Modal on Escape, due to a same-phase/same-target registration-order race on `window`.
2. `RestrictionsPicker.tsx`'s in-modal popover (inside `RosterEditorModal`) would never receive the Escape keydown at all — the capture-phase `stopPropagation()` blocks it before it ever reaches down to the popover's own bubble-phase handler.

Redesigned the fix in `321dbbd2`: reverted `Modal.tsx`/`domHelpers.ts`/`Modal.test.tsx` entirely back to `dev-paul`, and instead fixed the original bug at its source — `DashboardView`'s global Escape handler now bails via `useHasOpenModal()` whenever any Modal is open, before touching any widget. Zero blast radius on nested modal content since Modal's own event handling is completely untouched. PR description and title updated to match. Fail-before/pass-after and full validate/build re-verified — see updated description for details.

---

OPS-PIvers:
Good catch on the group-build-mode ordering — agreed it's pre-existing (the group-build-mode branch's own `return` happens before my new guard is even reached, same as it would have before this PR) and out of scope here. Noted in the debugger backlog rather than fixed, since I couldn't confirm the two states are ever actually reachable simultaneously in practice.

---

PR #2396, opened 2026-08-05
fix(drawing): rehydrate page objects before PNG/PDF export

OPS-PIvers:
## Summary

Drawing widgets that have completed the objects-subcollection migration exported **background-only** PNGs and PDFs — every stroke, shape, text object and image was silently dropped from the saved file.

`migrateDrawingToSubcollection` empties `pages[].objects` on the dashboard document by design (`utils/migrateDrawingToSubcollection.ts:183-187`): post-migration `pages[]` is a denormalized `{ id, background }` cache, and the objects live at `/users/{uid}/dashboards/{id}/drawings/{widgetId}/pages/{pageId}/objects`. But all three export handlers passed page data sourced straight from `widget.config`:

| Export | Handler | Was passed | Result post-migration |
| --- | --- | --- | --- |
| PNG (this page) | `handleExportCurrentPng` | `activePage` | background only |
| PNG (all pages) | `handleExportAllPng` | `pages` | background only |
| PDF | `handleExportPdf` | `pages` | background only |

`exportCanvas.renderPageToPng` painted the background template via `paintBackground`, then iterated an empty `page.objects`. The widget's live `objects` (read from the subcollection by `useDrawingObjectsDoc`) was never spliced back in.

`PageStrip` already draws this distinction for its thumbnails — it takes `activePageObjects` for the live page and renders a placeholder for non-active migrated pages — so the export handlers were the remaining gap, not a deliberate choice.

## Changes

- **`utils/hydrateDrawingPages.ts`** (new) — `hydrateDrawingPagesFromSubcollection()` reads `objects[]` back for a set of pages from the page-nested subcollection. The caller's live page is served from its in-memory `objects` instead of a redundant read, so the export matches what's on screen even before a just-drawn stroke round-trips through the listener. Objects are sorted ascending by `z` to match the live renderer.
- **`components/widgets/DrawingWidget/Widget.tsx`**
  - Current-page PNG now exports `{ ...activePage, objects }`.
  - All-pages PNG and PDF go through a new `resolveExportPages()` callback. `useDrawingObjectsDoc` is page-scoped (only the visible page has a listener), so pages other than the active one are read from Firestore.
  - A migrated widget with no addressable subcollection path (missing `uid` or `dashboardId`) now throws rather than silently emitting blank pages. The error message deliberately avoids the word "blocked" so the PDF handler's pop-up-blocker branch doesn't misclassify it.
  - Pre-migration configs continue reading from the dashboard doc; behaviour there is unchanged.

## Testing

- `tests/components/widgets/DrawingWidget/exportSubcollection.test.tsx` (new) — drives the real export popover and asserts the pages reaching the export pipeline carry their objects: active page from the live subscription, non-active page rehydrated from Firestore. Verified as a real regression gate: all three migrated-widget cases fail against the pre-fix `Widget.tsx` (objects arrive as `[]`), and pass with it. A fourth case pins the pre-migration path.
- `tests/utils/hydrateDrawingPages.test.ts` (new) — covers path construction, `z` ordering, live-page reuse (asserts the non-live page is the only read, and that the live array is copied rather than aliased since the exporter sorts), and the empty-page-list case.
- `pnpm run validate` passes clean (type-check root + functions, lint, format-check, root tests, functions tests, test-count guard).
- No runtime browser verification: reaching a migrated widget requires real Google Sign-In plus live Firestore, and `VITE_AUTH_BYPASS` short-circuits the Firestore listeners the bug depends on. The failing-then-passing test reproduction stands in for it.

## Related data-loss finding — not fixed here

The task also asked whether the same stripping breaks dashboard JSON export/import. Findings:

- The Sidebar JSON export/import described in `CLAUDE.md` **does not exist in the codebase** — that documentation is stale. No dashboard-to-file download or `accept=".json"` import UI remains.
- The equivalent paths that *do* exist have the same defect, and each loses drawing content for migrated widgets because they copy only the dashboard document:
  - **Google Drive `.spart` sync** — `googleDriveService.exportDashboard` does `JSON.stringify(dashboard)`, so the Drive backup stores empty `pages[].objects`.
  - **`duplicateDashboard`** (`context/DashboardContext.tsx:3654`) — `structuredClone` of the dashboard doc; the `/drawings/...` subcollections are never copied, so a duplicated board's whiteboards come up empty.
  - **Synced/shared boards** — already documented as a known limitation in `useDrawingObjectsDoc.ts:39-53`.

These are genuine data-loss bugs but need a different fix (hydrating subcollections at write time, plus re-writing them on import/duplicate) and touch the sync pipeline, so they're deliberately left out of this PR rather than bundled in.

---
_Generated by [Claude Code](https://claude.ai/code/session_015yPzPeEDFYu1zq89Z8w4Jd)_

### DIFF
--- a/components/widgets/DrawingWidget/Widget.tsx
+++ b/components/widgets/DrawingWidget/Widget.tsx
@@ -736,12 +736,11 @@ export const DrawingWidget: React.FC<{
   // page has no live subscription, so we read it from Firestore here.
   const resolveExportPages = useCallback(async (): Promise<DrawingPage[]> => {
     if (!config.subcollectionMigrated) {
-      // Pre-migration the dashboard doc is still the source of truth, but the
-      // active page's array can lag `objects` by a render, so prefer the live
-      // slice for it and take the rest verbatim.
-      return pages.map((p) =>
-        p.id === activePage.id ? { ...p, objects } : { ...p }
-      );
+      // Pre-migration the dashboard doc is still the source of truth and
+      // `objects` IS `activePage.objects` (see the assignment above), so
+      // every page already carries its content. `renderPageToPng` copies
+      // before sorting, so handing it these arrays directly is safe.
+      return pages;
     }
     const uid = user?.uid;
     if (!uid || !dashboardId) {
--- a/components/widgets/DrawingWidget/Widget.tsx
+++ b/components/widgets/DrawingWidget/Widget.tsx
@@ -726,14 +726,7 @@ export const DrawingWidget: React.FC<{
     return `Whiteboard-${timestamp}`;
   };
 
-  // Pages to hand the export pipeline, with `objects[]` guaranteed populated.
-  //
-  // Post-migration (PR 2.6) the dashboard doc's `pages[].objects` is emptied
-  // on purpose — it's a denormalized `{ id, background }` cache, and the real
-  // objects live in the page-nested subcollection. Handing `pages` straight
-  // to the exporter would therefore paint the background template and nothing
-  // else. The active page is covered by the live `objects` array; every OTHER
-  // page has no live subscription, so we read it from Firestore here.
+  // Post-migration pages[].objects is empty; the active page uses live objects, every other page is read from Firestore.
   const resolveExportPages = useCallback(async (): Promise<DrawingPage[]> => {
     if (!config.subcollectionMigrated) {
       // Pre-migration the dashboard doc is still the source of truth and
--- a/components/widgets/DrawingWidget/Widget.tsx
+++ b/components/widgets/DrawingWidget/Widget.tsx
@@ -51,6 +51,7 @@ import { DRAWING_DEFAULTS } from './constants';
 import { useDrawingCanvas } from './useDrawingCanvas';
 import { migrateDrawingConfig, nextZ } from '@/utils/migrateDrawingConfig';
 import { deleteDrawingPageSubcollection } from '@/utils/deleteDrawingPageSubcollection';
+import { hydrateDrawingPagesFromSubcollection } from '@/utils/hydrateDrawingPages';
 import { db } from '@/config/firebase';
 import { logError } from '@/utils/logError';
 import type { DrawingPage } from '@/types';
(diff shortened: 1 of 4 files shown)


utils/hydrateDrawingPages.ts
```diff
@@ -0,0 +1,75 @@
...
+ * The active page is served from the caller's live `objects[]` rather than a
+ * fresh read: that array is the same state the canvas is painting, so the
+ * export matches what the teacher sees on screen even if a just-drawn stroke
+ * has not round-tripped through the Firestore listener yet. It also saves a
+ * redundant read on the common single-page export.
+ *
+ * Reads run in parallel — a page count in the low tens is the realistic
+ * ceiling for a classroom whiteboard, and the export is a user-initiated
+ * one-shot, so latency matters more than read smoothing.
+ */
```
OPS-PIvers: Multi-paragraph JSDoc block violates CLAUDE.md: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* The non-obvious invariant here is worth keeping (the post-migration objects-stripping behavior genuinely surprises readers), but the three-paragraph explanation can compress to a single line:

```suggestion
// Post-migration pages[].objects is stripped; read non-active pages from subcollection; live page uses in-memory objects to match what's on screen.
```

---

components/widgets/DrawingWidget/Widget.tsx
```diff
@@ -725,6 +726,49 @@ export const DrawingWidget: React.FC<{
...
   };
 
+  // Pages to hand the export pipeline, with `objects[]` guaranteed populated.
+  //
+  // Post-migration (PR 2.6) the dashboard doc's `pages[].objects` is emptied
+  // on purpose — it's a denormalized `{ id, background }` cache, and the real
+  // objects live in the page-nested subcollection. Handing `pages` straight
+  // to the exporter would therefore paint the background template and nothing
+  // else. The active page is covered by the live `objects` array; every OTHER
+  // page has no live subscription, so we read it from Firestore here.
```
OPS-PIvers: Same issue as the `hydrateDrawingPages.ts` docstring — eight-line comment block violates the "one short line max" convention. The core why (objects are stripped post-migration, non-active pages need a Firestore read) fits on one line:

```suggestion
  // Post-migration pages[].objects is empty; splice live objects for the active page, read the rest from Firestore.
```

---

OPS-PIvers (commented):
The fix is correct and the test coverage is solid. Two comment-style findings across the new files — both violate CLAUDE.md's "one short line max" convention for inline comments and docstrings.

---

utils/hydrateDrawingPages.ts
OPS-PIvers: Valid — collapsed the multi-paragraph JSDoc to a single line in `922b402`, using your suggested wording. The `{@link}` rationale about the denormalized cache and non-active-page reads is preserved in condensed form.

---

components/widgets/DrawingWidget/Widget.tsx
OPS-PIvers: Valid — collapsed the eight-line block to a single line in `922b402`. Kept the core "why" (post-migration `pages[].objects` is empty; active page uses live objects, others read from Firestore).

---

OPS-PIvers:
The existing review threads (now outdated) covered the JSDoc in `hydrateDrawingPages.ts` and the original block in `Widget.tsx` — both addressed in 922b402. Three more multi-line comment blocks added in this PR weren't covered by those threads. CLAUDE.md: *"one short line max."*

**`components/widgets/DrawingWidget/Widget.tsx`** — inside `resolveExportPages` and `handleExportCurrentPng`:
- 4-line pre-migration note → `// Pre-migration: pages[].objects is the live source of truth — no subcollection read needed.`
- 4-line uid/dashboardId guard → `// Migrated but can't address subcollection — fail loudly rather than exporting blank pages.`
- 3-line `activePage.objects` note in `handleExportCurrentPng` → `// Post-migration activePage.objects is empty; splice in live objects before single-page export.`

**`tests/components/widgets/DrawingWidget/exportSubcollection.test.tsx`** — the 12-line JSDoc block (lines 12–23). The `describe` and `it` strings already document the regression; remove the block.

**`tests/utils/hydrateDrawingPages.test.ts`** — the 3-line `// Mock the Firestore surface…` comment (lines 6–8). Collapses to: `// collection() returns the joined path so getDocs can route canned responses by subcollection address.`

---

PR #2314, opened 2026-07-31
nightly(dashboard-layout): stop OverflowMenu's Escape from minimizing an unrelated widget

OPS-PIvers:
## Root cause
`components/common/sessionViews/OverflowMenu.tsx`'s `onMenuKeyDown` closed the menu on Escape but never called `stopPropagation()`. The menu is portalled to `document.body` via `createPortal`, outside any `.widget` DraggableWindow ancestor. The unstopped keydown bubbles to `DashboardView`'s global `window`-level Escape handler, which — finding no `.widget` ancestor for the portalled menu — falls back to targeting the topmost z-index widget and minimizes it. Net effect: a teacher opens this menu (live in `QuizResults.tsx`/`Results.tsx` for Video Activity) and presses Escape to dismiss it, silently minimizing an unrelated widget on their board (e.g. a running timer).

This is the same recurring bug class already fixed in `ToolDockItem`/`RemoteControlMenu`/`ClassRosterMenu` (#2266) and `BoardNavFab`/`CollectionSwitcherMenu`/`BoardActionsFab` (#2289) — `OverflowMenu` was a fresh, previously-unfixed instance.

## Live consumers
`components/widgets/QuizWidget/components/QuizResults.tsx` and `components/widgets/VideoActivityWidget/components/Results.tsx` both render `OverflowMenu` inside live dashboard widgets.

## Fix
Added `e.stopPropagation()` in the Escape branch of `onMenuKeyDown`, mirroring the established fix pattern from #2266/#2289.

## Band-aid rejected
Could have special-cased `DashboardView`'s global handler to recognize portalled dropdowns, but the codebase's established, correct pattern (per #2266/#2289) is for each Escape-to-dismiss popover to stop its own propagation at the source — not for the global fallback handler to grow more exceptions.

## Regression test
`tests/components/common/sessionViews/OverflowMenu.test.tsx` — asserts a window-level `keydown` spy (stand-in for `DashboardView`'s global handler) is never invoked when Escape is pressed inside the open menu.

**Fail-before** (orchestrator-verified via file-swap against `dev-paul`'s pre-fix source):
```
Number of calls: 1
Test Files  1 failed (1)
     Tests  1 failed | 6 passed (7)
```

**Pass-after** (orchestrator-verified):
```
Test Files  1 passed (1)
     Tests  7 passed (7)
```

## Validate + build
[...]

PR #2401, opened 2026-08-06
fix(common): stop Escape from AIGeneratorOverlay reaching ancestor Modal

OPS-PIvers:
## Root cause

`AIGeneratorOverlay` (shared "Draft with AI" panel used by the Quiz/Video Activity/Mini App editors, all mounted via `EditorModalShell` → `Modal`) handled Escape with no `stopPropagation()`. `Modal.tsx` (the ancestor full-screen editor shell) registers its own unconditional `window`-level Escape handler that closes/prompts-to-discard the editor. Pressing Escape while focus was inside the AI panel (e.g. typing in the prompt textarea) bubbled past the overlay and also fired the outer editor Modal's handler — closing or prompting to discard the *entire* editor instead of just dismissing the small AI panel.

Confirmed live: `AIGeneratorOverlay` is consumed by `QuizEditor.tsx`, `VideoActivityEditor.tsx`, and `MiniAppEditorModal.tsx`.

## Fix

Added `e.stopPropagation()` in the overlay's Escape handler before calling `onClose()`, consistent with the same bug class's established fix location in `ToolDockItem`/`RemoteControlMenu`/`ClassRosterMenu` (#2266), `BoardNavFab`/`CollectionSwitcherMenu`/`BoardActionsFab` (#2289), and `OverflowMenu` (#2314).

## Rejected band-aid

Did not teach `Modal`/`isEscapeFromWidgetInput` to special-case this overlay — the established, already-repeated pattern is for the nested popover/overlay itself to own `stopPropagation()`.

## Regression test

`tests/components/common/AIGeneratorOverlay.escapePropagation.test.tsx` — renders the real `Modal` wrapping the real `AIGeneratorOverlay`, focuses a child textarea, fires Escape, asserts the overlay's own `onClose` fired but the ancestor `Modal`'s `onClose` did not.

- **Fail-before** (orchestrator-verified via file-swap against `dev-paul`): `expected "vi.fn()" to not be called at all, but actually been called 1 times`
- **Pass-after**: passing (ran together with `Modal.test.tsx` and `ActiveClassChip.escapePropagation.test.tsx`, 18/18 total, no regression on the sibling pattern)

## Verification

- `pnpm run validate` — `598 test files / 7283 tests` (root), `40 test files / 778 tests` (functions), test:counts guard passed
- Independently re-verified by the orchestrator via file-swap fail-before/pass-after, and a full isolated `validate`+`build` re-run post-rebase onto latest `dev-paul`

## Backlog (not fixed here — flagged for a future run)

`QuizEditorModal.tsx`'s own redundant `window`-level Escape listener for `showAiPrompt` still doesn't `stopPropagation()` — no longer observably harmful in the common case now that this fix stops the event at the overlay, but could still theoretically race with `Modal`'s handler in a focus-outside-overlay case. Out of this area's glob (`components/widgets/QuizWidget/**`).

## Risk

**contained** — single-file, minimal diff, matches an established and repeatedly-shipped fix pattern.

---

---

### DIFF
--- a/components/common/AIGeneratorOverlay.tsx
+++ b/components/common/AIGeneratorOverlay.tsx
@@ -65,7 +65,10 @@ export const AIGeneratorOverlay: React.FC<AIGeneratorOverlayProps> = ({
       aria-label={title}
       className="absolute inset-0 z-20 bg-white/95 backdrop-blur-sm flex items-center justify-center p-6 animate-in fade-in duration-200"
       onKeyDown={(e) => {
-        if (e.key === 'Escape') onClose();
+        if (e.key !== 'Escape') return;
+        // Stop the bubble to the ancestor Modal's window-level Escape handler, which would otherwise also close/discard the whole editor (same bug class as #2266/#2289/#2314).
+        e.stopPropagation();
+        onClose();
       }}
     >
       <div className="w-full max-w-sm space-y-4">
--- a/tests/components/common/AIGeneratorOverlay.escapePropagation.test.tsx
+++ b/tests/components/common/AIGeneratorOverlay.escapePropagation.test.tsx
@@ -1,0 +1,62 @@
+/**
+ * Regression test: AIGeneratorOverlay's Escape handler closed the AI
+ * generator panel, but never called stopPropagation(). The overlay is
+ * rendered as `children` inside the shared `Modal` primitive (via
+ * EditorModalShell, used by the Quiz/Video Activity/Mini App editors).
+ * Modal registers its own unconditional `window`-level Escape handler
+ * that calls the ancestor editor's onClose. Because AIGeneratorOverlay
+ * never stopped propagation, pressing Escape while focus was inside the
+ * "Draft with AI" panel (e.g. typing in the prompt textarea) bubbled past
+ * the overlay and ALSO triggered Modal's onClose — closing/discarding the
+ * entire editor instead of just dismissing the small AI panel.
+ *
+ * FIX: the handler now calls event.stopPropagation() before invoking
+ * onClose(), matching the same fix already applied to ToolDockItem,
+ * RemoteControlMenu, ClassRosterMenu, OverflowMenu, and ActiveClassChip
+ * for this exact bug class.
+ */
+
+import React from 'react';
+import { describe, it, expect, vi, afterEach } from 'vitest';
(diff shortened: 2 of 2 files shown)


OPS-PIvers:
The 16-line multi-paragraph block comment at the top of `tests/components/common/AIGeneratorOverlay.escapePropagation.test.tsx` violates CLAUDE.md: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."*

The `describe` and `it` names already document the regression behavior ("Escape does not leak to ancestor Modal" / "closes only the overlay on Escape, not the ancestor Modal"). The bug history, the fix rationale, and the list of previously patched components all belong in the PR description, not the test file.

Suggest removing the block comment entirely — the test is self-documenting through its names.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-07

### Summary
Adds `e.stopPropagation()` to `AIGeneratorOverlay`'s Escape handler so pressing Escape inside the AI panel dismisses only the overlay instead of also bubbling to the ancestor `Modal`'s window-level Escape handler and discarding the whole editor. Uses the established early-return guard pattern and ships with a regression test.

### Code Quality
✓ No issues found. `if (e.key !== 'Escape') return; e.stopPropagation(); onClose();` is the clean, established guard shape (matching the referenced ToolDockItem / RemoteControlMenu / etc. fixes). Because React 17+ delegates events at the app root (an ancestor of which is `window`), calling the synthetic `stopPropagation()` invokes the native `stopPropagation` and prevents the event from reaching Modal's `window`-level `keydown` listener — so the fix is mechanically correct, not just cosmetic. No modal-stacking breakage: only Escape is intercepted, and only when the overlay is mounted/open.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. No data-layer changes.

### Regression Risk
✓ Low risk. Single event handler in a shared common component; behavior narrows (stops an unwanted double-close) rather than widening.

### Missing Tests
✓ Adequate coverage. New `AIGeneratorOverlay.escapePropagation.test.tsx` renders the overlay inside a real `Modal` and asserts Escape calls the overlay's onClose but not the Modal's onClose.

### Overall
**Merge readiness:** Ready
Correct, minimal fix for a real double-close bug, with a regression test that exercises the actual Modal nesting.

---

---

PR #2403, opened 2026-08-06
fix(functions): paginate expireSubShares expired-doc lookup

OPS-PIvers:
## Root cause

`functions/src/expireSubShares.ts` (the hourly sweep that deletes expired substitute-mode `shared_boards`/`shared_collections` docs) queried expired docs with a single un-paginated `.limit(MAX_DELETES_PER_RUN)` (500), ordered implicitly by `expiresAt`.

Docs carrying `driveGrants[]` intentionally sit in the query's result set for up to 7 days (the client-revoker grace window). Once more than one page of simultaneously-expiring grace-window docs accumulated, that same oldest page was re-fetched every hourly run — permanently pushing out any later-sorting expired share, **including one with no grants that was supposed to be deleted immediately**. Sensitive substitute-portal snapshots would stay reachable far longer than intended.

Notably, `gcPlcOrphans.ts` and `expireActivityWallShares.ts` both explicitly cite this file as the template they were modeled on — both were independently fixed for the identical pagination-cliff bug in prior nightly runs, but the fix was never carried back to the original. This file also had zero test coverage before this change.

## Fix

Split read vs. write concerns: the fetch now paginates (`startAfter` cursor on `orderBy('expiresAt').orderBy(FieldPath.documentId())`) up to a `MAX_READ_PER_RUN` safety ceiling (5000), mirroring the established pattern in the two derivative files. `MAX_DELETES_PER_RUN` still caps actual batched writes per run so one slow sweep can't monopolize the scheduler slot — any write overflow is picked up on the next hourly tick instead of being starved indefinitely.

## Rejected band-aid

Did not just raise `MAX_DELETES_PER_RUN` (would still cliff at a bigger number) or add an index-only ordering tweak without pagination.

## Regression test

New file `functions/src/expireSubShares.test.ts` (previously zero coverage): 5 baseline-behavior tests + 1 pagination regression test + 1 wrapper-import test.

- **Fail-before**: orchestrator independently re-verified with a targeted logic mutation (not just an export-shape revert) — forcing `fetchExpiredPaginated`'s loop to stop after the first page reproduced the exact bug: `AssertionError: expected [] to deeply equal [ 'target-no-grants' ]`, while the other 6 baseline tests in the same file continued to pass (confirming the test pins the pagination logic specifically, not just module existence).
- **Pass-after**: 7/7 passing

## Verification

- `pnpm run validate` — clean (root: 597 test files/7282 tests, functions: 41 test files/785 tests — up from 40 files pre-change, test:counts guard passed)
- `firestore.rules` not touched, `test:rules` not required
- Independently re-verified by the orchestrator via targeted mutation fail-before/pass-after, and a full isolated `validate`+`build` re-run post-rebase onto latest `dev-paul`

## Risk

**contained** — single-file Cloud Function fix with new dedicated test coverage; mirrors an already-proven fix pattern shipped twice before for sibling sweep functions.

---

---

### DIFF
--- a/functions/src/expireSubShares.test.ts
+++ b/functions/src/expireSubShares.test.ts
@@ -1,0 +1,456 @@
+// Unit tests for the `expireSubShares` hourly sweep.
+//
+// Regression coverage for the pagination-cliff bug: the expired-substitute-doc
+// lookup used a single un-paginated `.limit(MAX_DELETES_PER_RUN)` query. Docs
+// with `driveGrants[]` sit in query results for up to `ORPHAN_GRACE_DAYS`
+// without being deleted (intentional grace window), so once more than one
+// page of simultaneously-expiring grace-window docs accumulated, that same
+// oldest-`expiresAt` page was re-fetched every run — starving any
+// later-sorting candidate (including a doc with NO grants, meant to delete
+// *immediately*) out of every run's window until the backlog aged out. Mirrors
+// the identical bug already fixed in `gcPlcOrphans.ts` / `expireActivityWall
+// Shares.ts` — both of which were originally modeled on THIS file's shape.
+//
+// Two layers, mirroring `expireActivityWallShares.test.ts`:
+//   1. BASELINE SWEEP BEHAVIOR — immediate delete, grace-window skip,
+//      grace-expired forced delete, boards subcollection cleanup.
+//   2. PAGINATION REGRESSION — a ready-to-delete doc that sorts past the
+//      first page must still be read and deleted in the same run.
+
+import { describe, it, expect, vi } from 'vitest';
+
+vi.mock('firebase-admin', () => ({
+  apps: [{ name: '[DEFAULT]' }],
+  initializeApp: vi.fn(),
+  // The sweep paginates with admin.firestore.FieldPath.documentId(); expose
+  // it as the '__name__' sentinel so the stub CollectionRef.orderBy gets a
+  // stable value (mirrors gcPlcOrphans.test.ts's / expireActivityWallShares
+  // .test.ts's stub).
+  firestore: Object.assign(vi.fn(), {
+    FieldPath: { documentId: vi.fn(() => '__name__') },
+  }),
+}));
+
+vi.mock('firebase-functions/v2/scheduler', () => ({
+  onSchedule: (_opts: unknown, handler: () => Promise<void>) => handler,
+}));
+
+import { sweepCollection, READ_PAGE_SIZE } from './expireSubShares';
+
+const NOW = 1_700_000_000_000;
+const DAY = 24 * 60 * 60 * 1000;
+
+// ===========================================================================
+// Stub Firestore mirroring the Admin SDK surface `sweepCollection` uses:
+//   db.collection(name).where(f, op, v).orderBy(field).limit(n)
+//     .startAfter(docSnap).get()
+//   docSnap.ref.collection('boards').get()
+//   db.batch().delete(ref).commit()
+// `orderBy`/`startAfter` mirror `expireActivityWallShares.test.ts`'s stub:
+// sort by the ordered field(s) (falling back to doc id for the '__name__'
+// sentinel) and slice past the cursor doc's ordering-field values — this is
+// what actually exercises cross-page pagination in the regression test below.
+// Absent an explicit `orderBy`, Firestore itself implicitly orders an
+// inequality-filtered query ascending by the filtered field — the stub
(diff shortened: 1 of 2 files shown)


OPS-PIvers:
Two categories of multi-line comment violations (CLAUDE.md: *"one short line max"*):

**Test file — `functions/src/expireSubShares.test.ts`:** The 21-line `//` block at the top duplicates context already in the module header and PR description. The `describe`/`it` strings convey the intent. Remove the block.

**Production file — `functions/src/expireSubShares.ts`:** Several new multi-paragraph JSDoc/comment blocks added in this PR:
- The 12-line block appended to the existing module header (the `*` list ending at "Fetching now paginates…") — the key constraint compresses to one sentence appended to the existing header, or just a single new line.
- `MAX_READ_PER_RUN` — 5-line `/** */` → `// Per-run read ceiling; paginates so a grace-window backlog never starves later-sorting candidates.`
- `fetchExpiredPaginated` — multi-line JSDoc → `// Paginates the expired-doc query via startAfter cursor so grace-window docs can't block later candidates.`
- `sweepCollection` JSDoc — the new multi-line note added to the existing doc; fold the key point ("now reads from fetchExpiredPaginated") into a brief update to the existing single-line description.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-07

### Summary
Paginates the `expireSubShares` expired-doc lookup, replacing the single un-paginated `.limit(MAX_DELETES_PER_RUN)` page with a `startAfter` cursor loop over `orderBy('expiresAt').orderBy(FieldPath.documentId())`, so grace-window docs can no longer starve later-sorting ready-to-delete candidates. The pagination logic is correct and the change ships with thorough regression tests. **Duplication note:** this is the same expireSubShares pagination fix as the already-reviewed, still-open #2391 (nightly/build-tooling-2026-08-05) — the two nightly branches have piled up a duplicate fix; only one should be merged and the other closed to avoid a redundant/conflicting merge.

### Code Quality
✓ No issues found. Pagination loop is sound: `while (results.length < MAX_READ_PER_RUN)`, per-page limit is `min(READ_PAGE_SIZE, remaining)`, cursor advances via `lastDoc = page.docs[last]`, and it breaks on both `page.empty` and a short page (`page.size < pageLimit)` — no infinite loop, stable ordering via the documentId tiebreaker. `await` used throughout (no floating promises), `console.warn` on the runaway ceiling is intentional operational logging (not a stray `console.log`). The `MAX_DELETES_PER_RUN` write cap is correctly re-applied after the read side (which now can surface more docs), with overflow deferred to the next tick.

Minor (non-blocking): appending `.orderBy(FieldPath.documentId())` to the existing `(intendedMode ASC, expiresAt ASC)` composite index is fine — Firestore implicitly appends `__name__`, so no new index is required — but worth a manual confirmation that the existing declared index still satisfies the query in prod.

### Widget Standards
✓ Not applicable (Cloud Functions only).

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. No new collections/rules; query shape unchanged aside from ordering. No auth bypass or client secrets (server-side function).

### Regression Risk
Low-to-moderate — touches `functions/` but NOT `functions/src/index.ts` exports of the scheduled function; the `expireSubShares` scheduled export signature is unchanged. `sweepCollection`, `MAX_DELETES_PER_RUN`, `MAX_READ_PER_RUN`, and `READ_PAGE_SIZE` are newly exported (for testability) — additive named exports, low collision risk.

### Missing Tests
✓ Adequate coverage. New `expireSubShares.test.ts` covers baseline behavior (immediate delete, grace-window skip, grace-expired forced delete, boards subcollection reap) plus the pagination regression (no-grants target sorting past a full first page of grace-window fillers) and the scheduled-wrapper export.

### Overall
Correct, well-tested pagination fix — but coordinate with #2391 so the identical fix isn't merged twice, and confirm the composite index in prod.

---

---

OPS-PIvers:
Closing as a duplicate in favor of #2391 (2026-08-05). I compared both diffs directly: #2391 is the more complete fix — it identifies a sharper root cause (a Firestore query with no explicit `orderBy` falls back to document-ID order rather than the range-filtered field, the same mechanism already confirmed in `expireActivityWallShares.ts` #2268), extracts a fully testable `runExpireSubShares(db, now)` core covering both sweeps together (including the `Promise.allSettled` error handling), and ships a larger regression suite. This PR's fix is correct but narrower in scope — recommend merging #2391 instead.


---

PR #2400, opened 2026-08-06
fix(widgets): stop Escape propagation in RandomClassContextButton popover

OPS-PIvers:
## Root cause

`components/widgets/random/RandomClassContextButton.tsx` — the class-switcher popover (rendered inside the live `RandomWidget`) is portalled to `document.body` outside any `.widget` `DraggableWindow` ancestor, and dismissed on Escape via a `document.addEventListener('keydown', ...)` listener that never called `stopPropagation()`.

`components/layout/DashboardView.tsx` installs a global `window`-level Escape handler. When it fires and finds no `.widget` ancestor for the focused element (always true for portal-to-body content), it falls back to minimizing the topmost z-index widget on the whole board. Dismissing this small popover with Escape would therefore silently minimize (or un-flip settings on) an unrelated widget elsewhere on the dashboard.

This is the same bug class already fixed 4 times in this repo (#2266, #2289, #2314, #2373 — the last of which fixed `ActiveClassChip`, the component `RandomClassContextButton` is a documented fork of). The fork reintroduced the gap that `ActiveClassChip` had already patched.

## Fix

Added `event.stopPropagation()` in the popover's Escape branch before `closeMenu()`, matching the established pattern (and comment) from the `ActiveClassChip` fix verbatim.

## Rejected band-aid

Did not touch `DashboardView.tsx`'s fallback-to-topmost-widget logic — that's the shared root cause behind all 5 instances of this bug class and is out of this fix's file scope. The sustainable fix is for each portalled popover to own its Escape interaction via `stopPropagation()`, consistent with all prior fixes.

## Regression test

`components/widgets/random/RandomClassContextButton.test.tsx` — opens the popover, attaches a `window`-level `keydown` spy, fires Escape on `document`, asserts the popover closed **and** the spy was never called.

- **Fail-before** (orchestrator-verified via file-swap against `dev-paul`): `expected "vi.fn()" to not be called at all, but actually been called 1 times`
- **Pass-after**: 13/13 tests in file passing

## Verification

- `pnpm run validate` (type-check:all, lint, format:check, unit + functions tests, test:counts) — clean
- `pnpm run build` — succeeded
- Independently re-verified by the orchestrator via file-swap fail-before/pass-after, and a full isolated `validate`+`build` re-run post-rebase onto latest `dev-paul`

## Backlog (not fixed here — flagged for a future run)

[...]

PR #2407, opened 2026-08-07
fix(widgets): stop Escape propagation in RandomClassContextButton popover

OPS-PIvers:
## Root cause

`RandomClassContextButton` (the Randomizer widget's class-switcher/mark-absent popover, mounted live in `RandomWidget.tsx`) portals its menu to `document.body` and closed it on Escape via a bare `document.addEventListener('keydown', ...)` handler that never called `stopPropagation()`.

Because the popover renders outside any `.widget`/`DraggableWindow` ancestor, the unhandled keydown bubbled from `document` up to `window`, where `DashboardView.tsx`'s global Escape handler fires unconditionally, finds no typing field / no `.widget` ancestor for the popover trigger, and falls back to minimizing the topmost z-index widget. Dismissing the Randomizer's class popover with Escape could silently minimize an unrelated widget on the live board.

This is the same "portalled popover Escape doesn't stopPropagation" bug class already fixed 4x elsewhere in this codebase (`ToolDockItem`/`RemoteControlMenu`/`ClassRosterMenu`, `BoardNavFab`/`CollectionSwitcherMenu`/`BoardActionsFab`, `OverflowMenu`) — this is a fresh, previously-unaudited instance in the Randomizer widget.

## Fix

Added `event.stopPropagation()` in the Escape handler, matching the established fix pattern exactly.

## Rejected band-aid

Guarding with `isEscapeFromWidgetInput(e)` (used elsewhere to protect text-input Escape) would not fix this — the popover isn't a text input, it's the unhandled-bubbling problem itself. The only root-cause fix is stopping propagation at the source, not special-casing `DashboardView`'s global handler to exclude this one popover.

## Regression test

`components/widgets/random/RandomClassContextButton.test.tsx` — opens the popover, dispatches a bubbling Escape `KeyboardEvent` on `document`, and asserts a `window`-level keydown spy is never invoked.

**Fail-before** (orchestrator-verified independently via file-swap against `dev-paul`'s pre-fix source):
```
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
 ❯ expect(windowKeydownSpy).not.toHaveBeenCalled();
 Test Files  1 failed (1)
      Tests  1 failed | 12 passed (13)
```

**Pass-after** (fix restored):
```
 Test Files  1 passed (1)
      Tests  13 passed (13)
```

## Full validation

`pnpm run validate` and `pnpm run build` both green on this branch (independently re-run by the orchestrator after a review-driven comment trim):
- type-check:all — clean
- lint — 0 errors/warnings
- format:check — clean
- test — 598 test files / 7283+ tests passed
- functions test — 40 test files / 778 tests passed (unaffected)
- test:counts guard — passed
- build — succeeded

## Risk tag

**Contained.** Single-file, 2-line logic fix plus one new test, matching an exact, already-4x-precedented pattern.

---
Part of the nightly automated code-health run (run #36, 2026-08-07). See `docs/routines/debugger.md` for the full routine.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HNqPofhNVn9C75uyVwr7Px)_

### DIFF
--- a/components/widgets/random/RandomClassContextButton.test.tsx
+++ b/components/widgets/random/RandomClassContextButton.test.tsx
@@ -281,4 +281,32 @@ describe('RandomClassContextButton', () => {
     ).not.toBeInTheDocument();
     expect(container.firstChild).not.toBeNull();
   });
+
+  it('closes on Escape and stops propagation before it reaches window listeners', () => {
+    const r1 = makeRoster('r1', 'Period 1');
+    const r2 = makeRoster('r2', 'Period 2');
+    mockUseDashboard([r1, r2], 'r1');
+    render(
+      <RandomClassContextButton
+        roster={r1}
+        rosterMode="class"
+        onOpenAbsentModal={noop}
+      />
+    );
+    fireEvent.click(
+      screen.getByRole('button', { name: /Active class: Period 1/i })
+    );
+    expect(screen.getAllByRole('menuitemradio')).toHaveLength(2);
+
+    const windowKeydownSpy = vi.fn();
+    window.addEventListener('keydown', windowKeydownSpy);
+
+    try {
+      fireEvent.keyDown(document, { key: 'Escape', bubbles: true });
+      expect(screen.queryByRole('menuitemradio')).not.toBeInTheDocument();
+      expect(windowKeydownSpy).not.toHaveBeenCalled();
+    } finally {
+      window.removeEventListener('keydown', windowKeydownSpy);
+    }
+  });
 });
--- a/components/widgets/random/RandomClassContextButton.tsx
+++ b/components/widgets/random/RandomClassContextButton.tsx
@@ -81,13 +81,7 @@ export const RandomClassContextButton: React.FC<
     };
     const handleKey = (event: KeyboardEvent) => {
       if (event.key !== 'Escape') return;
-      // This popover is portalled to document.body, outside any `.widget`
-      // DraggableWindow. Without stopping propagation, an unhandled Escape
-      // here bubbles up to DashboardView's global window-level Escape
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-07

### Summary
Adds `event.stopPropagation()` to the RandomClassContextButton popover's Escape handler so dismissing the portalled menu doesn't bubble to DashboardView's global handler and minimize the topmost widget. The fix itself is correct and well-tested. **Duplication flag:** this is the same fix as #2400 (nightly/widgets-2026-08-06) and the already-reviewed, established-pattern #2388 (08-05). All three are still open and unmerged — an identical-fix pile-up. Only one should be merged; the other two should be closed.

### Code Quality
- Correct: early-returns on non-Escape keys, then `stopPropagation()` before `closeMenu()`. Comment explains the portal-outside-`.widget` rationale.
- Cleaner than the #2400 variant — the `key !== 'Escape'` early-return reads better than the inline `if (event.key === 'Escape')`, and the test additionally asserts the `menuitemradio` count.
- No TypeScript/lint concerns.

### Widget Standards
✓ Not applicable / event-handler change only. No front-face content or registry wiring touched.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found.

### Regression Risk
Low on its own. The real risk here is process, not code: three open PRs mutating the same handler will conflict. Merge one and close #2400 and #2388 (or #2400 and this one) to avoid churn.

### Missing Tests
✓ Adequate coverage. Opens the popover, dispatches Escape, asserts the menu closes and the window keydown spy is never called.

### Overall
Fix is correct and is the strongest of the three duplicates; merge exactly one and close the other two (#2400, #2388).

---

---

PR #2405, opened 2026-08-07
fix(a11y): retrofit 2 more orphaned SettingsLabel group headings to as="span"

OPS-PIvers:
## Summary

Nightly unifier run 50, dimension D3 (Settings Panel Label Primitives). Continues the runs 26-49 "group-heading `SettingsLabel` retrofit" series: converts `<SettingsLabel>` instances that head a **group** of sibling controls (not a single 1:1 control) from the default `as="label"` to `as="span"` + `role="group"`/`aria-labelledby` on the sibling container.

**⚠️ Update:** this PR originally touched 7 files. An automated review plus independent diff comparison found that **PR #2397** (nightly run 08-06, still open) already contains functionally-identical fixes for 5 of those files (`GridPresetCard.tsx`, `ClockWidget/Settings.tsx`, `Countdown/Settings.tsx`, `LunchCount/Settings.tsx`, `TimeTool/Settings.tsx`) — three consecutive nightly runs (#2385 on 08-05, #2397 on 08-06, this one on 08-07) independently swept the same never-merged `dev-paul` state and produced overlapping diffs. Reverted those 5 files here to avoid shipping a conflicting duplicate; this PR now only carries the 2 genuinely new instances from tonight's sweep.

**Canonical form:** `<SettingsLabel as="span" id="...">Heading</SettingsLabel>` paired with `role="group" aria-labelledby="..."` on the container of sibling controls.

**Instances unified (2 files, 2 headings):**
- `components/widgets/PollWidget/Settings.tsx` — "Actions" (2-button reset/end group)
- `components/widgets/SeatingChart/Settings.tsx` — "Actions" (2 full-width destructive-action buttons, now wrapped in a `role="group"` container)

**Behavior-preservation evidence:**
- `components/common/SettingsLabel.tsx`'s `combinedClasses` is computed once and shared identically by both the `as="label"` and `as="span"` branches — zero visual delta.
- `pnpm exec tsc --noEmit` — clean
- `pnpm exec eslint <changed files> --max-warnings 0` — clean
- `pnpm exec prettier --check <changed files>` — clean
- `pnpm test` — 597/597 test files, 7282/7282 tests passing (identical to baseline), captured before the trim; the trim only reverted files, it re-ran `tsc --noEmit` clean afterward

**Risk tag:** mechanical (pure a11y/semantics change, zero visual delta)

## ⚠️ Repo hygiene finding this run surfaced

[...]

PR #2385, opened 2026-08-05
fix(a11y): retrofit orphaned SettingsLabel group headings to as="span"

OPS-PIvers:
## Nightly Unifier — Run 50, Dimension D3 (Settings Panel Label Primitives)

**Canonical form:** `<SettingsLabel as="span" id="unique-id">Group Heading</SettingsLabel>` paired with `role="group" aria-labelledby="unique-id"` on the container of sibling controls — used when a label heads a *group* of controls (button/chip groups) rather than pairing 1:1 with a single form control. A bare `<label>` in that case is semantically orphaned (no single control to associate with). This continues the runs 26–49 group-heading retrofit series.

**Instances unified (5, across 4 files):**
- `components/admin/BackgroundManager/GridPresetCard.tsx` — "Access Level" and "Buildings" button groups (id scoped per-preset via `preset.id`, since this card renders in a list)
- `components/widgets/ClockWidget/Settings.tsx` — "Display Style" and "Color Palette" button groups (id scoped via `useId()`)
- `components/widgets/Countdown/Settings.tsx` — "Event Title Color" swatch group (id scoped via `useId()`)
- `components/widgets/LunchCount/Settings.tsx` — "Grade Level" button group (id scoped via `useId()`)

**Enforcement:** No new lint rule added — this dimension's enforcement is the existing `SettingsLabel` component's `as="span"` variant itself (added in an earlier PR, #2141 review-fix). Tonight's work is a retrofit of pre-existing bare-`<label>` usages to the already-established group-heading form, continuing the ongoing series that has covered this same anti-pattern across many other files in prior nightly runs.

**Behavior/visual-preservation evidence:**
- Visual output is unchanged — `SettingsLabel`'s `as="span"` renders identical Tailwind classes to `as="label"` (`text-xxs font-black text-slate-400 uppercase tracking-widest block mb-2`), just a different DOM tag.
- Accessibility is *improved*, not changed in a risky way: previously an orphaned `<label>` with no `htmlFor`/associated control now becomes a `<span>` correctly paired with its sibling button group via `role="group"`/`aria-labelledby`.
- Each `id` is scoped per-component-instance (`useId()` for singleton Settings panels; `preset.id`-derived for the list-rendered `GridPresetCard`) to avoid DOM id collisions across multiple widget/preset instances on the same dashboard — a documented gotcha from earlier runs in this series.
- `pnpm run validate` (type-check, lint, format-check, full unit test suite — root + functions) passes clean on this branch.

[...]

PR #2367, opened 2026-08-04
fix(a11y): retrofit orphaned SettingsLabel group headings to as="span" (D3, nightly unifier)

OPS-PIvers:
## Canonical form

`SettingsLabel` heading a group of sibling toggle/chip/swatch buttons (no single paired input) should use `as="span" id={...}` + `role="group" aria-labelledby={...}` on the sibling-controls container, instead of the default `as="label"` (which is semantically orphaned — a `<label>` with no single associated control). This is the established convention from the runs 26-48 group-heading retrofit series (`DrawingWidget`, `RevealGrid`, `RandomSettings`, `MusicWidget`, `SyntaxFramer`, `Checklist`, `MathToolInstance`, `ClockConfigurationPanel`, `TextWidget`, `WorkSymbols`, `TimeToolConfigurationPanel`).

## Instances unified (10 across 6 files)

- `components/widgets/Breathing/BreathingSettings.tsx` — "Pattern", "Visual Style", "Color Theme"
- `components/admin/BreathingConfigurationPanel.tsx` — "Default Pattern", "Default Visual Style", "Default Color Theme"
- `components/widgets/NextUp/Settings.tsx` — "Visual Style"
- `components/widgets/HotspotImage/Settings.tsx` — "Popover Theme"
- `components/widgets/DiceWidget/Settings.tsx` — "Number of Dice"
- `components/widgets/MaterialsWidget/Settings.tsx` — "Available Materials"

Each converted `SettingsLabel` gets `as="span" id={<name>LabelId}` (via `useId()`, since none of these 6 files had an existing `widget.id`-interpolation convention of their own) and the sibling button-group container gets `role="group" aria-labelledby={<name>LabelId}`.

## Enforcement

No new enforcement added this run (this is itself the continuation of an established retrofit convention, not a fresh anti-pattern) — a genuinely fresh sweep of files outside the previously-audited list found these 10 additional untouched instances, consistent with the pattern from runs 46-48 that a "closed" retrofit series is only closed for the files actually audited, not repo-wide.

## Behavior/visual-preservation evidence

Read `components/common/SettingsLabel.tsx` directly: `combinedClasses` is computed once from `baseClasses`/`layoutClasses`/`className` and is shared byte-identically by both the `as="label"` and `as="span"` render branches — only the wrapping element and `htmlFor` semantics differ. This is a pure a11y-semantics fix (orphaned `<label>` → `role="group"`/`aria-labelledby` pairing) with zero visual delta.

[...]

PR #2263, opened 2026-07-23
D3 nightly unify: SyntaxFramer Alignment label → SettingsLabel group-heading form

OPS-PIvers:
## Summary

Nightly Unifier routine (run 41), dimension D3 — Settings Panel Label Primitives.

- **Canonical form:** `<SettingsLabel>` group-heading variant (`as="span" id="..."` + `role="group" aria-labelledby="..."` on the sibling-controls container) for labels that head a group of buttons/chips rather than pairing 1:1 with a single control, where a bare `<label>` would be semantically orphaned. This variant was added in an earlier PR (#2141) and this routine has been retrofitting pre-existing group-heading `SettingsLabel` call sites to it since run 31/32.
- **Instance unified:** `components/widgets/SyntaxFramer/Settings.tsx` — the "Alignment" label heading a 2-button (left/center) alignment-selector group. Converted from bare `<SettingsLabel icon={AlignLeft}>` to `as="span" id={syntaxframer-alignment-label-${widget.id}}` + `role="group" aria-labelledby` on the button-group `<div>`. `widget.id`-scoped (not `useId()`) to match this same file's existing convention for the sibling "Mode" label (converted last run, PR #2259) — `SyntaxFramerAppearanceSettings` renders per-widget-instance via `DraggableWindow`, so a static id would collide across two simultaneously-flipped instances of the widget.
- This closes the multi-run group-heading retrofit series tracked since run 31 (DrawingWidget → RevealGrid ×2 → RandomSettings → MusicWidget ×2 → SyntaxFramer ×2). No further candidates found.

## Behavior/visual preservation evidence

Read `components/common/SettingsLabel.tsx` directly: `combinedClasses` (the only thing that affects rendering) is computed once from `baseClasses`/`layoutClasses`/`className` before branching on `as` — the `as="span"` and default `as="label"` branches apply the identical class string to their respective element. Zero visual delta.

## Validation

- `pnpm run type-check:all`: 0 errors (root + functions)
- `pnpm run lint` (app + functions): 0 errors/warnings
- `pnpm run format:check`: clean
- `pnpm test`: 580/580 test files, 7045/7045 tests passing
- `pnpm -C functions test`: 38/38 test files, 719/719 tests passing
- `pnpm run test:counts`: guard passed
- `pnpm run build`: succeeds (main bundle + SSR prerender)

## Risk tag

**mechanical** (pure equivalence — semantics-only accessibility fix, zero visual delta, single-file diff)

---
[...]

PR #2416, opened 2026-08-09
fix(a11y): retrofit orphaned SettingsLabel group headings in shared settings primitives

OPS-PIvers:
## Nightly Unifier — Dimension D3 (Settings Panel Label Primitives)

**Risk tag:** mechanical (pure DOM-tag + ARIA equivalence, zero visual delta)

### Canonical form

`<SettingsLabel as="span" id={uniqueId}>Heading</SettingsLabel>` + `role="group" aria-labelledby={uniqueId}` on the sibling-controls container, used when a label heads a *group* of controls (button/chip groups) rather than pairing 1:1 with a single form control. A bare `<label>` in that case is semantically orphaned — no single control to associate with.

This is an established, previously-shipped convention (see `docs/routines/unifier.md` D3 series — PRs #2367, #2345, #2263, #2385, #2405, etc.), not a new pattern.

### Instances unified (4, across 3 files)

- `components/common/TypographySettings.tsx` — "Typography" font grid (line ~32), "Text Color" swatch grid (line ~65)
- `components/common/TextSizePresetSettings.tsx` — "Text Size" preset grid (line ~47)
- `components/common/SurfaceColorSettings.tsx` — surface/card color group (line ~28)

These are shared settings primitives consumed by 25+ widget `Settings.tsx` and admin `ConfigurationPanel.tsx` files — fixing them here fixes every current and future consumer at the source, rather than requiring a per-widget sweep.

### Enforcement

None added beyond the fix itself — the fix *is* the enforcement here, since it's applied at the shared-primitive level rather than per-consumer. Any current or future widget/admin panel that uses `TypographySettings`, `TextSizePresetSettings`, or `SurfaceColorSettings` automatically gets the accessible group-heading markup.

### Behavior/visual preservation evidence

- `components/common/SettingsLabel.tsx` computes `combinedClasses` once, identically, regardless of the `as` prop — the only difference between `as="label"` and `as="span"` is the rendered DOM tag and the `role="group"`/`aria-labelledby` wiring on the container. No className, spacing, color, or icon changes.
- `pnpm run validate` — exit 0 (type-check, lint `--max-warnings 0`, format-check, root tests: 604 files / 7321+ tests, functions tests: 41 files / 785 tests, `test:counts` guard passed)
- `pnpm run build` — app build + SSR bundle + prerender (`/privacy`, `/terms`, `/support`) all succeeded
- `pnpm exec prettier --check` run individually against the 3 changed files — clean
- Both the subagent and the orchestrator independently ran the full validate + build suite against this change

### Deferred (left for a future run, not this PR)

- `components/widgets/TimeTool/Settings.tsx` has 5 orphaned group headings in the identical shape (Mode, Display Style, Sound Selector/Alert Sound, Number Style, Color Palette) — left out to keep this diff tight and reviewable, per the routine's one-pattern-per-PR guidance. Flagged as the next clear D3 candidate.
- Near-miss lighter-weight label variants (missing `font-black`, `text-slate-500` instead of `text-slate-400`) across several admin `*ConfigurationPanel.tsx` files are pre-existing catalogued intentional exceptions (D3-E25) — converting would change visual weight, not touched.

### Note on provenance

These exact 3 files were previously proposed in PR #2397 (nightly run 50), which was closed unmerged. Verified directly against current file content (not the memory doc, which had assumed #2397 landed) that the fix was never actually applied to `dev-paul`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BNJegXGzf1TTwyyoDnVGz2)_

### DIFF
--- a/components/common/SurfaceColorSettings.tsx
+++ b/components/common/SurfaceColorSettings.tsx
@@ -46,7 +46,7 @@ export const SurfaceColorSettings = <
                   : 'border-slate-200'
               }`}
               style={{ backgroundColor: color }}
-              aria-label={`Select surface color ${color}`}
+              aria-label={`Select ${label.toLowerCase()} ${color}`}
             />
           ))}
         </div>
@@ -58,7 +58,7 @@ export const SurfaceColorSettings = <
             updateConfig({ cardColor: e.target.value } as Partial<T>)
           }
           className="h-8 w-full rounded-md border border-slate-200 bg-white"
-          aria-label={`Custom ${label.toLowerCase()} color`}
+          aria-label={`Custom ${label}`}
         />
 
         <div>
--- a/components/common/SurfaceColorSettings.tsx
+++ b/components/common/SurfaceColorSettings.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useId } from 'react';
 import { Palette, LucideIcon } from 'lucide-react';
 import { WidgetConfig } from '@/types';
 import { SettingsLabel } from './SettingsLabel';
@@ -21,11 +21,18 @@ export const SurfaceColorSettings = <
 }: SurfaceColorSettingsProps<T>) => {
   const cardColor = config.cardColor ?? '#ffffff';
   const cardOpacity = config.cardOpacity ?? 1;
+  const surfaceLabelId = useId();
 
   return (
     <div>
-      <SettingsLabel icon={icon}>{label}</SettingsLabel>
-      <div className="space-y-3 rounded-xl border border-slate-100 bg-slate-50 p-3">
+      <SettingsLabel icon={icon} as="span" id={surfaceLabelId}>
+        {label}
+      </SettingsLabel>
+      <div
+        className="space-y-3 rounded-xl border border-slate-100 bg-slate-50 p-3"
+        role="group"
+        aria-labelledby={surfaceLabelId}
+      >
         <div className="flex flex-wrap gap-2">
           {SURFACE_COLOR_PRESETS.map((color) => (
             <button
--- a/components/common/SurfaceColorSettings.tsx
+++ b/components/common/SurfaceColorSettings.tsx
@@ -46,7 +46,7 @@ export const SurfaceColorSettings = <
                   : 'border-slate-200'
               }`}
               style={{ backgroundColor: color }}
-              aria-label={`Select ${label.toLowerCase()} ${color}`}
+              aria-label={`Select ${label.toLowerCase()} color ${color}`}
             />
           ))}
         </div>
@@ -58,7 +58,7 @@ export const SurfaceColorSettings = <
             updateConfig({ cardColor: e.target.value } as Partial<T>)
           }
           className="h-8 w-full rounded-md border border-slate-200 bg-white"
-          aria-label={`Custom ${label}`}
+          aria-label={`Custom ${label.toLowerCase()} color`}
         />
 
         <div>
--- a/components/common/SurfaceColorSettings.tsx
+++ b/components/common/SurfaceColorSettings.tsx
@@ -78,7 +78,7 @@ export const SurfaceColorSettings = <
               } as Partial<T>)
             }
             className="w-full accent-indigo-600"
-            aria-label="Surface opacity"
+            aria-label={`${label} opacity`}
           />
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


OPS-PIvers:
Latest push (`0fa22ad`) settles this — worth flagging the pattern for whoever reviews next: the automated review's last two rounds directly contradicted each other on the same lines (round 4 said drop "color" from the aria-labels and use `${label}`; round 5 said that was a regression and to restore "color" + `.toLowerCase()`).

Resolved by picking the version that keeps both the dynamic label (fixes the real original bug — hardcoded "surface" instead of the widget's actual `label` prop) and the word "color" (avoids the ambiguous-hex-alone concern), and by reverting the `TypographySettings` custom-color-input nesting that caused a genuine ~8px width regression, rather than pursuing the "unify the ARIA group boundary" structural suggestion further at this point in the PR.

Not fixed, and not planned for this PR: the `aria-description`-for-custom-color suggestion (pre-existing design gap, not a regression) and the "orphaned `useId()`" note (no observable effect — nothing renders that would reference it when the picker is hidden).

validate + build both green on `0fa22ad`. This PR has been through several review rounds now (SettingsLabel retrofit → button types → radiogroup → aria-pressed → dynamic labels → structural grouping → this settling pass) — happy to keep iterating on anything concrete, but flagging that we're now deep into wording bikeshedding on internal settings-panel controls rather than fixing user-facing bugs.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-09

### Summary
Retrofits the three shared settings primitives (`TypographySettings`, `SurfaceColorSettings`, `TextSizePresetSettings`) so their group headings are real ARIA group labels (`SettingsLabel as="span" id` + `role="group"` + `aria-labelledby`) instead of orphaned `<label>`s, adds `type="button"` and `aria-pressed` to the swatch/preset buttons, makes swatch aria-labels use the dynamic `label` prop, and dedupes the local `scaleToPreset` in favor of the shared `presetFromScale`. Clean, tightly-scoped a11y improvement.

### Code Quality
✓ No issues. Verified locally on the branch HEAD: `tsc --noEmit` passes. `presetFromScale` is a real export in `config/widgetAppearance.ts` and `SettingsLabel` supports `id` + `as="span"`, so the imports/props are valid. The `scaleToPreset` → `presetFromScale` swap removes duplicated threshold logic.

### Widget Standards
✓ Not applicable — these are shared settings-panel primitives (back-face), which are exempt from container-query scaling. Normal Tailwind classes are appropriate here.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues — no data-layer or rules changes.

### Regression Risk
✓ Low. Additive ARIA attributes plus a pure-function extraction. `presetFromScale` reproduces the previous `scaleToPreset` thresholds. No behavior change to persisted config.

### Review-thread status
All 19 inline threads on this PR were resolved in follow-up commits before this review. Current HEAD already carries: `type="button"` and `aria-pressed` on every button group; dynamic `aria-label`s with the "color" noun restored and no `label` doubling; `useId()` hoisted to the first statement; and the `w-full` width regression reverted. The reviewer's `role="radiogroup"`/`role="radio"`/`aria-checked` suggestions were deliberately superseded by the `aria-pressed` toggle-button pattern — a valid choice, since a true radiogroup would also require roving-tabindex/arrow-key handling that isn't implemented here; `aria-pressed` exposes selection state to AT without promising keyboard behavior that doesn't exist. Resolving those threads accordingly.

Three optional, non-blocking enhancements remain open (not bugs):
- `TypographySettings` custom `<input type="color">` sits just outside the `role="group"` boundary; `SurfaceColorSettings` keeps its custom picker inside the group. Parallel structure would be tidier.
- With the default `fontColor` (`#334155`) not present in `TEXT_COLOR_PRESETS`, every swatch reports `aria-pressed={false}` until a preset is picked; an `aria-description`/current-value hint on the group would announce the active color.
- `textColorLabelId` is generated by `useId()` even when `showColorPicker` is false (harmless orphan id).

### Missing Tests
No unit tests accompany the a11y attributes. Low concern for additive ARIA on presentational primitives, but a small render test asserting `role="group"` + `aria-pressed` toggling would lock in the behavior.

### Overall
Correct, self-contained a11y improvement; type-check passes and all actionable threads are addressed. The three open items are optional polish.

---

---

PR #2437, opened 2026-08-12
D4: use @/ alias in stress-reference spike import

OPS-PIvers:
## Nightly Unifier — D4 Import Path Convention

**Canonical form:** `@/` alias for all cross-directory imports (repo root alias, configured in `tsconfig.json` `paths` and `vite.config.ts` `resolve.alias`). Relative `'../foo'` is only acceptable for true same-directory/direct-parent-sibling imports.

**Instance unified:** `scripts/spikes/stress-reference/reference.ts` and its `reference.test.ts` each imported `nucleiData` from the sibling `scripts/spikes/stress-detection/` directory via a relative `'../stress-detection/nuclei.json'` escape — the same cross-subdirectory-relative-import anti-pattern already fixed repo-wide in `components/plc/**` and `components/widgets/**`. Both now import via `@/scripts/spikes/stress-detection/nuclei.json`.

This file predates this run (added 2026-08-04) and had never been touched by a D4 sweep before.

**Enforcement:** None added this run — the existing `no-restricted-imports` ESLint rules are scoped to `components/plc/**` and `components/widgets/**` (the two directories where this bug class has recurred repeatedly). `scripts/**` is excluded from ESLint entirely (see `eslint.config.js` `ignores`), so this directory is enforced by convention/review rather than a lint rule. As part of this run we re-verified the existing plc `no-restricted-imports` regex is still fully in sync with the live `PLC_SECTIONS` list and the actual `components/plc/` subdirectory listing — no drift found there.

**Behavior-preservation evidence (mechanical, zero logic change):**
- `@/*` → `./*` is configured identically in `tsconfig.json` and `vite.config.ts`; `vitest.config.ts` merges `vite.config.ts` and does not exclude `scripts/**`, so the alias resolves identically for the app build, `tsc`, and the test run — this file is only ever exercised via its own `reference.test.ts` under vitest (per its own doc comment: "SPIKE / DECISION HARNESS. Not wired into the app... See reference.test.ts").
- `pnpm run lint` (`--max-warnings 0`, app + functions) — clean
- `pnpm run format:check` — clean
- `pnpm run test` — 607 files / 7381 tests passing (targeted re-run of `reference.test.ts` alone: 50/50 passing)
- `pnpm run build` — client bundle + SSR bundle + prerender all succeed

**Risk tag:** mechanical (pure import-path equivalence, two-line diff, no behavior or visual change).

---
_Generated by [Claude Code](https://claude.ai/code/session_01RZ9odAcJi4nR4ZCyXEikHg)_

### DIFF
--- a/scripts/spikes/stress-reference/reference.test.ts
+++ b/scripts/spikes/stress-reference/reference.test.ts
@@ -12,7 +12,7 @@
  */
 import { describe, expect, it } from 'vitest';
 
-import nucleiData from '../stress-detection/nuclei.json';
+import nucleiData from '@/scripts/spikes/stress-detection/nuclei.json';
 import {
   type Lang,
   type StressReading,
--- a/scripts/spikes/stress-reference/reference.ts
+++ b/scripts/spikes/stress-reference/reference.ts
@@ -15,7 +15,7 @@
  * the map and `ˌ` is discarded everywhere below.
  */
 
-import nucleiData from '../stress-detection/nuclei.json';
+import nucleiData from '@/scripts/spikes/stress-detection/nuclei.json';
 
 /** The 244 nucleus-bearing tokens, derived from the model's own vocabulary. */
 const NUCLEUS_TOKENS: readonly string[] = [...nucleiData.nuclei].sort(


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-12

### Summary
Converts two relative imports in `scripts/spikes/stress-reference/` (`'../stress-detection/nuclei.json'`) to the `@/` alias, aligning them with the D4 import-path convention. Two-line diff across `reference.ts` and its test.

### Code Quality
✓ No issues found. Both call sites converted consistently; the alias resolves to the project root as configured in `vite.config.ts` and `tsconfig.json`, so `@/scripts/spikes/stress-detection/nuclei.json` is the correct target path.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found.

### Regression Risk
✓ Low risk — about as low as a diff gets. Pure module-specifier change with no behavioral surface. Does not touch `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, or `functions/src/index.ts`.

The one thing worth checking on an alias conversion is whether every tool that reads these files resolves `@/` identically, since `scripts/` is treated unusually: it's a top-level `ignores` entry in `eslint.config.js` (so nothing there is linted), but it *is* type-checked by `tsc` (no `scripts/` exclude in `tsconfig.json`) and exercised by Vitest, whose config merges `vite.config.ts`'s alias without excluding `scripts/**`. The PR's own routine notes record verifying exactly this before treating the change as safe, which is the right diligence for a directory the tooling handles inconsistently. `nuclei.json` is imported as a default export, so `resolveJsonModule` behavior is unchanged by the specifier form.

### Missing Tests
✓ Adequate — no new behavior to test. `reference.test.ts` is itself one of the two changed files, so the existing spike suite exercises the converted specifier directly; a broken alias would fail at import time.

### Overall
**Merge readiness:** Ready
Trivially safe convention fix, with the `scripts/`-specific tooling differences actually verified rather than assumed. The decision not to add a `no-restricted-imports` rule for `scripts/**` alongside a two-line fix is the right scope call — that's a separate question about whether an unlinted grab-bag directory should get scoped import linting at all, and it's already captured in the routine's notes.

---

---

PR #2439, opened 2026-08-12
fix(widgets): stop Escape leaking from Catalyst set picker to DashboardView

OPS-PIvers:
## Root cause

`components/widgets/Catalyst/CatalystSetPickerPopover.tsx` (rendered from `Dock.tsx`, portalled to `document.body` outside any `.widget` ancestor) had **no local Escape handling at all** — its only dismissal path was `useClickOutside`, which listens for `pointerdown`, never `keydown`. Pressing Escape while the picker is open falls through untouched to `DashboardView`'s global `window`-level Escape handler, which finds no `.widget` ancestor for the popover and falls back to minimizing the topmost z-index widget on the board — an unrelated widget disappearing just from dismissing the set picker.

This is a fresh instance of the recurring "portalled popover Escape doesn't stopPropagation" bug class already fixed 10x elsewhere in this codebase (`ActiveClassChip`, `FolderPickerPopover`, `RandomClassContextButton`, `LiveControl`, `PageStrip`, `SettingsPanel`, etc.) — but a *missing-handler* variant rather than a *missing-stopPropagation* variant.

## Fix

Added a `document`-level `keydown` listener (mirroring `ClassRosterMenu`'s established pattern) that closes the popover and calls `event.stopPropagation()` on Escape, guarded by `isEscapeFromWidgetInput` so it doesn't swallow Escape meant for a widget's own text input.

## Rejected band-aid

None needed — this is the same well-precedented root-cause fix already applied repo-wide for this bug class; there was no narrower symptom-only fix available.

## Regression test

`components/widgets/Catalyst/CatalystSetPickerPopover.test.tsx`:
- **Before fix:** `AssertionError: expected "vi.fn()" to be called at least once` (`onClose` never called on Escape).
- **After fix:** `Test Files 1 passed (1) / Tests 1 passed (1)` — `onClose` fires and the `window`-level keydown spy is never called (propagation stopped).

## Evidence

- `pnpm run validate` (type-check, lint, format-check, root tests 608 files/7382 tests, functions tests 41 files/798 tests, test:counts guard) — all green.

## Risk

**Contained.** Single component, additive `useEffect` mirroring an existing established pattern used by many sibling components; no change to any other widget or shared primitive.

---

---

### DIFF
--- a/components/widgets/Catalyst/CatalystSetPickerPopover.test.tsx
+++ b/components/widgets/Catalyst/CatalystSetPickerPopover.test.tsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, cleanup, fireEvent } from '@testing-library/react';
+import { CatalystSetPickerPopover } from './CatalystSetPickerPopover';
+import { useCatalystSets } from '@/hooks/useCatalystSets';
+import { DEFAULT_GLOBAL_STYLE } from '@/types';
+
+vi.mock('@/hooks/useCatalystSets', () => ({
+  useCatalystSets: vi.fn(),
+}));
+
+afterEach(cleanup);
+
+const anchorRect = {
+  left: 10,
+  right: 50,
+  top: 10,
+  bottom: 40,
+  width: 40,
+  height: 30,
+} as DOMRect;
+
+describe('CatalystSetPickerPopover — Escape does not leak to window-level handlers', () => {
+  it('closes on Escape and stops propagation before it reaches window listeners', () => {
+    (useCatalystSets as unknown as ReturnType<typeof vi.fn>).mockReturnValue({
+      sets: [],
+      loading: false,
+    });
+    const onClose = vi.fn();
+    render(
+      <CatalystSetPickerPopover
+        anchorRect={anchorRect}
+        globalStyle={DEFAULT_GLOBAL_STYLE}
+        onSelectRoutine={vi.fn()}
+        onClose={onClose}
+      />
+    );
+
+    expect(screen.getByText('No sets configured.')).toBeInTheDocument();
+
+    const windowKeydownSpy = vi.fn();
+    window.addEventListener('keydown', windowKeydownSpy);
+
+    try {
+      fireEvent.keyDown(document, { key: 'Escape', bubbles: true });
+      expect(onClose).toHaveBeenCalled();
+      // Without stopPropagation this would reach DashboardView's global window-level handler.
+      expect(windowKeydownSpy).not.toHaveBeenCalled();
+    } finally {
+      window.removeEventListener('keydown', windowKeydownSpy);
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-12

### Summary
Adds a `document`-level Escape handler to `CatalystSetPickerPopover`, which is portalled outside any `.widget` ancestor and previously had no local Escape handling at all (`useClickOutside` listens for `pointerdown` only). Escape fell through to `DashboardView`'s global handler and minimized the topmost widget instead of closing the picker. One `useEffect` plus a regression test. This is the "missing-handler" variant of the recurring portalled-popover bug class — 11th fix in the family.

### Code Quality
✓ No issues found. The handler follows the established idiom exactly: `key !== 'Escape'` early return → `isEscapeFromWidgetInput` guard → `stopPropagation()` → `onClose()`, with a symmetric `removeEventListener` teardown.

I checked the `isEscapeFromWidgetInput` guard's interaction with this specific portal, since a mis-scoped guard would make the new handler dead code. `utils/domHelpers.ts:31-47` returns true only for targets under `[data-widget-portal]` or for form controls under `[data-draggable-window]`. This popover's `createPortal` target carries neither attribute, so in-popover Escapes correctly fall through to the new handler, while Escape typed into a text input elsewhere in a widget still yields precedence to that input. The guard is doing what it should here, not silently swallowing the fix.

One minor note: the effect's dep array is `[onClose]`, so an unmemoized `onClose` at a call site re-registers the listener each render. Functionally harmless (add/remove are balanced and the handler is idempotent), just slightly more churn than necessary — a `useRef` latch or a memoized callback at the parent would avoid it. Not worth changing on its own.

### Widget Standards
✓ No issues found. No new widget, so the 8-file registration checklist doesn't apply, and no registry map changes. No front-face content was touched, so container-query scaling rules aren't in play — the pre-existing hardcoded `style={{ width: 144, height: 188 }}` on the set tiles sits inside a `position: fixed` portal with no `container-type: size` boundary, which is the same documented exclusion the css-scaling journal applies to other portalled popovers. Not introduced or worsened here.

### Admin Config Standards
✓ Not applicable — no `components/admin/` changes.

### Firestore & Security
✓ No issues found. No new collections, no rules changes, no auth surface. `useCatalystSets` is mocked in the test rather than hitting Firestore.

### Regression Risk
✓ Low risk. Self-contained addition to one component; no shared state, registry, or type changes. Does not touch `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, or `functions/src/index.ts`.

### Missing Tests
✓ Adequate for the change. The test asserts both halves that matter — `onClose` fires **and** a window-level spy never sees the event — so it fails against both the pre-fix state and a hypothetical fix that closed the popover without stopping propagation.

One coverage note: the test dispatches on `document`, whose `target` is not an `Element`, so `isEscapeFromWidgetInput` short-circuits at its first guard and the branch is never exercised. A second case dispatching from a focused button *inside* the popover would pin the "guard doesn't swallow in-popover Escapes" property that I had to verify by reading `domHelpers.ts`. Worth adding if this component grows a text input later, since that's when the guard becomes load-bearing.

### Overall
**Merge readiness:** Ready
Correct application of a well-precedented fix pattern, with the guard interaction verified rather than assumed. The routine log correctly flags `RandomGroups.tsx`'s color-picker popover as the remaining live instance of this same variant and deliberately leaves it for a separate PR — right call for keeping this one single-purpose.

---

---

PR #2440, opened 2026-08-12
fix(dashboard): clear pending long-press timer on unmount

OPS-PIvers:
## Root cause

`hooks/useLongPress.ts` schedules a 600ms hold timer in `handlePointerDown` but never cleared it on unmount. This hook backs 4 dock components (`ToolDockItem`, `FolderItem`, `SortableFolderWidget`, `SavedWidgetDockItem`). If the pressed element unmounts mid-press — e.g. an item removed from the dock via real-time Firestore sync, or a parent re-render — with no `pointerup`/`pointercancel` ever reaching the hook, the pending timer still fires `onLongPress` 600ms later against a stale target, triggering edit-mode/widget-library state changes for a press the user can no longer see or intended.

## Fix

Added `useEffect(() => clear, [clear])`, using the hook's existing memoized `clear` callback, so the pending timer is always torn down on unmount.

## Rejected band-aid

Guarding `onLongPress` itself with an `isMountedRef` check (swallow the callback instead of cancelling the timer) — rejected because it leaves the stale timer running instead of tearing it down, and doesn't generalize to the unmount-cleanup pattern already used elsewhere in this codebase for the same class of resource leak.

## Regression test

`hooks/useLongPress.test.ts` (4 tests total, including the new one):
- **Before fix:** `does not fire onLongPress after the owning component unmounts mid-press` failed — `expected "vi.fn()" to not be called at all, but actually been called 1 times`.
- **After fix:** `4 passed (4)`.

## Evidence

- `pnpm run validate` (type-check, lint, format-check, root tests 608 files/7385 tests, functions tests 41 files/798 tests, test:counts guard) — all green.

## Risk

**Contained.** Single hook, additive cleanup effect using an already-memoized callback; no change to call-site behavior while mounted.

---

---

### DIFF
--- a/hooks/useLongPress.test.ts
+++ b/hooks/useLongPress.test.ts
@@ -1,0 +1,59 @@
+import { renderHook } from '@testing-library/react';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { useLongPress } from './useLongPress';
+
+const fakePointerEvent = (x = 0, y = 0) =>
+  ({ clientX: x, clientY: y }) as unknown as React.PointerEvent;
+
+describe('useLongPress', () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('fires onLongPress after the hold delay while still mounted', () => {
+    vi.useFakeTimers();
+    const onLongPress = vi.fn();
+    const { result } = renderHook(() => useLongPress(onLongPress));
+
+    result.current.onPointerDown(fakePointerEvent());
+    vi.advanceTimersByTime(600);
+
+    expect(onLongPress).toHaveBeenCalledTimes(1);
+  });
+
+  it('cancels the pending timer when the pointer moves past the threshold', () => {
+    vi.useFakeTimers();
+    const onLongPress = vi.fn();
+    const { result } = renderHook(() => useLongPress(onLongPress));
+
+    result.current.onPointerDown(fakePointerEvent(0, 0));
+    result.current.onPointerMove(fakePointerEvent(50, 50));
+    vi.advanceTimersByTime(600);
+
+    expect(onLongPress).not.toHaveBeenCalled();
+  });
+
+  it('cancels the pending timer on pointer up', () => {
+    vi.useFakeTimers();
+    const onLongPress = vi.fn();
+    const { result } = renderHook(() => useLongPress(onLongPress));
+
+    result.current.onPointerDown(fakePointerEvent());
+    result.current.onPointerUp();
+    vi.advanceTimersByTime(600);
+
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-12

### Summary
Adds an unmount cleanup to `hooks/useLongPress.ts` so a pending 600ms hold timer is cancelled when the owning element unmounts mid-press — previously no `pointerup`/`pointercancel` would ever arrive in that case, and `onLongPress` fired against a stale target 600ms later. One-line `useEffect(() => clear, [clear])` plus a new 4-test file covering fire, move-cancel, up-cancel, and the unmount case.

### Code Quality
✓ No issues found.

I specifically checked the risk this pattern usually carries — an unstable cleanup dep silently cancelling *live* presses on every re-render. It's safe here: `clear` is `useCallback(..., [])` with an empty dep array and touches only refs, so its identity is stable for the hook's lifetime and the effect's cleanup runs on unmount only. Passing `clear` directly as the effect body (rather than `() => { ... }`) is idiomatic and reads correctly.

This is also a legitimate `useEffect` under CLAUDE.md's "escape hatch, not a default" rule — cancelling a `setTimeout` on unmount is synchronizing with an external system (the timer), not deriving state.

### Widget Standards
✓ Not applicable — a hook, no widget front-face content. No container-query concerns.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. No Firestore, network, or auth surface touched.

### Regression Risk
✓ Low risk. `hooks/useLongPress.ts` backs four dock components (`ToolDockItem`, `FolderItem`, `SortableFolderWidget`, `SavedWidgetDockItem`), so the blast radius is wide — but the change is strictly additive teardown. No existing return value, handler identity, or timing changes, and the only new behavior is "don't fire after unmount," which no caller could have been relying on.

Does not touch `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, or `functions/src/index.ts`.

### Missing Tests
✓ Adequate coverage. The four cases map cleanly onto the hook's four exit paths, and the unmount test is the one that fails without the fix.

One gap worth naming: nothing exercises the `options.disabled` short-circuit or the `options.onPointerDown` passthrough (the dnd-kit listener bridge). Both are pre-existing untested branches rather than something this PR introduces, and the `disabled` path is the one most likely to regress silently — the `handlePointerDown` dep array includes `options`, so an inline object literal at a call site re-creates the callback every render. Not in scope here, but a natural follow-up now that the file has a test harness.

### Overall
**Merge readiness:** Ready
Correct, minimal fix for a real stale-timer leak, with the `clear` stability precondition verified rather than assumed, and a regression test that fails without it.

---

---

PR #2441, opened 2026-08-12
fix(admin): case-insensitive beta-user dedup in GlobalPermissionsManager

OPS-PIvers:
## Root cause

`GlobalPermissionsManager.tsx`'s `addBetaUser` lowercases the new email before storing it (`email.trim().toLowerCase()`), but deduped it against the existing `betaUsers` array via a case-sensitive `.includes()`. A legacy mixed-case entry — written before this normalization existed, or edited directly via the Firestore console — doesn't match the lowercased candidate, so re-adding "the same" email in a different case created a second chip for the same person.

This is the identical bug class already fixed in the sibling components `BetaUsersPanel.tsx` (#2375) and `PresetSubEmailsManager.tsx` (#2389) — `GlobalPermissionsManager.tsx`'s own `addBetaUser` (a separate implementation for the *global* feature-permissions editor) still had the case-sensitive membership check.

## Fix

Changed the membership check to case-insensitive:
```ts
if (!permission.betaUsers.some((e) => e.toLowerCase() === trimmedEmail)) {
```

## Rejected band-aid

Lowercasing existing array entries at render/load time instead of fixing the comparison — rejected because it doesn't fix the actual dedup check (a future write path using the same `.includes()`-style comparison would still misbehave), and it would silently rewrite admin-authored data.

## Regression test

`components/admin/GlobalPermissionsManager.test.tsx`: seeds a `live-session` permission with beta user `Teacher@School.ORG`, then adds `teacher@school.org`.
- **Before fix:** `AssertionError: expected [...] to have a length of 1 but got 2` — duplicate chip created.
- **After fix:** exactly one chip.

## Evidence

- `pnpm run validate` (type-check, lint, format-check, root tests 608 files/7382 tests, functions tests 41 files/798 tests, test:counts guard) — all green.

## Risk

**Contained.** Single component, single comparison change matching an already-established sibling pattern; no change to write-time normalization or data shape.

---

---

### DIFF
--- a/components/admin/GlobalPermissionsManager.test.tsx
+++ b/components/admin/GlobalPermissionsManager.test.tsx
@@ -1,0 +1,97 @@
+import React from 'react';
+import {
+  render,
+  screen,
+  fireEvent,
+  waitFor,
+  cleanup,
+} from '@testing-library/react';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { GlobalPermissionsManager } from './GlobalPermissionsManager';
+
+// Minimal lucide-react stub — avoids loading the full ~25,000-line bundle.
+vi.mock('lucide-react', () => {
+  function icon(name: string) {
+    const Stub = (props: React.HTMLAttributes<HTMLSpanElement>) =>
+      React.createElement('span', { 'data-icon': name, ...props });
+    Stub.displayName = name;
+    return Stub;
+  }
+  const mocks: Record<string, unknown> = {};
+  return new Proxy(mocks, {
+    get(target, prop) {
+      if (prop === '__esModule') return true;
+      if (prop === 'then') return undefined;
+      if (typeof prop === 'string' && !(prop in target)) {
+        target[prop] = icon(prop);
+      }
+      return target[prop as string];
+    },
+  });
+});
+
+vi.mock('@/config/firebase', () => ({ db: {} }));
+
+const existingPermission = {
+  featureId: 'live-session',
+  accessLevel: 'beta',
+  betaUsers: ['Teacher@School.ORG'],
+  enabled: true,
+  buildings: [],
+  config: {},
+};
+
+vi.mock('firebase/firestore', () => ({
+  collection: vi.fn(),
+  doc: vi.fn(),
+  setDoc: vi.fn(),
+  addDoc: vi.fn(),
+  serverTimestamp: vi.fn(),
+  getDocs: vi.fn(() => ({
+    forEach: (cb: (doc: { data: () => unknown }) => void) => {
+      cb({ data: () => existingPermission });
+    },
+  })),
+}));
+
+vi.mock('@/context/useAuth', () => ({
+  useAuth: () => ({
+    user: { email: 'admin@test.com' },
+    appSettings: {},
+    updateAppSettings: vi.fn(),
+  }),
+}));
+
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-12

### Summary
Makes the beta-user membership check in `GlobalPermissionsManager.addBetaUser` case-insensitive, so a legacy mixed-case entry (pre-dating normalization, or written directly via the Firestore console) no longer produces a duplicate chip when the same mailbox is re-added in lowercase. One-line fix plus a new colocated component test. Third instance of a bug class already fixed in `BetaUsersPanel.tsx` (#2375) and `PresetSubEmailsManager.tsx` (#2389).

### Code Quality
✓ No issues found. The fix is minimal and correct: `trimmedEmail` is already lowercased on the line above (`email.trim().toLowerCase()`), so `permission.betaUsers.some((e) => e.toLowerCase() === trimmedEmail)` compares like against like. Matches the sibling components' pattern exactly.

The `lucide-react` Proxy stub in the test is a nice touch for keeping the test light, and it correctly handles the `then` trap (returning `undefined`) so Vitest's module resolution doesn't mistake the mock namespace for a thenable.

### Widget Standards
✓ Not applicable — admin panel, not a widget.

### Admin Config Standards
✓ Correct level. This is the *global* feature-permissions editor (`global_permissions/{featureId}`), which is where beta-user lists belong — not user-level widget settings. No `SchemaDrivenConfigurationPanel` involvement needed for a one-line predicate fix.

### Firestore & Security
✓ No issues found. No new collections or rules.

I verified the access-control side, since a dedup change here could in principle strand a user's beta access: `AuthContext.isBetaUser` (`context/AuthContext.tsx:2290`) already compares with `betaUsers.some((e) => e.toLowerCase() === lowerEmail)`. So a surviving legacy mixed-case entry still grants access correctly, and suppressing the lowercase duplicate does **not** revoke anything. The fix is consistent with the runtime matcher rather than merely cosmetic.

### Regression Risk
✓ Low risk. Doesn't touch `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, or `functions/src/index.ts`. Single predicate change, no signature or data-shape change.

Two minor notes, neither blocking:
1. **Legacy entries stay un-normalized.** The dedup now prevents an admin from normalizing a stale `Teacher@School.ORG` entry by re-adding it lowercase — they'd have to remove the chip and re-add. Given `isBetaUser` is already case-insensitive there's no functional consequence, so this is the right call over a silent normalize-on-write that would mutate stored data as a side effect of a no-op add.
2. **`removeBetaUser` still filters case-sensitively** (`e !== email`). Not a bug in practice — the chip's `onClick` passes the exact stored string back — but it's the one remaining asymmetry if a future call site ever passes a user-typed email to remove.

### Missing Tests
✓ Adequate for the change. The single test targets exactly the regression (mixed-case stored entry + lowercase re-add → still one chip) and would fail against the pre-fix `.includes()`.

Optional strengthening: the test asserts on rendered chip count only. Asserting that `setDoc`/the permission updater was *not* called on the duplicate path would pin the "no write" half of the behavior too, which is the part that actually matters for Firestore churn.

### Overall
**Merge readiness:** Ready
Correct one-line fix for a thrice-recurring bug class, verified consistent with the case-insensitive matcher in `AuthContext`, with a regression test that fails without it.

---

---

PR #2219, opened 2026-07-16
Unify ActivityWall library empty state onto ScaledEmptyState

OPS-PIvers:
## Canonical form
`<ScaledEmptyState icon={IconName} title="..." subtitle="..." />` from `components/common/ScaledEmptyState.tsx` — the shared, container-query-scaled empty state for widget front-face content.

## Instance unified
`components/widgets/ActivityWall/Widget.tsx` — the Activity Library's "No activities yet" hand-rolled empty state (`LibraryBig` icon + title + subtitle in a `bg-slate-50 rounded-xl border` card), a long-standing **VISUAL-RISK** backlog item (re-confirmed as recently as run 27) that was previously blocked on a real container-scoping issue: `ScaledEmptyState`'s `cqmin` sizing resolves against the nearest ancestor with `container-type: size`, which today is only established once at the whole-widget level. This empty state lives in a sub-panel shorter than the full widget (a header above it eats vertical space), so converting without a fix would size the icon/text off the taller outer widget box, not the actual available space — a real, visible size mismatch, not a cosmetic one.

**Fix applied**: added a `container-type: size` boundary scoped *only* to the empty-state branch's own wrapper div (not shared with the populated-list branch, which is a separate ternary arm) — so the previously-documented risk of "rescoping affecting other cqmin content in the same panel" doesn't apply here; the populated view's sizing is untouched. This same nested-container-boundary pattern is already established elsewhere in the codebase (`Checklist/Widget.tsx`, `MaterialsWidget/index.tsx`, `Stations/components/StationCard.tsx`, `ConceptWeb/Widget.tsx`), independently verified by grep before shipping.

The `bg-slate-50 rounded-xl border border-slate-200` card background is preserved via the outer wrapper div's className (not `ScaledEmptyState`'s own, which stays transparent by default) — `ScaledEmptyState` renders fully inside it at `h-full w-full`.

## Visual-preservation notes (disclosed honestly, not claimed as pixel-perfect)
- Icon, icon color (`text-brand-blue-primary`), title color (`text-slate-800`), subtitle color (`text-slate-500`), and the card background/border are all preserved via `iconClassName`/`titleClassName`/`subtitleClassName` overrides and the wrapper div.
[...]

PR #2395, opened 2026-08-05
feat(ai): move Gemini to Vertex AI, update model IDs (+ terms audit)

OPS-PIvers:
Started as an audit of which Google AI surface the app calls and whether student routes could reach it; the audit's recommendation was approved, so this PR now also carries the migration. Full findings in `docs/gemini-api-terms-audit.md`.

## ⚠️ Deploy prerequisites — AI calls fail until these are done

Both are console operations, deliberately not performed by this PR:

1. Enable `aiplatform.googleapis.com` on the `spartboard` project.
2. Grant `roles/aiplatform.user` to the Cloud Functions runtime service account (the default one — `firebase.json` sets no `serviceAccount` override).

~~Separately, **existing Firestore model overrides are not covered by any code change.** If `global_permissions/gemini-functions` already has a preview model ID written into it, it still wins over the new defaults. Needs a one-time check.~~

**Updated — no manual Firestore check is needed.** `normalizeModelName()` now rejects deprecated ids (`gemini-1.*`, `gemini-2.0-*`, and any `-preview` id, including the old `gemini-3-flash-preview` / `gemini-3.1-flash-lite-preview` defaults) and returns `undefined`, which every caller already treats as "use the default". A stale override in `global_permissions/gemini-functions` therefore self-heals on the next call rather than winning over the new defaults, and the admin picker shows an inline "deprecated model" error on the stored value next time someone opens that panel. The only overrides still worth eyes-on are well-formed, non-deprecated ids that are merely semantically stale — those are honored as written, by design.

Deleting the now-unused `GEMINI_API_KEY` secret from Secret Manager is optional cleanup — do it only after the deploy is verified.

## Why

The app called the **Gemini Developer API** (`generativelanguage.googleapis.com`) with an API key. Those Additional Terms carry an **Age Requirements** section — a top-level clause, above the Unpaid/Paid split, so it binds on both tiers:

> You must be 18 years of age or older to use the APIs. You also will not use the Services as part of a website, application, or other service (collectively, "API Clients") that is directed towards or is likely to be accessed by individuals under the age of 18.

SpartBoard is a K-12 app with student login routes. The audit confirmed no student-submitted content ever reaches Gemini — all four callables require a verified email claim that student tokens structurally never carry — but the clause attaches to the application, not to which code path calls the API, so the separation doesn't resolve it.

[...]

PR #2445, opened 2026-08-13
fix(activitywall): unify "no responses" empty state to ScaledEmptyState

OPS-PIvers:
## Nightly Unifier — D1 Widget Empty States (run 55)

**Canonical form:** `<ScaledEmptyState icon={...} title="..." subtitle="..." />` from `components/common/ScaledEmptyState.tsx`.

**Instance unified:** `components/widgets/ActivityWall/Widget.tsx` — the "Responses will appear here after participants submit." hand-rolled empty state, one of 5 standing VISUAL-RISK backlog items blocked since run 26 (PR #2143 was closed unmerged after review caught a real regression: `ScaledEmptyState`'s `cqmin` icon sizing resolved against the whole-widget container boundary instead of the smaller submissions sub-panel, causing the icon to roughly double in size).

**Fix:** The empty state, the word-cloud/rows branch, and the photo-grid branch are three mutually-exclusive arms of the same ternary under one shared `bg-slate-50` wrapper — only one ever mounts. Wrapping only the empty arm's own `<div>` in `style={{ containerType: 'size' }}` establishes an isolated `cqmin` boundary that cannot affect the populated siblings, since they never coexist in the DOM. This is the identical technique already shipped and proven safe for the sibling "No Activities Yet" state (`:1561`, PR #2219, run 35) in the same file.

Title/subtitle copy ("No Responses Yet" / "They'll appear here after participants submit.") splits the original single sentence, resolving a prior run's "needs a human to pick title text" note. Updated the one test asserting the old sentence verbatim.

**Enforcement:** N/A — this resolves a structural blocker on an existing shared component/technique rather than introducing a new pattern; no new enforcement needed.

**Behavior-preservation evidence:**
- `pnpm exec tsc --noEmit` → exit 0
- `pnpm run lint` (app + functions) → exit 0
- `pnpm run format:check` → exit 0
- `pnpm run test` → 614 test files / 7402 tests passed
- `pnpm -C functions test` → 41/41 files, 803 tests passed
- `pnpm run build` (client + SSR + prerender) → exit 0
- Independently re-verified by the orchestrator in-worktree (structural diff review confirming the ternary-exclusivity argument, comparison against the shipped `:1561` sibling pattern, and `ScaledEmptyState` prop-signature check) before opening this PR.

**Risk tag:** `visual-risk` — flagging for a rendered-preview eyes-on check despite the structural proof, consistent with how the sibling `:1561` fix (PR #2219) was tagged, since this closes out a backlog item with a documented prior regression.

**Also includes:** a small `docs/routines/unifier.md` update flipping this backlog row to SHIPPED and logging the run — the orchestrator will reconcile this with the rest of tonight's memory-doc update in a separate dedicated PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JiLP8NERPUhPEEsgzUqtSZ)_


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-13

### Summary
Converts `ActivityWall/Widget.tsx`'s hand-rolled "Responses will appear here" empty state to `ScaledEmptyState`, wrapping it in a scoped `container-type: size` boundary to resolve the container-scope regression that closed PR #2143 back at run 26. Splits the copy into a title ("No Responses Yet") and subtitle, and updates the one assertion in `Widget.test.tsx`. Two files.

### Code Quality
The conversion itself is correct — `ScaledEmptyState` is already imported (`:40`), `MessageSquare` is still referenced so no import goes stale, and `iconClassName`/`titleClassName`/`subtitleClassName` are all real props on the component (`components/common/ScaledEmptyState.tsx:4-13`).

**One finding — the added comment block is 10 lines, double its own precedent.** This file's sibling empty state at `:1592` (the run-35 `:1561` fix this change is modeled on) carries a **5-line** comment saying the same thing. The new block reproduces those 5 lines nearly verbatim and then adds 5 more arguing why the boundary is safe ("Safe because this div only exists in the `length === 0` branch… so their sizing is untouched"). That second half is the review argument for the change rather than something a future reader needs in order to not break it — the category the repo's "one short line max" convention targets, and the exact nit that's been raised repeatedly on #2395 this month. Suggest trimming to match the sibling at `:1592` so the two states read identically; the mutual-exclusivity argument is already captured in the unifier log and this PR description.

### Widget Standards
✓ Structurally sound. The core claim — that a `container-type: size` boundary on only the empty arm cannot affect the populated arms — holds: the empty / word-cloud-text / photo-grid branches are three arms of one ternary, so at most one is ever mounted, and the new `<div>` exists only inside the `length === 0` arm. This is genuinely the same shape as the already-shipped `:1561` fix rather than a superficial resemblance.

⚠️ **The scope fix does not by itself preserve the icon's rendered size — worth measuring in the preview, since icon size is precisely what sank #2143.** The two sets of constants differ by more than 2×:

| | Icon | Subtitle |
| --- | --- | --- |
| Before (hand-rolled, whole-widget scope) | `min(24px, 7cqmin)` | `min(11px, 3.8cqmin)` |
| After (`ScaledEmptyState`, scoped) | `min(48px, 15cqmin)` | `min(12px, 3cqmin)` |

The new boundary shrinks the base that `cqmin` resolves against, which pushes size *down*; the doubled coefficient and doubled cap push it *up*. Ignoring the caps, the icon only stays at-or-below its previous size when the scoped region's smaller dimension is **≤ ~47%** of the widget's smaller dimension (`15·I ≤ 7·W`). At the default 520×420 (`config/widgetDefaults.ts:143`) that means the responses panel would need to be ≲196px tall — plausible but not obviously true once the header and action bar are accounted for. So the concrete thing for the rendered-preview check is: **does the `MessageSquare` icon render at roughly its old 24px, or noticeably larger?** The structural argument in the PR description is sound but answers scoping, not sizing.

Also expected and consistent with prior conversions: `ScaledEmptyState` hardcodes `uppercase tracking-widest`, so the title renders as "NO RESPONSES YET" — the same accepted canonical casing delta noted for RandomGroups/WorkSymbols/`:1561`.

### Admin Config Standards

### Firestore & Security
✓ No issues found — presentational only; no collections, rules, or auth paths involved.

### Regression Risk
Low-to-moderate, and correctly self-tagged visual-risk. Confined to one branch of one ternary in one widget; no registry, types, context, or Cloud Function changes. The residual risk is entirely visual (icon scale above), which is why the draft + preview gate is the right call. The other four VISUAL-RISK backlog items were correctly left alone — they're modal overlays without the mutually-exclusive-ternary precondition this fix depends on.

### Missing Tests
✓ Adequate for the behavior change. `Widget.test.tsx:246-249` is updated to assert both the new title and the new subtitle rather than just swapping the string, which is slightly better than the minimum. The sizing question above isn't jsdom-testable (no layout, so `cqmin` never resolves) — a rendered preview is the only instrument for it, which is what the visual-risk tag is for.

### Overall
The structural fix is real and well-reasoned — recognizing this shared the solved `:1561` sibling's precondition is the right insight, and declining the other four items for lacking it shows the reasoning was applied rather than pattern-matched. Two things before merge: (1) trim the 10-line comment to match the 5-line sibling at `:1592`, and (2) in the preview, confirm the icon renders near its previous 24px rather than up toward the new 48px cap — that specific number is what closed #2143.

---

---

OPS-PIvers:
Good catch — confirmed this was a real regression, not just a theoretical risk. I read `ScaledEmptyState.tsx` directly and did the math against the actual widget structure (the response-monitor header above this panel — title/prompt, sync banner, action-button row — eats enough vertical space that the sub-panel's smaller dimension at the default 520×420 widget size lands around 280-320px, well past the ~47% threshold where the icon would've stayed at-or-below its old size). So the icon really would have rendered ~42-48px instead of 24px — the same class of regression that closed #2143.

Pushed a fix (`ba4cd1f`):
- Added an optional `iconSize` prop to `ScaledEmptyState` (default unchanged at `min(48px, 15cqmin)`, so all other 40+ existing consumers are unaffected).
- This call site now passes the exact original pre-conversion values, `min(24px, 7cqmin)`.
- This isn't just "closer" — it's a provable upper bound: the sub-panel's smaller dimension can never exceed the whole widget's smaller dimension (it's strictly contained within it), so reusing the identical formula guarantees the new rendered icon size is always ≤ the original 24px, regardless of how the header's height varies across widget resizes or activity states.
- Also trimmed the comment back to the sibling `:1592`'s 5-line form, dropping the review-argument half you flagged.

`pnpm run validate` (614 root test files/7402 tests + 41/41 functions/803 tests) and `pnpm run build` both green on the updated commit.

---

PR #2446, opened 2026-08-13
fix(drawing): retrofit orphaned Background group heading to SettingsLabel span pattern

OPS-PIvers:
## Nightly Unifier — D3 Settings Panel Label Primitives (run 55)

**Canonical form:** `<SettingsLabel as="span" id="unique-id">Group Heading</SettingsLabel>` + `role="group" aria-labelledby="unique-id"` on the sibling controls container — the group-heading form of `SettingsLabel` established over runs 26-53 for labels that head a group of controls rather than pairing 1:1 with a single input.

**Instance unified:** `components/widgets/DrawingWidget/Settings.tsx:140-157` — the "Background" group heading. This file already had the *correct* pattern for its "Color Presets" group (lines 84-93, shipped run 32) but the "Background" group 20 lines later was left on the old shape: a bare `<SettingsLabel>` (rendered as an orphaned `<label>` with no `htmlFor`) duplicated by a separate `aria-label="Background template"` on the sibling `role="group"` container. Converted to `as="span" id={`drawing-background-label-${widget.id}`}` (instance-scoped, matching this file's own established convention for the Color Presets fix) and pointed the group's `aria-labelledby` at it, removing the now-redundant `aria-label`.

Found via a targeted sweep for `role="group"` + `aria-label` (instead of `aria-labelledby`) pairs inside `Settings.tsx`/`*ConfigurationPanel.tsx` files — a previously-undocumented instance, not on any prior run's backlog.

**Investigated and correctly left untouched** (open candidates carried over from run 53/PR #2424):
- `components/admin/MathToolsConfigurationPanel.tsx:141,143,147` — confirmed these are CSS-grid `<div>` column headers over independent per-row content, not a real `<table>` and not a flat sibling-control group; neither `SettingsLabel` form fits. Needs a structural change (real `<table>`/`<th scope="col">`) outside this routine's mechanical scope — left as a backlog item for a future dedicated pass.
- `components/widgets/Stations/Settings.tsx:232,271` — heading over heterogeneous composite rows / a single conditional block, not a flat button/checkbox group. Correctly out of scope.
- `components/widgets/NumberLine/Settings.tsx:106,264,345` — sections whose sub-fields already have their own individual `<label htmlFor>` pairings. Correctly out of scope.

**Enforcement:** N/A — applying an existing, already-enforced-by-convention shared component; the "second instance in the same file" class this closes is caught by review discipline (reading the whole file, not just historical diffs), not automatable via lint.

**Behavior-preservation evidence:** `SettingsLabel.tsx`'s `combinedClasses` confirmed identical for `as="label"` vs `as="span"` (read directly, not trusted from precedent citation) — zero visual delta.
- `pnpm exec tsc --noEmit` → exit 0
- `pnpm run lint` (app + functions) → exit 0
- `pnpm run format:check` → exit 0
- `pnpm run test` → 614 test files / 7402 tests passed
- `pnpm -C functions test` → 41/41 files, 803 tests passed
- `pnpm run build` (client + SSR + prerender) → exit 0
- All independently re-run by the orchestrator directly in the worktree after the subagent hit the stuck-background-wait pattern partway through its own verification; orchestrator reviewed the diff, the surrounding file (confirming "Brush Thickness"/"Shape Fill" are genuine 1:1 pairings correctly left as `as="label"`), and the `SettingsLabel.tsx` source before running its own full `pnpm run validate` + `pnpm run build` (both green) and shipping.

**Risk tag:** `mechanical` — pure equivalence, single-file, already-proven pattern.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JiLP8NERPUhPEEsgzUqtSZ)_

### DIFF
--- a/components/widgets/DrawingWidget/Settings.tsx
+++ b/components/widgets/DrawingWidget/Settings.tsx
@@ -137,7 +137,13 @@ export const DrawingSettings: React.FC<{ widget: WidgetData }> = ({
       </div>
 
       <div>
-        <SettingsLabel icon={LayoutGrid}>Background</SettingsLabel>
+        <SettingsLabel
+          icon={LayoutGrid}
+          as="span"
+          id={`drawing-background-label-${widget.id}`}
+        >
+          Background
+        </SettingsLabel>
         {/* Toggle-button group (not radiogroup) — selecting a background is a
             mode toggle. The button + aria-pressed pattern gives us native
             Tab/Space/Enter without the roving-tabindex machinery that a true
@@ -145,7 +151,7 @@ export const DrawingSettings: React.FC<{ widget: WidgetData }> = ({
             AnnotationOverlay.tsx (PR 1685 round-1 fix). */}
         <div
           role="group"
-          aria-label="Background template"
+          aria-labelledby={`drawing-background-label-${widget.id}`}
           className="flex gap-2 px-2"
         >
           {BACKGROUND_OPTIONS.map(({ value, label }) => {


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-13

### Summary
Retrofits the "Background" group heading in `DrawingWidget/Settings.tsx` to the repo's established group-label pattern: `SettingsLabel` gains `as="span"` and a widget-scoped `id`, and the sibling `role="group"` container swaps its `aria-label="Background template"` for `aria-labelledby` pointing at that id. Nine lines changed in one file.

### Code Quality
✓ No issues found. The `as`/`id` props are real and documented for exactly this case — `components/common/SettingsLabel.tsx:9-11` carries the comment *"'span' renders a heading for a group of sibling controls… a bare `<label>` would be orphaned; pair with `role="group"` + `aria-labelledby={id}`"*. This is that pairing, verbatim.

The `id` is interpolated with `widget.id` (`drawing-background-label-${widget.id}`), which is necessary rather than decorative: multiple Drawing widgets can be open on one dashboard, and a static id would produce duplicate DOM ids and an `aria-labelledby` that resolves to whichever mounted first. Matches the convention used by the sibling "Color Presets" group in this same file.

Removing the now-redundant `aria-label` is correct — with both present, `aria-labelledby` wins and the `aria-label` is dead weight that can silently drift out of sync with the visible text.

### Widget Standards
✓ Not applicable in the container-query sense. This is a back-face settings panel, where CLAUDE.md explicitly exempts scaling ("Settings panels do NOT need container query scaling — use normal Tailwind classes"). No front-face content touched.

### Admin Config Standards
✓ Not applicable — user-level widget settings, correctly not an admin-level concern. Background template choice is per-widget state, so this is the right level.

### Firestore & Security
✓ No issues found — presentational/a11y only, no persistence or permission surface.

### Regression Risk
✓ Low. One file, no behavior change to the toggle group itself (the `aria-pressed` button pattern and its explanatory comment are untouched).

One deliberate behavior change worth a human eye, since it's the only user-observable delta: the group's accessible name goes from **"Background template"** to **"Background"**, because it now derives from the visible heading text rather than a separate string. That's the intended tradeoff of this pattern — the accessible name matching the visible label is generally better for voice control ("click Background") and prevents label drift — but it is a small loss of specificity for a screen-reader user scanning groups. Worth a conscious accept rather than a silent one, and consistent with how the other retrofits in this series landed.

### Missing Tests
Gap, non-blocking. No test file exists for `components/widgets/DrawingWidget/Settings.tsx`, so nothing pins the `aria-labelledby` wiring. The analogous `tests/components/common/TypographySettings.test.tsx` exists for the shared primitive, which covers `SettingsLabel`'s own rendering — so the props are tested even though this call site isn't. Given ~20 files have now had this same retrofit, a single test asserting "group is labelled by its heading" on one representative widget would guard the pattern more cheaply than per-file tests. Not something to block this PR on.

### Overall
Correct, idiomatic, and matches both the shared primitive's documented contract and this file's own sibling group. The only judgment call is the accessible-name change from "Background template" to "Background" — intended, but confirm it reads well in a screen-reader pass alongside the other retrofitted groups.

---

---

OPS-PIvers:
Thanks for the review. Confirming the accessible-name change ("Background template" → "Background") is a conscious accept, not an oversight: it's the intended effect of this pattern — the group's accessible name deriving from its own visible heading text — and matches how the ~20 other files in this group-heading retrofit series have landed (the name always tracks the heading text, never a separately-authored `aria-label`). No plans for a dedicated screen-reader pass on this specific change; it's consistent with the established series.

On the missing-test gap: agreed it's non-blocking, and I'll leave the "one representative test for the pattern" suggestion as a backlog idea rather than adding it to this PR's scope.

---

PR #2450, opened 2026-08-14
fix(a11y): pair orphaned group-heading labels with their radiogroups

OPS-PIvers:
## Summary

Nightly consistency run (D3 — Settings Panel Label Primitives). Part of the long-running group-heading `SettingsLabel` retrofit series (runs 26-55).

**Canonical form:** `<SettingsLabel as="span" id="unique-id">Heading</SettingsLabel>` + `role="group"|"radiogroup" aria-labelledby="unique-id"` on the sibling-control container — replacing an orphaned `<label>` (default `as="label"`, no `htmlFor`) whose radiogroup carried its own redundant `aria-label` duplicating the visible heading text instead of pointing at it via `aria-labelledby`. Matches the convention already established in 12+ other radiogroup instances repo-wide (`TextSizePresetSettings`, `SurfaceColorSettings`, `TimeTool/Settings.tsx` ×6, `SpecialistSchedule/Settings.tsx`, `MusicWidget/Settings.tsx` ×2, and the two already-fixed groups within `TimeToolConfigurationPanel.tsx` itself).

**Instances unified (5, across 2 files):**
- `components/widgets/Scoreboard/Settings.tsx` — "Layout" radiogroup (Cards/List toggle)
- `components/admin/TimeToolConfigurationPanel.tsx` — "Default Mode", "Display Style", "Number Style", "Default Alert Sound"

**Enforcement:** None added — this fix is a manual retrofit following the established repo-wide convention (no automated lint rule exists for this shape; the pattern has been applied consistently by hand across ~20 files in this series).

**Behavior/visual-preservation evidence:**
- Read `components/common/SettingsLabel.tsx` source directly: `combinedClasses` is computed once and applied identically in both the `<span>` and `<label>` render branches — only the DOM tag differs. Zero visual delta.
- The `aria-label` → `aria-labelledby` swap changes only accessible-name computation, not layout/paint.
- `tsc --noEmit`: 0 errors. `eslint . --max-warnings 0`: clean on changed files (also independently re-verified by the orchestrator). `prettier --check`: clean on changed files (independently re-verified). Full `pnpm run test`: 614 root test files / 7402 tests passed; `pnpm -C functions run test`: 41 files / 803 tests passed. `pnpm run build`: succeeded (client + SSR + prerender). Orchestrator additionally ran a targeted `vitest` pass covering `components/widgets/Scoreboard/**` and `components/admin/TimeToolConfigurationPanel*` (200 files / 1530 tests, all passing).

**Risk tag: mechanical** (pure DOM-tag + ARIA-attribute equivalence, no visual or behavioral change).

## Deferred to backlog / left alone
- `DrawingWidget/Settings.tsx:115,190`, `RevealGrid/Settings.tsx:624,648` — each heads exactly one control, not a group; not a fit.
- `ClockConfigurationPanel.tsx:123` "Default Theme Color" — heterogeneous composite row (swatch + text input + button), same declined shape as prior `Stations/Settings.tsx` case.
- `SurfaceColorSettings.tsx` inner `aria-label` — nested sub-radiogroup with no visible heading of its own; `aria-label` is correct there, not a mismatch.
- Re-confirmed the 3 candidates already declined in a prior run (`MathToolsConfigurationPanel.tsx` grid headers, `Stations/Settings.tsx` composite rows, `NumberLine/Settings.tsx` individually-labeled sub-fields) are still correctly out of mechanical scope.
- Standing NEEDS-REVIEW items (undocumented 3rd label-weight tier; `role="group"`+`aria-pressed` vs. full `role="radiogroup"` architecture question) intentionally left untouched — both require a human decision, not a mechanical fix.

## Test plan
- [x] `tsc --noEmit`
- [x] `eslint . --max-warnings 0`
- [x] `prettier --check`
- [x] `pnpm run test` (root)
- [x] `pnpm -C functions run test`
- [x] `pnpm run build`
- [ ] Optional: eyes-on check that Scoreboard/TimeTool admin config settings panels render identically (change is DOM-tag/ARIA-only, not expected to affect rendering)

---
_Generated by [Claude Code](https://claude.ai/code/session_01KsQHVe9MfQ9BDFzJbZ7NqQ)_

### DIFF
--- a/components/admin/TimeToolConfigurationPanel.tsx
+++ b/components/admin/TimeToolConfigurationPanel.tsx
@@ -83,6 +83,10 @@ export const TimeToolConfigurationPanel: React.FC<
     useBuildingSelection(BUILDINGS);
   const accentColorLabelId = useId();
   const trafficLightColorLabelId = useId();
+  const defaultModeLabelId = useId();
+  const displayStyleLabelId = useId();
+  const numberStyleLabelId = useId();
+  const defaultAlertSoundLabelId = useId();
 
   const buildingDefaults = config.buildingDefaults ?? {};
   const currentBuildingConfig: BuildingTimeToolDefaults = buildingDefaults[
@@ -184,11 +188,13 @@ export const TimeToolConfigurationPanel: React.FC<
 
         {/* Default Mode */}
         <div>
-          <SettingsLabel className="mb-1">Default Mode</SettingsLabel>
+          <SettingsLabel as="span" id={defaultModeLabelId} className="mb-1">
+            Default Mode
+          </SettingsLabel>
           <div
             className="flex gap-1.5"
             role="radiogroup"
-            aria-label="Default Mode"
+            aria-labelledby={defaultModeLabelId}
           >
             {MODES.map(({ value, label }) => (
               <button
@@ -208,11 +214,13 @@ export const TimeToolConfigurationPanel: React.FC<
 
         {/* Display Style */}
         <div>
-          <SettingsLabel className="mb-1">Display Style</SettingsLabel>
+          <SettingsLabel as="span" id={displayStyleLabelId} className="mb-1">
+            Display Style
+          </SettingsLabel>
           <div
             className="flex gap-1.5"
             role="radiogroup"
-            aria-label="Display Style"
+            aria-labelledby={displayStyleLabelId}
           >
             {VISUAL_TYPES.map(({ value, label }) => (
               <button
@@ -234,11 +242,13 @@ export const TimeToolConfigurationPanel: React.FC<
 
         {/* Number Style */}
         <div>
-          <SettingsLabel className="mb-1">Number Style</SettingsLabel>
+          <SettingsLabel as="span" id={numberStyleLabelId} className="mb-1">
+            Number Style
+          </SettingsLabel>
           <div
             className="flex gap-1.5"
             role="radiogroup"
-            aria-label="Number Style"
+            aria-labelledby={numberStyleLabelId}
           >
             {CLOCK_STYLES.map(({ value, label }) => (
               <button
@@ -260,11 +270,17 @@ export const TimeToolConfigurationPanel: React.FC<
 
         {/* Alert Sound */}
         <div>
-          <SettingsLabel className="mb-1">Default Alert Sound</SettingsLabel>
+          <SettingsLabel
+            as="span"
+            id={defaultAlertSoundLabelId}
+            className="mb-1"
+          >
+            Default Alert Sound
+          </SettingsLabel>
           <div
             className="flex gap-1.5"
             role="radiogroup"
-            aria-label="Default Alert Sound"
+            aria-labelledby={defaultAlertSoundLabelId}
           >
             {SOUNDS.map((sound) => (
               <button
--- a/components/widgets/Scoreboard/Settings.tsx
+++ b/components/widgets/Scoreboard/Settings.tsx
@@ -176,14 +176,18 @@ export const ScoreboardSettings: React.FC<{ widget: WidgetData }> = ({
     addToast('All scores reset to 0', 'info');
   };
 
+  const layoutLabelId = `scoreboard-layout-label-${widget.id}`;
+
   return (
     <div className="space-y-6">
       <div>
-        <SettingsLabel>Layout</SettingsLabel>
+        <SettingsLabel as="span" id={layoutLabelId}>
+          Layout
+        </SettingsLabel>
         <div
           className="flex items-center bg-slate-200/80 rounded-lg p-0.5"
           role="radiogroup"
-          aria-label="Scoreboard layout"
+          aria-labelledby={layoutLabelId}
         >
           <button
             type="button"
--- a/components/widgets/Scoreboard/Widget.test.tsx
+++ b/components/widgets/Scoreboard/Widget.test.tsx
@@ -400,6 +400,25 @@ describe('ScoreboardSettings', () => {
     mockAddToast.mockClear();
   });
 
+  // Pins the aria-labelledby ↔ SettingsLabel id pairing; dropping the id leaves the group unnamed with nothing else failing.
+  it('names the layout radiogroup from its visible heading', () => {
+    const widget: WidgetData = {
+      id: 'scoreboard-id',
+      type: 'scoreboard',
+      config: { teams: [] } as ScoreboardConfig,
+      x: 0,
+      y: 0,
+      w: 100,
+      h: 100,
+      z: 1,
+      flipped: true,
+    };
+
+    render(<ScoreboardSettings widget={widget} />);
+
+    expect(screen.getByRole('radiogroup')).toHaveAccessibleName('Layout');
+  });
+
   it('imports groups from random widget', () => {
     const widget: WidgetData = {
       id: 'scoreboard-id',
--- a/tests/components/admin/TimeToolConfigurationPanel.labels.test.tsx
+++ b/tests/components/admin/TimeToolConfigurationPanel.labels.test.tsx
@@ -1,0 +1,37 @@
+// Pins each radiogroup's aria-labelledby ↔ SettingsLabel id pairing; dropping an id leaves the group unnamed with nothing else failing.
+
+import React from 'react';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, cleanup } from '@testing-library/react';
+import type { TimeToolGlobalConfig } from '@/types';
+
+vi.mock('@/hooks/useAdminBuildings', () => ({
+  useAdminBuildings: () => [{ id: 'b1', name: 'Test School' }],
+}));
+
+import { TimeToolConfigurationPanel } from '@/components/admin/TimeToolConfigurationPanel';
+
+afterEach(cleanup);
+
+const RADIOGROUP_NAMES = [
+  'Default Mode',
+  'Display Style',
+  'Number Style',
+  'Default Alert Sound',
+];
+
+describe('TimeToolConfigurationPanel — radiogroup accessible names', () => {
+  it.each(RADIOGROUP_NAMES)(
+    'names the %s radiogroup from its visible heading',
+    (name) => {
+      render(
+        <TimeToolConfigurationPanel
+          config={{} as TimeToolGlobalConfig}
+          onChange={vi.fn()}
+        />
+      );
+
+      expect(screen.getByRole('radiogroup', { name })).toBeInTheDocument();
+    }
+  );
+});


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-14

### Summary
Pairs 5 orphaned group-heading labels with their `role="radiogroup"` containers across `admin/TimeToolConfigurationPanel.tsx` (Default Mode, Display Style, Number Style, Default Alert Sound) and `widgets/Scoreboard/Settings.tsx` (Layout), swapping `aria-label` for `aria-labelledby` pointing at a `SettingsLabel as="span" id=...`. Continues the multi-run D3 retrofit series. Correct and low-risk.

### Code Quality
✓ No issues found. Verified `SettingsLabel` (`components/common/SettingsLabel.tsx`) does support both `as="span"` and `id`, and that `combinedClasses` is computed identically for the `span` and `label` branches — so this is a pure a11y change with **no visual delta**. `className="mb-1"` is preserved verbatim at every call site.

Two minor notes:
- **id-generation is inconsistent between the two files.** `TimeToolConfigurationPanel` uses `useId()` (matching the `accentColorLabelId` / `trafficLightColorLabelId` pair already in that file); `Scoreboard/Settings.tsx` uses a template literal, `` `scoreboard-layout-label-${widget.id}` ``. Both are collision-free and correct — `widget.id` is unique per instance — but `useId()` is the React 19 idiom and is what the sibling file in this same diff uses. Worth standardizing as the series wraps up.
- Replacing `aria-label="Scoreboard layout"` with `aria-labelledby` pointing at the visible text changes the accessible name from *"Scoreboard layout"* to *"Layout"*. That's the intended direction (accessible name now matches visible text, satisfying WCAG 2.5.3), just worth stating explicitly since it is a user-observable change for screen-reader users.

### Widget Standards
✓ Not applicable. `Scoreboard/Settings.tsx` is a back-face settings panel, which per CLAUDE.md correctly uses normal Tailwind classes rather than container-query units. No front-face content touched.

### Admin Config Standards
✓ No issues found. `TimeToolConfigurationPanel` is Admin Settings → Feature Permissions → Widget Config, and the change is confined to label wiring — no setting moved between admin and user level, no schema change.

### Firestore & Security

### Regression Risk
✓ Low. No changes to `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, or `functions/`. Independently verified `pnpm run type-check` passes clean on this branch.

One correction to the run log's scope claim, so the backlog doesn't close prematurely: the `docs/routines/unifier.md` entry for this run describes it as *"closing the last known `aria-label`-instead-of-`aria-labelledby` mismatches."* A repo-wide sweep finds one remaining instance of the same class — **`components/settingsModal/sections/DockSection.tsx:69`**, where a visible `<h3>` reading *"Position"* sits directly above a `role="radiogroup"` whose `aria-label` resolves to *"Dock Position"*. Visible text and accessible name actually diverge there, which is the WCAG 2.5.3 (Label in Name) problem this series exists to fix. It's a different shape (an `<h3>`, not an orphaned `<label>`/`SettingsLabel`), so it's defensible to scope it out of the D3 series — but it should be an explicit decision rather than an implicit "none left." The other 5 `role="radiogroup"` + `aria-label` hits (`SurfaceColorSettings`, `LibraryToolbar`, `PublishScoresModal`, `WidgetBuildingToggle`, `PlcAssignmentConfigModal`) were checked individually and are all legitimate — no visible sibling heading, so `aria-label` is the correct mechanism there.

### Missing Tests
⚠️ Gap worth naming. Neither changed file has coverage for the new wiring: `Scoreboard/` has a `Widget.test.tsx` but no `Settings` test, and `TimeToolConfigurationPanel.tsx` has no test file at all. `aria-labelledby` ↔ `id` pairing is exactly the kind of thing that silently rots — a later refactor dropping the `id` prop or renaming the variable leaves a group with no accessible name, and nothing fails. A single `toHaveAccessibleName()` assertion per group would pin all 5. Not blocking for an a11y-only diff, but this series is now ~6 runs deep with no test anchoring any of it.

### Overall
Correct, well-scoped, visually inert, and type-clean. The only follow-ups are the `useId()` consistency nit and an explicit decision on `DockSection.tsx:69` before declaring the dimension closed.

---

---

OPS-PIvers:
Thanks for the review. Responses:

**`useId()` vs. `widget.id` interpolation:** intentional, not an inconsistency — this is the established convention split used throughout the whole group-heading retrofit series: singleton admin panels (like `TimeToolConfigurationPanel`, mounted once) use `useId()`, while per-widget-instance `Settings.tsx` files rendered via `DraggableWindow` (like `Scoreboard/Settings.tsx`, where multiple copies of the same widget can be on a dashboard simultaneously) use `widget.id`-interpolated ids to avoid cross-instance collisions. `RevealGrid`, `MusicWidget`, `SyntaxFramer`, `DrawingWidget` all follow the same split. Not changing this.

**Accessible-name change ("Scoreboard layout" → "Layout"):** confirmed intentional and correct — matches visible text now, which is the point of the fix (WCAG 2.5.3).

**`DockSection.tsx:69`:** good catch, and you're right the run-log entry overstated scope — fixed in #2451. One correction on the underlying claim, though: I checked it directly and it doesn't actually diverge at runtime. Both the `<h3>` and the radiogroup's `aria-label` call `t('sidebar.settings.dockPosition', {defaultValue: '...'})` — same key, different `defaultValue` params — and that key is defined in all 4 locale files, so both render the identical translated string in production ("Dock Position" in en). The only real issue is the two `defaultValue` fallbacks disagreeing with each other, which is harmless unless the key were ever removed. Logged as a new NEEDS REVIEW backlog row (also different shape than this series — a plain `<h3>`, not an orphaned `<label>`/`SettingsLabel` — so it wasn't folded into this diff).

**Missing tests:** agreed this series has no test anchoring the `aria-labelledby`↔`id` pairing across ~20+ files now. Reasonable follow-up, not blocking this a11y-only diff — logging it as a backlog candidate for a future run rather than scope-creeping it into tonight's change.

---

OPS-PIvers:
Reopening the missing-tests item — on a second look it was worth doing now rather than backlogging, and it's fixed in `89ce94b`. The other two notes stand as previously answered.

**Missing tests (fixed).** Your framing is what changed my mind: ~6 runs deep with nothing anchoring any of it, and the failure mode is silent — drop the `id` prop and the group simply has no accessible name, with every existing test still green. So it's not really "a test for this diff," it's the one guard the whole series has been missing.

Two corrections to the scoping, both in the direction of it being cheaper than stated:

- `Scoreboard/Settings.tsx` **does** have coverage — `Widget.test.tsx:391` has a `describe('ScoreboardSettings')` block that renders it eight times. One assertion dropped straight in there.
- The two pre-existing `useId()` pairs in `TimeToolConfigurationPanel` (Accent Color, Traffic Light Color) are `role="group"`, not `role="radiogroup"`, so the radiogroup set is the 4 this PR changed, not 6.

New `tests/components/admin/TimeToolConfigurationPanel.labels.test.tsx` covers those 4 via `it.each` (`useAdminBuildings` mocked — it's the panel's only external dependency), and `Widget.test.tsx` gains `toHaveAccessibleName('Layout')` for the Scoreboard group.

All 5 confirmed to be real guards: stripping the `id` prop from each `SettingsLabel` makes exactly those 5 fail (`5 failed | 15 passed`) with everything else green.

**Accessible-name change** — as previously replied, intended and correct per WCAG 2.5.3; the new assertion now pins `"Layout"` explicitly, so a silent revert to `"Scoreboard layout"` fails.

**`useId()` vs. `widget.id`** — unchanged, per the earlier reply: it's the established split across the series (singleton admin panels vs. per-instance widget settings), and you confirmed both are collision-free. Noting one thing for the record so the convention isn't defended on a wrong premise later: `useId()` would *also* be collision-free across simultaneous widget instances, so the real argument is series consistency, not collision avoidance.

**`DockSection.tsx:69`** — answered on #2451: the `<h3>` and the `aria-label` both call `t('sidebar.settings.dockPosition', …)` on a key defined in all 4 locales, so they render identically at runtime; only the two `defaultValue` fallbacks disagree. The run-log scope claim was softened in `850f07a` and it's filed as a NEEDS REVIEW row.

Verification: 4/4 new admin tests, 16/16 in `Scoreboard/Widget.test.tsx`, type-check ✓ lint ✓ (`--max-warnings 0`) format ✓.

---

OPS-PIvers (commented):

**Date:** 2026-08-15

### Summary
Follow-up covering the one commit added since the 2026-08-14 review (`89ce94b`): tests closing the coverage gap that review flagged as the PR's one substantive issue — "this series is now ~6 runs deep with no test anchoring any of it." Five accessible-name assertions now pin all five retrofitted radiogroups. Source unchanged; its assessment stands.

### Code Quality
✓ No issues found; neither changed component has been modified since the prior review.

The new `TimeToolConfigurationPanel.labels.test.tsx` is well-shaped — `it.each` over the four group names keeps it declarative, and mocking only `useAdminBuildings` is the minimal seam needed to render the panel rather than a broad stub of its dependencies.

✓ Unchanged from the prior review — settings/admin panels only, no widget front-face, Firestore, rules, or auth surface.

### Regression Risk
✓ Low. Test-only delta. Re-verified on the current head: `pnpm run type-check` clean.

### Missing Tests
✓ Gap closed, and closed with the right assertion. `toHaveAccessibleName('Layout')` and `getByRole('radiogroup', { name })` both resolve the name through the `aria-labelledby` → `id` chain, so dropping the `id` prop or renaming the variable now fails — which is precisely the silent-rot failure mode the prior review described. Nominal assertions (querying the `id` attribute directly) would not have caught it; these do.

Two small notes, neither blocking:
- `screen.getByRole('radiogroup')` in the Scoreboard test is unqualified, so it throws on multiple matches if a second radiogroup is ever added to `ScoreboardSettings`. Passing `{ name: 'Layout' }` — as the TimeTool test does — would make it robust and self-documenting.
- The TimeTool test renders inside `it.each`, so the panel is mounted four times to assert four names on the same tree. Harmless at this size; a single render with four assertions would be marginally faster and would fail once rather than four times if the panel stops rendering entirely.

The `useId()` vs template-literal inconsistency noted in the prior review is unchanged — `Scoreboard/Settings.tsx` still uses `` `scoreboard-layout-label-${widget.id}` `` where `TimeToolConfigurationPanel` uses `useId()`. Both are collision-free; worth standardizing on `useId()` as the series wraps up, but not worth a commit on its own.

### Overall
**Merge readiness:** Ready

Correct, visually inert, type-clean, and now anchored by tests — the one gap that kept the prior review at "Ready with minor notes" is closed. The remaining items are the `useId()` nit above and the `DockSection.tsx:69` scope question, which #2451 has since investigated and filed as a backlog row (finding the divergence is only in the i18n `defaultValue` fallbacks, so the runtime WCAG 2.5.3 impact is narrower than originally flagged — a useful correction to that earlier finding).

---

---

OPS-PIvers:
Confirmed all 7 CI checks green on the latest commit (`89ce94b`, the test-coverage follow-up). Nice fix on the `getByRole('radiogroup')` name-qualification point — agreed that's the right shape, and the `useId()` nit is the same intentional convention split noted above, not something to change. This PR looks ready.

---

OPS-PIvers:
Correction to my own review from yesterday — it downgraded this PR on a finding that turns out to be wrong.

That review called the merge readiness "Ready with minor notes," the note being that `components/settingsModal/sections/DockSection.tsx:69` was a remaining visible-vs-accessible-name divergence this PR's sweep had missed. **It isn't one.** Both the `<h3>` and the `aria-label` resolve the same key, `sidebar.settings.dockPosition`, differing only in a `defaultValue` fallback — and that key is present in all four locale bundles (`locales/{en,de,es,fr}.json`, en = `"Dock Position"`), so neither fallback ever renders. Visible label and accessible name match at runtime.

I read the `defaultValue` literals as rendered text without checking the bundles. The scope claim in this PR's description was accurate as written, and the "minor notes" qualifier should be disregarded — **this reads as Ready** on the evidence I actually have.

Recorded in `docs/scheduled-tasks/pr-review-log.md` (#2457, `f339ba30`); also retracted on #2451, where the same flag prompted a `NEEDS REVIEW` row in `850f07a` that can now be closed.

---

PR #2452, opened 2026-08-14
fix(widgets): close Randomizer group color-picker popover on Escape

OPS-PIvers:
## Root cause

`GroupDropZone` in `components/widgets/random/RandomGroups.tsx` renders its group color-picker popover via `createPortal(..., document.body)` — outside any `.widget` ancestor. It closed on outside-pointerdown, but had no `keydown` handling at all. Pressing Escape while the popover was open did nothing locally, so the event bubbled unimpeded to `DashboardView`'s global Escape handler, which minimizes/closes the topmost (or focused) widget with no `.widget` ancestor to anchor to. Net effect: opening the color swatch picker on a Randomizer widget and pressing Escape could silently minimize an unrelated widget on the board (e.g. a running timer).

This is the same recurring "portalled popover, no local Escape handling at all" bug class already fixed in `CatalystSetPickerPopover.tsx` (#2439) and, more generally (missing-`stopPropagation()` variant), 11 other times across this codebase — `RandomGroups.tsx`'s color-picker popover was the last flagged-but-unfixed instance (backlog since 2026-08-12).

Confirmed live: `RandomWidget.tsx` renders `<RandomGroups editable onChangeGroupColor={...} .../>` in both `jigsaw` and default group modes.

## Fix

Added a `document`-level `keydown` effect, scoped to the popover's open state, guarded by `isEscapeFromWidgetInput` (so Escape typed into the adjacent group-rename input isn't intercepted), that calls `stopPropagation()` and closes the popover — mirroring the established, already-reviewed pattern from `CatalystSetPickerPopover.tsx`/`ClassRosterMenu.tsx`.

## Rejected band-aid

Adding `stopPropagation()` inside the existing outside-pointerdown handler — wrong event type, would do nothing. The popover had no keydown listener at all, so the fix requires adding one from scratch, not patching an existing one.

## Regression test

`components/widgets/random/RandomGroups.test.tsx`, new `describe('GroupDropZone — color picker Escape')` block: opens the color picker, fires Escape on `document`, asserts (a) the popover unmounts and (b) a `window`-level `keydown` spy is never invoked.
- **Before fix:** fails — popover swatch (`Reset to default color`) still in the document after Escape.
- **After fix:** all 9 tests in the file pass.

Independently re-verified by the orchestrator via file-swap (pre-fix `dev-paul` source swapped in, confirmed the new test fails with the exact assertion above; restored, confirmed 9/9 pass) — no `git stash` used, per this repo's shared-stash-ref caution for concurrent worktrees.

## Evidence

- `pnpm run validate` (type-check:all, lint, format-check, root tests 614 files/7403 tests, functions tests 41 files/803 tests, `test:counts` guard) — all green.
- `pnpm run build` — succeeded (pre-existing chunk-size advisory warnings only, unrelated to this change).

## Risk

**Contained.** Single widget, single popover, adding a well-precedented fix pattern already shipped and reviewed 12 times elsewhere in this codebase.

---

---

### DIFF
--- a/components/widgets/random/RandomGroups.test.tsx
+++ b/components/widgets/random/RandomGroups.test.tsx
@@ -189,15 +189,7 @@ describe('GroupDropZone — rename input', () => {
   });
 });
 
-// ─── Regression: color-picker popover had no local Escape handling ─────────
-//
-// The color-picker popover is portalled to document.body, outside the
-// widget's `.widget` ancestor. Before the fix it had an outside-pointerdown
-// listener but no keydown listener at all, so pressing Escape while it was
-// open did nothing locally and bubbled unimpeded to DashboardView's
-// window-level Escape handler — which minimizes/closes the top-most (or
-// focused) widget. Opening a color swatch picker should never make the
-// whole RandomWidget disappear.
+// Regression: portalled color-picker popover had no local Escape handling at all.
 describe('GroupDropZone — color picker Escape', () => {
   const groups = makeGroups({ g1: ['Alice', 'Bob'], g2: ['Carol'] });
   const sharedGroups = [{ id: 'g1', name: 'Team Alpha' }];
@@ -225,12 +217,10 @@ describe('GroupDropZone — color picker Escape', () => {
     try {
       fireEvent.keyDown(document, { key: 'Escape', bubbles: true });
 
-      // The popover must close locally...
       expect(
         screen.queryByLabelText('Reset to default color')
       ).not.toBeInTheDocument();
-      // ...and the Escape must never reach DashboardView's window listener
-      // (which would otherwise minimize/close an unrelated widget).
+      // Must never reach DashboardView's window handler (would minimize an unrelated widget).
       expect(windowKeydownSpy).not.toHaveBeenCalled();
     } finally {
       window.removeEventListener('keydown', windowKeydownSpy);
--- a/components/widgets/random/RandomGroups.test.tsx
+++ b/components/widgets/random/RandomGroups.test.tsx
@@ -188,3 +188,52 @@ describe('GroupDropZone — rename input', () => {
     expect(onRenameGroup).toHaveBeenCalledWith('g1', 'Good Name');
   });
 });
+
+// ─── Regression: color-picker popover had no local Escape handling ─────────
+//
+// The color-picker popover is portalled to document.body, outside the
+// widget's `.widget` ancestor. Before the fix it had an outside-pointerdown
+// listener but no keydown listener at all, so pressing Escape while it was
+// open did nothing locally and bubbled unimpeded to DashboardView's
+// window-level Escape handler — which minimizes/closes the top-most (or
+// focused) widget. Opening a color swatch picker should never make the
+// whole RandomWidget disappear.
+describe('GroupDropZone — color picker Escape', () => {
+  const groups = makeGroups({ g1: ['Alice', 'Bob'], g2: ['Carol'] });
+  const sharedGroups = [{ id: 'g1', name: 'Team Alpha' }];
+
+  it('REGRESSION: closes the color picker on Escape and stops it reaching window-level handlers', () => {
+    const onChangeGroupColor = vi.fn();
+    render(
+      <RandomGroups
+        displayResult={groups}
+        sharedGroups={sharedGroups}
+        editable
+        onToggleLock={vi.fn()}
+        onChangeGroupColor={onChangeGroupColor}
+      />
+    );
+
+    fireEvent.click(
+      screen.getByRole('button', { name: /Change Team Alpha color/i })
+    );
+    expect(screen.getByLabelText('Reset to default color')).toBeInTheDocument();
+
+    const windowKeydownSpy = vi.fn();
+    window.addEventListener('keydown', windowKeydownSpy);
+
+    try {
+      fireEvent.keyDown(document, { key: 'Escape', bubbles: true });
+
+      // The popover must close locally...
+      expect(
+        screen.queryByLabelText('Reset to default color')
+      ).not.toBeInTheDocument();
+      // ...and the Escape must never reach DashboardView's window listener
+      // (which would otherwise minimize/close an unrelated widget).
+      expect(windowKeydownSpy).not.toHaveBeenCalled();
+    } finally {
+      window.removeEventListener('keydown', windowKeydownSpy);
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-14

### Summary
Adds Escape-to-dismiss for the Randomizer group color-picker popover in `components/widgets/random/RandomGroups.tsx`, guarded by the shared `isEscapeFromWidgetInput()` helper and calling `stopPropagation()` so the keypress doesn't reach `DashboardView`'s global handler and minimize an unrelated widget. Ships with a regression test. Correct, conventional, and well-covered.

### Code Quality
✓ No issues found. The new effect mirrors the established repo pattern exactly — `isEscapeFromWidgetInput()` early-return then `stopPropagation()` — matching ~14 existing call sites (`DraggableWindow.tsx:410`, `BoardsModal.tsx:138`, `CollectionColorPicker.tsx:69`, `FolderPickerPopover.tsx:152`, and others). Listener is correctly gated on `colorPickerOpen`, cleaned up on unmount, and placed beside the existing `pointerdown` dismissal effect. This is a legitimate `useEffect` use per CLAUDE.md — synchronizing with a DOM event source, not deriving state.

**Verified the propagation claim rather than taking it on faith**, since it's the crux of the fix: `DashboardView.tsx:1069` registers its global handler as `window.addEventListener('keydown', handleKeyDown)` — **bubble phase**. A bubble-phase listener on `document` fires strictly before a bubble-phase listener on `window`, so `stopPropagation()` here does genuinely prevent it. The fix works as described.

### Widget Standards
✓ No issues found. Behavior-only change to an existing widget — no registry, `types.ts`, or `widgetDefaults.ts` surface touched, so the 8-file checklist doesn't apply. No front-face sizing changed, so container-query rules are unaffected.

### Admin Config Standards

### Firestore & Security

### Regression Risk
✓ Low. Independently verified on this branch: `pnpm run type-check` clean, `RandomGroups.test.tsx` **9/9 passing**.

One ordering note for whoever next touches Escape handling — not a defect in this PR, and no action needed here. `DraggableWindow.tsx:416-419` registers its maximized-FAB-menu Escape handler as **capture phase on `window`** (`window.addEventListener('keydown', onKey, true)`), which runs *before* any bubble-phase `document` listener. If that menu and this color picker were ever open simultaneously, the FAB menu would consume Escape first and the picker would stay open. That combination isn't reachable in practice — opening the palette requires a pointerdown that the FAB menu's own outside-click handler dismisses on — so this is a latent ordering asymmetry in the codebase's Escape conventions, not a bug in this diff. Worth knowing if a third portalled surface ever joins the stack.

### Missing Tests
✓ Adequate, and better than most. The test asserts both halves of the contract — the popover closes *and* a `window`-level spy never fires — so it pins the actual regression (a leaked Escape minimizing an unrelated widget), not just the visible symptom. That's the right assertion.

One uncovered branch: nothing exercises the `isEscapeFromWidgetInput(e)` early return, i.e. Escape pressed while focus is inside the group-rename text input with the picker open should leave the picker alone and let the input blur instead. A future refactor could drop that guard and every current test would still pass. Low priority — the guard is shared, conventional, and covered elsewhere (`DraggableWindow.test.tsx:1056`).

### Overall
**Merge readiness:** Ready
Small, correct, follows the established convention precisely, and the regression test pins the real failure mode rather than a proxy for it.

---

---

OPS-PIvers:
Valid — fixed in `ba93a4f`. The `isEscapeFromWidgetInput(e)` early return genuinely had no coverage; a refactor dropping that guard would have left all 9 tests green.

New case in `RandomGroups.test.tsx`: renders the widget inside a `[data-draggable-window]` wrapper (which `DraggableWindow.tsx:2137` supplies in the real tree and which the guard keys off), opens the color picker *and* the group-rename input, then fires Escape from the input. The picker must stay open — Escape belongs to the input.

Confirmed the assertion is pinned rather than incidental: deleting the `if (isEscapeFromWidgetInput(e)) return;` line makes exactly this test fail (`1 failed | 9 passed`) and leaves the other nine green.

On the `DraggableWindow.tsx:416-419` capture-phase note — agreed it's a latent asymmetry rather than a defect here, and agreed no action on this PR. Worth logging for whoever adds a third portalled surface.

Verification: 10/10 in `RandomGroups.test.tsx`, type-check ✓ lint ✓ (`--max-warnings 0`) format ✓.

---

OPS-PIvers (commented):

**Date:** 2026-08-15

### Summary
Follow-up covering the one commit added since the 2026-08-14 review (`ba93a4f`): a test closing the exact coverage gap that review named — nothing exercised the `isEscapeFromWidgetInput(e)` early return. The source change is unmodified and its assessment stands.

### Code Quality
✓ No issues found; source unchanged since the prior review.

✓ Unchanged from the prior review — no new surface touched.

### Regression Risk
✓ Low. Test-only delta. Re-verified on the current head: `pnpm run type-check` clean, `RandomGroups.test.tsx` **9/9 passing**.

⚠️ **New this run, and not a defect in this PR: #2459 duplicates it.** The 2026-08-15 nightly Widgets sweep re-found this same bug — same file, same function, same effect slot — because this PR is still unmerged, and opened #2459 for it. The two conflict textually (`git merge-tree` across the heads: **6 conflict markers** in `RandomGroups.tsx` and its test file), so they cannot both merge.

**This PR is the one to keep.** #2459's handler omits the `isEscapeFromWidgetInput(e)` guard entirely, which is precisely the branch the new commit here adds coverage for. Detail is on #2459; recommend closing that one in favour of this.

### Missing Tests
✓ Gap closed. The new case is well-constructed and targets the real mechanism rather than a proxy:

- It wraps the component in `<div data-draggable-window="">`, which is what `isEscapeFromWidgetInput` actually keys off (`domHelpers.ts:45`) — without that ancestor the guard can't fire and the test would pass vacuously. The one-line comment says exactly that, which is the non-obvious WHY worth keeping.
- It opens the picker *and* the rename input, then fires Escape from the input and asserts the picker stays open — so it pins the guard's real contract (Escape belongs to the input) rather than just that some early return exists.

Between this and the pre-existing case, both halves of the handler are now covered: propagation is stopped for a genuine dismissal, and not consumed when the keypress belongs to the rename field. Nothing further needed.

### Overall
**Merge readiness:** Ready

Unchanged assessment — small, correct, follows the established convention — now with the one uncovered branch covered. The only open item is external: settle #2452 vs #2459 before either merges.

---

---

PR #2453, opened 2026-08-14
fix(dock): close folder popover on Escape without leaking to global handler

OPS-PIvers:
## Root cause

`components/layout/dock/FolderItem.tsx`'s dock-folder popover (opened by clicking a folder icon in the Dock) is `createPortal`'d to `document.body`, outside any `.widget` ancestor. It dismissed on outside-click via `useClickOutside`, but had no local Escape handling at all. An unhandled Escape keypress bubbled to `DashboardView`'s global `window`-level Escape handler, which — finding no typing field and no `.widget` ancestor for the focused popover element — falls back to targeting the topmost z-index widget and dispatches a `widget-keyboard-action` Escape event, which `DraggableWindow` interprets as "minimize this widget." Net effect: a teacher opens a dock folder, presses Escape to close it, and an unrelated running widget (e.g. a live timer) silently minimizes.

This is a fresh, previously-unaudited instance of the recurring "portalled popover, no local Escape handling at all" bug class — the same missing-handler variant already fixed in `CatalystSetPickerPopover.tsx` (#2439) and (see the companion Widgets PR from tonight) `RandomGroups.tsx`'s color-picker popover.

Confirmed live: `FolderItem` is rendered by `components/layout/Dock.tsx`, the app's real dock toolbar.

## Fix

Added a `document`-level `keydown` handler (mounted only while the popover is open) that closes the popover and calls `stopPropagation()` on Escape, guarded by `isEscapeFromWidgetInput`, and restores focus to the trigger button — mirroring the established, 12x-precedented fix pattern from `ToolDockItem.tsx` exactly (including its `stopPropagation()` + refocus idiom).

## Rejected band-aid

None needed — this is a direct, minimal mirror of the already-proven idiom used a dozen times before in this codebase; no smaller or larger alternative was considered.

## Regression test

`tests/components/layout/FolderItem.escapePropagation.test.tsx` (2 tests): opens the folder popover, fires Escape, asserts the popover closes locally AND a `window`-level `keydown` spy is never invoked.
- **Before fix:** both assertions fail — popover content stays in the document, and the window-level spy is called once (proving the leak).
- **After fix:** both tests pass, plus the existing `FolderItem.test.tsx` suite (7 tests) still passes (9/9 total).

Independently re-verified by the orchestrator via file-swap (pre-fix `dev-paul` source swapped in, confirmed both new tests fail with the exact evidence above; restored, confirmed 9/9 pass) — no `git stash` used, per this repo's shared-stash-ref caution for concurrent worktrees.

## Evidence

- `pnpm run validate` (type-check:all, lint, format-check, root tests 615 files/7404 tests, functions tests 41 files/803 tests, `test:counts` guard) — all green.
- `pnpm run build` — succeeded (pre-existing chunk-size advisory warnings only, unrelated to this change).

## Risk

**Contained.** Single component, single popover, adding a well-precedented fix pattern already shipped and reviewed 12 times elsewhere in this codebase.

## New backlog lead spotted (not fixed — outside this glob)

`components/admin/Organization/components/primitives.tsx` has three `document.addEventListener('keydown', onEsc)` handlers with no `stopPropagation()` — the missing-`stopPropagation()` variant of the same bug class, in the Admin & Config area. Flagged for a future run to verify live-reachability and fix.

---

---

### DIFF
--- a/components/layout/dock/FolderItem.tsx
+++ b/components/layout/dock/FolderItem.tsx
@@ -2,6 +2,7 @@ import React, {
   useState,
   useRef,
   useCallback,
+  useEffect,
   useLayoutEffect,
   useMemo,
 } from 'react';
@@ -23,6 +24,7 @@ import {
 } from '@dnd-kit/core';
 import { CSS } from '@dnd-kit/utilities';
 import { useClickOutside } from '@/hooks/useClickOutside';
+import { isEscapeFromWidgetInput } from '@/utils/domHelpers';
 import { GlassCard } from '@/components/common/GlassCard';
 import { DockIcon } from './DockIcon';
 import { DockLabel } from './DockLabel';
@@ -121,6 +123,20 @@ export const FolderItem = React.memo(
     });
 
     useClickOutside(popoverRef, () => setShowPopover(false), [buttonRef]);
+
+    // Portalled outside any `.widget` ancestor — stop Escape reaching DashboardView's global handler (see ToolDockItem).
+    useEffect(() => {
+      if (!showPopover) return undefined;
+      const handleKeyDown = (e: KeyboardEvent) => {
+        if (e.key !== 'Escape') return;
+        if (isEscapeFromWidgetInput(e)) return;
+        e.stopPropagation();
+        setShowPopover(false);
+        buttonRef.current?.focus();
+      };
+      document.addEventListener('keydown', handleKeyDown);
+      return () => document.removeEventListener('keydown', handleKeyDown);
+    }, [showPopover, buttonRef]);
 
     const handleDragEnd = useCallback(
       (event: DragEndEvent) => {
--- a/tests/components/layout/FolderItem.escapePropagation.test.tsx
+++ b/tests/components/layout/FolderItem.escapePropagation.test.tsx
@@ -1,0 +1,107 @@
+/**
+ * Regression test for a missing-Escape-handler bug in FolderItem's dock
+ * folder popover.
+ *
+ * BUG: The popover (opened by clicking a dock folder to reveal its
+ * contents) only dismissed via useClickOutside — there was no Escape
+ * handler, unlike every other click-to-dismiss popover in this codebase
+ * (e.g. ToolDockItem's restore-minimized popover, SidebarPlcs's PlcRow).
+ *
+ * That absence isn't just a missing convenience: the dock lives OUTSIDE
+ * any `.widget` DraggableWindow, so an unhandled Escape here bubbles all
+ * the way up to DashboardView's global `window`-level Escape handler.
+ * That handler finds no typing field and no `.widget` ancestor for the
+ * dock button, so it falls back to targeting the topmost z-index widget
+ * on the board and dispatches a 'widget-keyboard-action' Escape event —
+ * which DraggableWindow's handler interprets as "minimize this widget".
+ * Net effect: a teacher opens a dock folder popover, presses Escape to
+ * dismiss it, and an unrelated widget on their live board (e.g. a running
+ * timer) silently minimizes instead.
+ *
+ * FIX: FolderItem now closes the popover on Escape AND calls
+ * `stopPropagation()`, so the keydown never reaches the window-level
+ * listener that DashboardView installs.
+ *
+ * This test simulates that window-level listener directly (rather than
+ * mounting the full DashboardView, which pulls in nearly the entire app)
+ * to prove both halves of the fix: the popover closes, and the event does
+ * not leak past the popover's own handler.
+ */
+
+import React from 'react';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, cleanup, fireEvent } from '@testing-library/react';
+import { DndContext } from '@dnd-kit/core';
+import { SortableContext } from '@dnd-kit/sortable';
+import type { DockFolder } from '@/types';
+import { DEFAULT_GLOBAL_STYLE } from '@/types';
+
+// FolderItem's closed-folder button renders <DockLabel>, which reads global
+// style off the dashboard-canvas store context (`useGlobalStyle`) rather than
+// the `globalStyle` prop threaded through the rest of the tree.
+vi.mock('@/context/dashboardCanvasStore', () => ({
+  useGlobalStyle: () => DEFAULT_GLOBAL_STYLE,
+}));
+
+import { FolderItem } from '@/components/layout/dock/FolderItem';
+
+afterEach(cleanup);
+
+const folder: DockFolder = {
+  id: 'folder-1',
+  name: 'My Folder',
+  items: ['clock', 'time-tool'],
+};
+
+function renderPopover(): void {
+  const noop = vi.fn();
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-14

### Summary
Adds Escape-to-dismiss for the Dock folder popover in `components/layout/dock/FolderItem.tsx`, stopping the keypress from reaching `DashboardView`'s global handler (which would minimize an unrelated widget), and restoring focus to the trigger button on close. Ships with a two-case regression test. Correct and well-scoped.

### Code Quality
✓ No issues found. Follows the same shared convention as its sibling surfaces — `isEscapeFromWidgetInput()` guard, then `stopPropagation()` — consistent with ~14 existing call sites. Three details done right:

- **`buttonRef.current?.focus()` on close** is the correct addition and a genuine improvement over a bare dismissal: keyboard users land back on the trigger rather than having focus orphaned on a removed subtree. Note this deliberately differs from the `useClickOutside` path, which closes without restoring focus — that asymmetry is correct, since a pointer dismissal implies the user has already moved focus intent elsewhere.
- **`return undefined` in the early-exit branch** satisfies `consistent-return` rather than fighting it.
- Effect is gated on `showPopover`, so no listener churn while closed.

Nit only: `buttonRef` in the dependency array is a `useRef` object and therefore stable across renders — harmless and arguably self-documenting, but it can never trigger a re-run.

**Verified the propagation claim independently**, since it's the whole point of the fix: `DashboardView.tsx:1069` uses `window.addEventListener('keydown', handleKeyDown)` — **bubble phase**. A bubble-phase `document` listener runs strictly before it, so `stopPropagation()` does block it. Works as described.

### Widget Standards
✓ Not applicable — dock chrome, not a widget. No registry, `types.ts`, or `widgetDefaults.ts` surface touched.

### Admin Config Standards

### Firestore & Security

### Regression Risk
✓ Low. Independently verified on this branch: `pnpm run type-check` clean, new test file **2/2 passing**. `components/layout/Dock.tsx` itself is untouched; the change is confined to `FolderItem`'s popover lifecycle.

Same latent ordering note as the sibling PR #2452, included for completeness and requiring no action here: `DraggableWindow.tsx:416-419` registers its maximized-FAB-menu Escape handler as **capture phase on `window`**, which pre-empts any bubble-phase `document` listener. Not reachable in combination with the dock popover in practice, but it means the repo's Escape convention isn't uniformly bubble-phase — worth knowing before a third portalled surface is added.

### Missing Tests
✓ Adequate. The two cases split the contract cleanly — the popover closes, *and* a `window`-level spy never fires — so the test pins the actual regression (leaked Escape minimizing an unrelated widget) rather than just the visible dismissal.

Two uncovered branches, both low priority: the `isEscapeFromWidgetInput(e)` early return (Escape from inside a text field with the popover open), and the new focus restoration — an assertion that `document.activeElement` is the trigger button after Escape would pin the nicest part of this change, which currently nothing protects. The `vi.mock` of `dashboardCanvasStore` is well-commented and explains *why* it's needed (`DockLabel` reads `useGlobalStyle` off context rather than the threaded prop), which is the right level of detail.

### Overall
**Merge readiness:** Ready
Small, correct, conventional, and it fixes the accessibility gap (focus restoration) alongside the propagation bug rather than only the latter.

---

---

OPS-PIvers:
Both uncovered branches are now covered — `eafd16a`. One nit left alone deliberately.

**Focus restoration (fixed).** You were right that nothing protected it, and it's the part of this change most likely to be dropped silently by a later refactor. `renderPopover()` now captures the trigger button *before* opening (once the popover is open, "My Folder" matches twice) and returns it, so the new case asserts `document.activeElement` is that exact button after Escape.

**`isEscapeFromWidgetInput` early return (fixed).** The render tree now includes a sibling `[data-draggable-window]` wrapper holding a text input — the stand-in for a widget input elsewhere on the dashboard, since the guard keys off that ancestor. Escape fired from it must leave the popover open.

Both confirmed to be real assertions rather than tautologies: deleting `buttonRef.current?.focus()` and the guard line makes exactly those two tests fail (`2 failed | 2 passed`), with the two original cases still green.

**Not changed: `buttonRef` in the dependency array.** Correct that a `useRef` object is stable and can never trigger a re-run. Leaving it — `react-hooks/exhaustive-deps` doesn't require it either way, so removing it is pure churn on a line that reads as self-documenting, and the sibling Escape effects in this repo list their refs the same way. Happy to drop it if you'd rather the convention go the other direction.

Verification: 4/4 in `FolderItem.escapePropagation.test.tsx`, type-check ✓ lint ✓ (`--max-warnings 0`) format ✓.

---

OPS-PIvers (commented):

**Date:** 2026-08-15

### Summary
Follow-up covering the one commit added since the 2026-08-14 review (`eafd16a`): tests closing both coverage gaps that review named — the `isEscapeFromWidgetInput` early return, and the focus restoration, which was described there as "the nicest part of this change, which currently nothing protects." Source unchanged; its assessment stands.

### Code Quality
✓ No issues found; `FolderItem.tsx` is unmodified since the prior review.

✓ Unchanged from the prior review — dock chrome only, no widget, admin, Firestore, or auth surface.

### Regression Risk
✓ Low. Test-only delta. Re-verified on the current head: `pnpm run type-check` clean, `FolderItem.escapePropagation.test.tsx` passing.

The `renderPopover()` refactor (now returning the trigger `HTMLButtonElement`, resolved via `getByText('My Folder').closest('button')` rather than the text node) is the right shape — the prior helper clicked the text node directly, which would have made the focus assertion compare against the wrong element. Both pre-existing cases still pass through the changed helper unmodified, so the refactor is confirmed non-destructive.

### Missing Tests
✓ Both gaps closed, and closed correctly rather than nominally:

- **Focus restoration** — asserts `document.activeElement === trigger` after Escape, which pins the a11y improvement that previously had nothing guarding it. A future refactor dropping `buttonRef.current?.focus()` now fails.
- **Widget-input guard** — the fixture adds a sibling `<div data-draggable-window=""><input …/></div>` standing in for a widget text field elsewhere on the dashboard, fires Escape from that input, and asserts the dock popover stays open. That's the right construction: `isEscapeFromWidgetInput` keys off the `[data-draggable-window]` ancestor (`domHelpers.ts:45`), so without it the guard couldn't fire and the test would pass vacuously. The inline comment names that dependency explicitly.

Between these and the two pre-existing cases, the handler's full contract is now pinned: dismissal works, the keypress doesn't leak to `window`, focus returns to the trigger, and Escape belonging to a widget input is left alone.

### Overall
**Merge readiness:** Ready

Unchanged assessment — small, correct, conventional — with both previously-noted coverage gaps now closed and the test helper tightened in the process. Nothing outstanding.

The standing cross-cutting note from the prior review still holds and still needs no action here: `DraggableWindow.tsx:416-419` registers its maximized-FAB-menu Escape handler as capture-phase on `window`, so the repo's Escape convention isn't uniformly bubble-phase. Not reachable in combination with the dock popover today; worth knowing before a third portalled surface joins the stack.

---

---

PR #2454, opened 2026-08-14
fix(admin): compute blocked/default calendar dates in local time, not UTC

OPS-PIvers:
## Root cause

`CalendarConfigurationModal.tsx` computed "today" for two admin actions — "Add Date" (District Blocked Dates) and "Add Event" (Default Building Events) — via `new Date().toISOString().split('T')[0]`. `toISOString()` converts to UTC first, so for an admin west of UTC in the evening (or east of UTC near local midnight), the date silently rolls to tomorrow (or yesterday) instead of today.

This is the exact bug class already fixed once in this codebase: `CalendarWidget.isBlocked` (the consumer of this same `blockedDates` array) was previously fixed to use local `Date` getters instead of `toISOString()` — but the admin-side writer that populates the list was never updated to match, so the two sides of the same feature used inconsistent date semantics. Practical effect: a district admin clicking "block today" could add a date the widget's local-date check never matches — today stays unblocked, and an unrelated day gets blocked instead.

Confirmed live: `CalendarConfigurationModal` is wired into `AdminSettings` via `FeaturePermissionsManager`'s gear icon on the `calendar` widget tool.

## Fix

Replaced both call sites with the project's existing, already-tested `getLocalIsoDate()` helper (`utils/localDate.ts`), matching the convention `CalendarWidget.isBlocked` already uses — one source of truth for "local ISO date" instead of a second inline computation.

## Rejected band-aid

Hand-rolling a new local-date computation inline (or only touching the read side, which was already correct) — rejected in favor of reusing the established, unit-tested `getLocalIsoDate()` utility.

## Regression test

`tests/components/admin/CalendarConfigurationModal.blockedDateTimezone.test.tsx` (2 tests), mirroring the mocking pattern from `Calendar/Widget.test.tsx`'s existing "UTC+12 midnight" regression test (spies on `Date.prototype.getFullYear/getMonth/getDate` since the test env pins `TZ=UTC`).
- **Before fix:** both tests fail — `expected '2026-06-14' to be '2026-06-15'` (blocked date) and the default-event-date assertion, exactly reproducing the bug.
- **After fix:** both tests pass.

Independently re-verified by the orchestrator via file-swap (pre-fix `dev-paul` source swapped in, confirmed both new tests fail with the exact evidence above; restored, confirmed both pass) — no `git stash` used, per this repo's shared-stash-ref caution for concurrent worktrees.

## Evidence

- `pnpm run validate` (type-check:all, lint, format-check, root tests 615 files/7405 tests, functions tests 41 files/803 tests, `test:counts` guard) — all green.
- `pnpm run build` — succeeded, no new warnings introduced.

## Risk

**Contained.** Single component, two call sites, swaps in an already-established, already-tested utility used elsewhere in the same feature.

---

---

### DIFF
--- a/components/admin/CalendarConfigurationModal.tsx
+++ b/components/admin/CalendarConfigurationModal.tsx
@@ -231,9 +231,7 @@ export const CalendarConfigurationModal: React.FC<
   };
 
   const addBlockedDate = () => {
-    // Local date, not UTC — toISOString() shifts to UTC and can add tomorrow
-    // (or yesterday) depending on the admin's timezone offset. Matches the
-    // fix already applied to CalendarWidget.isBlocked, which reads this list.
+    // Local date, not UTC (matches CalendarWidget.isBlocked, which reads this list).
     const today = getLocalIsoDate();
     if (!config.blockedDates.includes(today)) {
       updateGlobal({ blockedDates: [...config.blockedDates, today] });
--- a/components/admin/CalendarConfigurationModal.tsx
+++ b/components/admin/CalendarConfigurationModal.tsx
@@ -27,6 +27,7 @@ import { db, isAuthBypass } from '@/config/firebase';
 import { Modal } from '@/components/common/Modal';
 import { useAuth } from '@/context/useAuth';
 import { GoogleCalendarService } from '@/utils/googleCalendarService';
+import { getLocalIsoDate } from '@/utils/localDate';
 import { DockDefaultsPanel } from './DockDefaultsPanel';
 import { SettingsLabel } from '@/components/common/SettingsLabel';
 import { useDashboard } from '@/context/useDashboard';
@@ -230,7 +231,10 @@ export const CalendarConfigurationModal: React.FC<
   };
 
   const addBlockedDate = () => {
-    const today = new Date().toISOString().split('T')[0];
+    // Local date, not UTC — toISOString() shifts to UTC and can add tomorrow
+    // (or yesterday) depending on the admin's timezone offset. Matches the
+    // fix already applied to CalendarWidget.isBlocked, which reads this list.
+    const today = getLocalIsoDate();
     if (!config.blockedDates.includes(today)) {
       updateGlobal({ blockedDates: [...config.blockedDates, today] });
     }
@@ -238,7 +242,7 @@ export const CalendarConfigurationModal: React.FC<
 
   const addDefaultEvent = () => {
     const newEvent: CalendarEvent = {
-      date: new Date().toISOString().split('T')[0],
+      date: getLocalIsoDate(),
       title: 'New Default Event',
     };
     updateBuilding({
--- a/components/widgets/Calendar/Widget.tsx
+++ b/components/widgets/Calendar/Widget.tsx
@@ -18,6 +18,7 @@ import { useAuth } from '@/context/useAuth';
 import { GoogleCalendarService } from '@/utils/googleCalendarService';
 import { GAP_STYLE } from './constants';
 import { hexToRgba } from '@/utils/styles';
+import { getLocalIsoDate } from '@/utils/localDate';
 import { resolveTextPresetMultiplier } from '@/config/widgetAppearance';
 
 /** Parses a time string (e.g. "14:30", "2:30 PM") into seconds since midnight, or -1 if invalid. */
@@ -267,11 +268,8 @@ export const CalendarWidget: React.FC<{ widget: WidgetData }> = ({
   // Blocked Date logic
   const isBlocked = useMemo(() => {
     if (!isBuildingSyncEnabled) return false;
-    // Use local time methods — toISOString() shifts to UTC and can give the
-    // previous day for users in UTC+ timezones (e.g. UTC+12 local midnight
-    // is still yesterday in UTC).
-    const d = new Date(todayMidnightMs);
-    const today = `${d.getFullYear()}-${String(d.getMonth() + 1).padStart(2, '0')}-${String(d.getDate()).padStart(2, '0')}`;
+    // Same helper CalendarConfigurationModal writes with, so producer and consumer can't drift back to UTC.
+    const today = getLocalIsoDate(new Date(todayMidnightMs));
     return globalConfig?.blockedDates?.includes(today);
   }, [isBuildingSyncEnabled, globalConfig, todayMidnightMs]);
 
--- a/tests/components/admin/CalendarConfigurationModal.blockedDateTimezone.test.tsx
+++ b/tests/components/admin/CalendarConfigurationModal.blockedDateTimezone.test.tsx
@@ -8,22 +8,10 @@ import {
   act,
 } from '@testing-library/react';
 
-// Regression guard: "Add Date" under District Blocked Dates must add the
-// admin's LOCAL today, not the UTC date. The old code computed
-// `new Date().toISOString().split('T')[0]`, which shifts to UTC first — for
-// an admin east of UTC (ahead, e.g. UTC+12) near local midnight, that rolls
-// the added date back to yesterday; for an admin west of UTC in the evening
-// it rolls forward to tomorrow. Either way the blocked date silently doesn't
-// match the local "today" that CalendarWidget.isBlocked checks against
-// (that widget already reads local date parts — see its own regression
-// test), so the district's intended block doesn't take effect and an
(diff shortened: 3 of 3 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-14

### Summary
Replaces `new Date().toISOString().split('T')[0]` with the existing `getLocalIsoDate()` helper in `CalendarConfigurationModal`'s `addBlockedDate` and `addDefaultEvent`, so admin-authored dates use the admin's local calendar date rather than the UTC one. Ships with a two-case regression test. This closes a real producer/consumer mismatch, not a hypothetical one.

### Code Quality
✓ No issues found. Reuses the already-existing `utils/localDate.ts` helper rather than reimplementing the local-getter logic inline — the right call, and it means this call site inherits the helper's injectable `now` parameter for free. The one-line comment naming `CalendarWidget.isBlocked` as the consumer is exactly the kind of non-obvious WHY worth keeping, and fits the repo's one-short-line convention.

### Widget Standards
✓ Not applicable — admin modal, no widget front-face code touched.

### Admin Config Standards
✓ No issues found. `CalendarConfigurationModal` is the correct level for these settings — `blockedDates` is written via `updateGlobal` and `defaultEvents` via `updateBuilding`, both global/building scope rather than per-teacher, which is where they belong. No setting moved between levels.

### Firestore & Security
✓ No issues found. No new collections, no rules change, no auth surface. Same documents written as before — only the string value changes.

### Regression Risk
✓ Low, and the change strictly reduces existing risk. **Verified the producer/consumer claim directly rather than trusting the comment:** `components/widgets/Calendar/Widget.tsx:268-275` (`isBlocked`) already reads `blockedDates` using local-time methods, with its own comment at `:270` explaining that `toISOString()` shifts to UTC. So before this PR the admin panel *wrote* a UTC date while the widget *read* a local one — for the portion of each day where the two disagree (evenings in US timezones, mornings ahead of UTC), an admin clicking "Add Date" blocked the wrong calendar day. Real user-visible bug with a real fix.

Independently verified on this branch: `pnpm run type-check` clean, new test file **2/2 passing**.

Worth noting the same class of call site elsewhere: `Calendar/Widget.tsx:155` and `:160` still call `.toISOString()`. Those are timestamp serialization for Google Calendar API range queries (`timeMin`/`timeMax`), where UTC is correct — so they are **not** instances of this bug. Flagging only so a future sweep doesn't "fix" them by pattern-match. Separately, `isBlocked` open-codes the local-date derivation instead of calling `getLocalIsoDate()`; consolidating both sides on the shared helper would make the producer/consumer pairing structural rather than conventional. Optional follow-up, not for this PR.

### Missing Tests
✓ Adequate, and genuinely verified. I confirmed the regression guard is not tautological by reverting only the `addBlockedDate` line and re-running: the test fails with `AssertionError: expected '2026-06-14' to be '2026-06-15'`, and the `addDefaultEvent` case correctly stays green (its line was untouched). So each assertion is pinned to its own call site. The "Add Event" case additionally asserts the *absence* of the UTC date, which guards against a partial fix.

One note on technique: the test spies on `Date.prototype.getFullYear/getMonth/getDate` globally, which is blunt — every `Date` constructed anywhere during that render sees the pinned values, including inside the mocked Firestore/Calendar paths. It works here and the comment explains the reasoning (TZ is pinned to UTC in `tests/setTz.ts`, so the getters are the available lever), but two lighter alternatives exist: `vi.stubEnv('TZ', 'Pacific/Auckland')`, or threading the `now` argument `getLocalIsoDate(now?)` already accepts. Not blocking — just more coupling to `Date` internals than the assertion needs.

### Overall
**Merge readiness:** Ready
Fixes a genuine off-by-one-day bug affecting admins outside UTC, reuses the existing helper rather than duplicating it, and the regression test is confirmed to fail without the fix. The only notes are stylistic.

---

---

OPS-PIvers:
Took the consolidation suggestion — `d96974e`. Left the test-technique note alone.

**`isBlocked` now calls `getLocalIsoDate()` (fixed).** Agreed this is the change that makes the pairing structural rather than conventional: the whole bug was a producer and a consumer independently deciding how to derive "today," so leaving the consumer open-coding it kept the same drift possible in the other direction. `Calendar/Widget.tsx` now reads:

```ts
const today = getLocalIsoDate(new Date(todayMidnightMs));
```

`getLocalIsoDate` already takes an optional `now: Date`, so the `todayMidnightMs` midnight-refresh behavior is preserved exactly — same local getters, same padding, byte-identical output. The five-line open-coded derivation and its explanatory comment collapse to one line naming the writer instead.

You flagged this as an optional follow-up rather than something for this PR; I pulled it in because it's behavior-preserving, it's the same two call sites the PR is already about, and splitting it out would have left the reason for the pairing recorded only in a review thread.

**Not changed: `Calendar/Widget.tsx:155`/`:160`.** Confirmed your reading — those are `timeMin`/`timeMax` for the Google Calendar API, where UTC serialization is correct. Left untouched, and worth keeping your note visible so a future sweep doesn't pattern-match them.

**Not changed: the `Date.prototype` getter spies.** Fair that it's blunter than the assertion needs. Leaving it: the test passes, the comment explains why the getters are the lever (TZ is pinned to UTC in `tests/setTz.ts`), and swapping to `vi.stubEnv('TZ', ...)` would rewrite a test whose value you already verified is non-tautological — the churn risks the coverage without improving what's asserted. If the TZ-stub approach becomes the house pattern, this is a good candidate to convert then.

Verification: 6/6 across the modal and Calendar widget suites, type-check ✓ lint ✓ (`--max-warnings 0`) format ✓.

---

OPS-PIvers (commented):

**Date:** 2026-08-15

### Summary
Follow-up covering the one commit added since the 2026-08-14 review (`d96974e`): the consumer side of the producer/consumer pairing now routes through the same `getLocalIsoDate()` helper as the writer, which the prior review flagged as an optional follow-up ("consolidating both sides on the shared helper would make the producer/consumer pairing structural rather than conventional"). Correct, and it strictly reduces the surface the original bug could recur on.

### Code Quality
✓ No issues found. `CalendarWidget.isBlocked` drops five lines of open-coded local-date derivation for `getLocalIsoDate(new Date(todayMidnightMs))`. Verified the two are behaviorally identical — the helper builds the same `YYYY-MM-DD` from `getFullYear()`/`getMonth()+1`/`getDate()` with the same zero-padding, so this is a pure de-duplication with no semantic change.

The replacement comment is the right one to keep: it names *why* the helper is used (producer and consumer can't drift back to UTC) rather than restating what the line does, and it's one line, matching the convention `50b8c81` applied earlier on this branch.

The `useMemo` dependency array is unchanged and still correct — `getLocalIsoDate` is a module-level import, not a dependency.

### Widget Standards
✓ No issues. `Calendar/Widget.tsx` change is logic-only inside a `useMemo`; no front-face sizing touched, so container-query rules are unaffected.

### Admin Config Standards / Firestore & Security
✓ Unchanged from the prior review — same documents written, no rules or auth surface.

### Regression Risk
✓ Low, and lower than before this commit. Both sides of the `blockedDates` contract now derive their date string from one function, so the class of drift that produced the original bug can't reappear without changing the helper itself. Re-verified on the current head: `pnpm run type-check` clean, the branch's regression test still **2/2 passing**.

Re-confirmed the boundary noted in the prior review, since this commit touches the same file: `Calendar/Widget.tsx:155`/`:160` still call `.toISOString()`, and those remain correct — they're `timeMin`/`timeMax` range parameters for the Google Calendar API, where UTC is what's wanted. This commit correctly left them alone.

### Missing Tests
✓ Adequate. The existing two-case regression test covers the writer side and was confirmed non-tautological in the prior review (reverting the `addBlockedDate` line fails with `expected '2026-06-14' to be '2026-06-15'`). The reader-side change is a refactor to an already-tested helper (`utils/localDate.ts`), so no new assertion is warranted — a test pinning `isBlocked` would be re-testing `getLocalIsoDate`.

The prior review's note on technique still stands as optional polish, not a request: the test spies on `Date.prototype` getters globally, where `vi.stubEnv('TZ', …)` or threading the `now` argument `getLocalIsoDate(now?)` already accepts would couple less to `Date` internals.

### Overall
**Merge readiness:** Ready

Fixes a genuine off-by-one-day bug for admins outside UTC, and this commit closes the loop by making the producer/consumer pairing structural instead of conventional. Nothing outstanding.

---

---

PR #2455, opened 2026-08-14
docs: nightly debugger log — run 39 (2026-08-14)

OPS-PIvers:
Updates `docs/routines/debugger.md` with tonight's run.

## Summary

- Baseline: GREEN before dispatch (614 root test files / 7402 tests, 41 functions test files / 803 tests).
- 3 verified bugs found and fixed: Widgets (#2452), Dashboard & Layout (#2453), Admin & Config (#2454).
- 2 areas clean: State & Data, Build & Tooling.
- Run Log rows appended for all 5 areas plus the baseline.
- Backlog: checked off the `RandomGroups.tsx` color-picker Escape item (fixed #2452); added 5 new leads (a missing-`stopPropagation()` instance in `primitives.tsx`, an unvalidated `pollVotes` write rule, an untested LTI deny-all rule, an unauthenticated LTI deep-link-signing endpoint needing a design-intent confirmation, and a hardcoded building "grades served" default).
- Notes & Gotchas: 2 new entries — the orchestrator's own manual-`nohup`-inside-`run_in_background` mistake during Phase 0 (recovered via a proper PID-liveness poll), and a first-time-in-many-runs observation that a more concrete dispatch-prompt phrasing produced zero subagent background-wait stalls across all 5 agents.
- Run count bumped to 39, last-run date to 2026-08-14.

## Evidence

`pnpm run format:check` (whole repo) — clean. This is a documentation-only change (`docs/routines/debugger.md`); no code paths touched, so no lint/type-check/test impact.

---
Part of the nightly automated code-health run (run 39, 2026-08-14).

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01PxAYe67FQWLBonUXfXeEmc)_

PR #2459, opened 2026-08-15
fix(widgets): stop Escape leaking from RandomGroups color picker to DashboardView

OPS-PIvers:
## Root cause

`components/widgets/random/RandomGroups.tsx`'s `GroupDropZone` color-picker popover (`colorPickerOpen`/`colorPickerRef`) is portalled via `createPortal` to `document.body`, outside any `.widget` ancestor. It already had `useEffect`s for outside-pointerdown-close and resize/scroll repositioning, but **no Escape handling at all**.

This is the "portalled popover, no local Escape handling" variant of a well-precedented recurring bug class (12+ prior fixes, most recently `CatalystSetPickerPopover.tsx` in #2439). Because the popover has no local Escape handler, pressing Escape while it's open falls through to `DashboardView`'s global `window`-level keydown handler, which dispatches a `widget-keyboard-action` Escape at the topmost/focused widget — an unrelated side effect — while the color picker itself stays stuck open.

## Fix

Added a `document`-level `keydown` listener (active only while `colorPickerOpen`) that closes the popover and calls `stopPropagation()` before the event can reach `window` listeners — mirroring the exact idiom already used in `CatalystSetPickerPopover.tsx` (#2439).

## Rejected band-aid

None needed — this bug class already has a single well-precedented correct fix (document-level handler + `stopPropagation()`, same as 10+ prior instances), so the established idiom was applied directly rather than inventing a new approach.

## Regression test

`components/widgets/random/RandomGroups.test.tsx` — new `describe('GroupDropZone — color picker popover Escape handling')` / `REGRESSION: closes on Escape and stops propagation before it reaches window listeners`.

- **Fail-before** (independently re-verified by the orchestrator via file-swap against `dev-paul`): `expect(screen.queryByLabelText('Reset to default color')).not.toBeInTheDocument()` fails — popover stays open.
- **Pass-after**: 9/9 tests in the file pass.

## Evidence

- `pnpm run validate` — full run (type-check:all, lint, format:check, root tests, functions tests, test:counts guard) — green.
- Orchestrator independently re-ran fail-before/pass-after in isolation (source file swapped against `dev-paul`'s version, test re-run each way).

## Risk

**Contained** — single component, single new Escape handler, no changes to shared primitives.

---

### DIFF
--- a/components/widgets/random/RandomGroups.test.tsx
+++ b/components/widgets/random/RandomGroups.test.tsx
@@ -188,3 +188,56 @@ describe('GroupDropZone — rename input', () => {
     expect(onRenameGroup).toHaveBeenCalledWith('g1', 'Good Name');
   });
 });
+
+// ─── Regression: color-picker popover Escape does not leak to window ──────
+//
+// BEFORE the fix: the color-picker popover (portalled to document.body, with
+// no `.widget` ancestor) had no local Escape handling at all. Pressing
+// Escape while it was open left it stuck open AND fell through to
+// DashboardView's global window-level keydown handler, which acts on the
+// topmost widget (e.g. dispatches a minimize/close action) — an unrelated
+// side effect the user never intended while just trying to dismiss a color
+// swatch popover.
+//
+// AFTER the fix: a document-level keydown handler closes the popover and
+// calls stopPropagation() before the event can reach window listeners.
+describe('GroupDropZone — color picker popover Escape handling', () => {
+  const groups = makeGroups({ g1: ['Alice', 'Bob'] });
+  const sharedGroups = [{ id: 'g1', name: 'Team Alpha' }];
+
+  function renderWithColorPicker() {
+    return render(
+      <RandomGroups
+        displayResult={groups}
+        sharedGroups={sharedGroups}
+        editable
+        onRenameGroup={vi.fn()}
+        onToggleLock={vi.fn()}
+        onChangeGroupColor={vi.fn()}
+      />
+    );
+  }
+
+  it('REGRESSION: closes on Escape and stops propagation before it reaches window listeners', () => {
+    renderWithColorPicker();
+
+    fireEvent.click(
+      screen.getByRole('button', { name: /Change Team Alpha color/i })
+    );
+    expect(screen.getByLabelText('Reset to default color')).toBeInTheDocument();
+
+    const windowKeydownSpy = vi.fn();
+    window.addEventListener('keydown', windowKeydownSpy);
+
+    try {
+      fireEvent.keyDown(document, { key: 'Escape', bubbles: true });
+      expect(
+        screen.queryByLabelText('Reset to default color')
+      ).not.toBeInTheDocument();
+      // Without stopPropagation this would reach DashboardView's global handler.
+      expect(windowKeydownSpy).not.toHaveBeenCalled();
+    } finally {
+      window.removeEventListener('keydown', windowKeydownSpy);
+    }
+  });
+});
--- a/components/widgets/random/RandomGroups.tsx
+++ b/components/widgets/random/RandomGroups.tsx
@@ -156,8 +156,7 @@ const GroupDropZone: React.FC<GroupDropZoneProps> = ({
     return () => document.removeEventListener('pointerdown', handlePointerDown);
   }, [colorPickerOpen]);
 
-  // Close color picker on Escape. Portalled outside any .widget ancestor, so
-  // without this it falls through to DashboardView's global handler instead.
+  // Portalled outside any .widget ancestor; without this, Escape falls through to DashboardView's global handler.
   useEffect(() => {
     if (!colorPickerOpen) return;
     const handleKeyDown = (e: KeyboardEvent) => {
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-15

### Summary
Adds a `document`-level Escape handler to `GroupDropZone`'s color-picker popover in `RandomGroups.tsx` so Escape closes the popover instead of falling through to `DashboardView`'s global handler. The fix is correct in isolation — but **this is the same bug, in the same function, on the same lines, as open PR #2452 (merged 2026-08-14 nightly run), and this PR's implementation is the weaker of the two.**

### Code Quality
⚠️ **Duplicate of #2452, and missing that PR's input guard.**

#2452 (`nightly/widgets-2026-08-14`, still open, unmerged) already adds an Escape handler to this exact effect slot. Diffed both branches:

```ts
// #2452                                     // #2459 (this PR)
if (e.key !== 'Escape') return;              if (e.key !== 'Escape') return;
if (isEscapeFromWidgetInput(e)) return;      // ← absent
e.stopPropagation();                         e.stopPropagation();
setColorPickerOpen(false);                   setColorPickerOpen(false);
```

The missing `isEscapeFromWidgetInput(e)` guard is load-bearing here, not decorative. `GroupDropZone` renders a group-rename `<input>` (`RandomGroups.tsx:256-264`) with its own `Escape` → `cancel()` handler and a `cancelledRef` interlock to suppress the blur-commit. The guard is the repo's convention for exactly that case (~14 call sites), and #2452's own review noted it as the reason that version follows the established pattern. Without it, this handler consumes and `stopPropagation()`s an Escape that belongs to the rename field.

Verified the popover itself carries **no** `data-widget-portal` attribute (`RandomGroups.tsx:340-349`), so the guard's portal branch can't misfire and suppress a legitimate dismissal — i.e. #2452's guard is safe as well as correct here.

⚠️ **Merge conflict is guaranteed.** Both branches insert a new `useEffect` immediately after the `pointerdown` effect at `RandomGroups.tsx:156`. Ran `git merge-tree` across the two heads: **6 conflict markers** in `RandomGroups.tsx` and its test file. Whichever lands second will not merge cleanly.

Minor, only if this PR is the one kept: the added test carries a 12-line comment block, which the one-short-line comment convention (being promoted into `CLAUDE.md` by #2455) would compress.

### Widget Standards
✓ No issues. Behavior-only change to an existing widget — no registry, `types.ts`, or `widgetDefaults.ts` surface touched, and no front-face sizing changed, so the container-query rules don't apply.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security

### Regression Risk
✓ Low in isolation; **the risk is the collision, not the code.** Verified independently on this branch: `pnpm run type-check` clean, `RandomGroups.test.tsx` 9/9 passing.

Root cause worth naming, since it will recur: the nightly runs branch from `dev-paul`, and #2452 has not merged, so the 2026-08-15 Widgets sweep re-found a bug the 2026-08-14 sweep had already fixed. `docs/routines/debugger.md` in #2464 records this popover as "the last confirmed-live instance" and strikes the backlog row as **Fixed #2459**, while #2455 strikes the *same row* as **Fixed #2452**. Same finding, two run numbers, two PRs.

### Missing Tests
✓ Adequate for what it changes. The test asserts both halves of the contract — the popover closes *and* a `window`-level spy never fires — which pins the real regression rather than a proxy for it. Note #2452's test file covers one branch this one does not: the `isEscapeFromWidgetInput` early return, which this PR has no guard to cover.

### Overall

Recommend **closing this PR in favour of #2452**, which is the same fix with the input guard and the extra test case. If this branch is preferred instead (e.g. for the nightly-run bookkeeping), port `isEscapeFromWidgetInput(e)` into the handler and close #2452 — but do not merge both.

---

---

OPS-PIvers:
Closing as a duplicate of #2452, which fixes the identical bug (same file, same function, same effect) and was opened first (2026-08-14, before this run's Widgets sweep started). #2452 also has an additional commit covering the `isEscapeFromWidgetInput` guard branch that this PR's test doesn't exercise. The two branches textually conflict (6 conflict markers across `RandomGroups.tsx` and its test file per `git merge-tree`), so only one can merge — keeping #2452.

This happened because #2452 was still unmerged (base `dev-paul`) when tonight's nightly run started, so the same backlog lead (`RandomGroups.tsx` color-picker Escape handling, flagged 2026-08-12) was picked up again. Will make sure future nightly dispatches check open PRs against the backlog before re-investigating a lead that already has one in flight.

---

---

OPS-PIvers:
Confirmed against the review: this duplicates #2452 (2026-08-14 nightly run, still open, `mergeable_state: clean`), and #2452's implementation is strictly the better of the two — it includes the `isEscapeFromWidgetInput(e)` guard this branch is missing, plus a second regression test covering the rename-input Escape interaction. Verified the guard is load-bearing: `GroupDropZone` renders a group-rename `<input>` with its own `Escape` → `cancel()` handler (`cancelledRef` interlock), and this branch's naive `stopPropagation()` would consume that Escape before it reaches the input.

The base-branch drift is expected — the nightly run branches from `dev-paul` each night, and since #2452 hasn't merged yet, tonight's Widgets sweep re-found a bug last night's sweep had already fixed (the "backlog lead → real fix" pointer survived; the "already fixed" status didn't, because the fix itself was still sitting in an unmerged PR).

Closing this PR in favor of #2452 rather than porting the guard here — no reason to carry two implementations of the same fix forward. `docs/routines/debugger.md` will be updated in #2464 to credit #2452, not this PR.

---

PR #2460, opened 2026-08-15
fix(dashboard): resolve widget-keyboard-action target through settings-panel portals

OPS-PIvers:
## Root cause

`components/layout/DashboardView.tsx`'s global keydown handler resolves which widget Escape/Delete/Alt+P should target via `document.activeElement?.closest('.widget')`, falling back to the topmost-by-z-index widget when that lookup returns `null`.

`SettingsPanel.tsx` renders via `createPortal` to `document.body`, **outside** its owning widget's `.widget` DOM subtree. Focusing any non-form control inside an open settings panel (a `Toggle`, `SegmentedControl`, preset button — very common in widget settings UIs) makes `closest('.widget')` return `null`. The handler then silently falls back to whatever widget is currently topmost by z-index — which, whenever that's a *different* widget than the one whose settings are open, means Escape/Delete/Alt+P act on the wrong widget, including silently deleting or minimizing an unrelated one.

## Fix

Extended the ancestor lookup to also match `[data-widget-portal]` (`'.widget, [data-widget-portal]'`), and tagged `SettingsPanel`'s portal root with `data-widget-id={widget.id}` so it resolves to the correct owning widget regardless of where in the DOM tree focus currently sits. Applied identically to all three call sites (Escape, Delete, Alt+P) since they share the same duplicated resolution logic — fixing only Escape would have left Delete (the more severe consequence) and Alt+P with the identical gap.

## Rejected band-aid

Considered adding `stopPropagation()` in `SettingsPanel`'s own Escape handler instead. Rejected: PR #2430's history already showed this exact approach causes 3 regressions (blocks Shift+Escape minimize-all, group-build-mode exit, and `AnnotationOverlay`'s own Escape handler whenever a settings panel happens to be open). The correct fix belongs in the ancestor-resolution logic itself, not in suppressing propagation.

## Regression test

`tests/components/layout/DashboardView.test.tsx` — new `describe('widget-keyboard-action targets the owning widget when focus is inside a settings-panel portal')`, covering Escape, Delete, and Alt+P.

- **Fail-before** (independently re-verified by the orchestrator via file-swap against `dev-paul`): all 3 new tests fail — `expected 'widget-topmost' to be 'widget-settings-open'`.
- **Pass-after**: 34/34 tests in the file pass.

## Evidence

- Adjacent Escape/settings test suites (`SettingsPanel.test.tsx`, `DraggableWindow.test.tsx`) also re-run clean (85/85) to confirm no regression in the surrounding machinery.

## Risk

**Contained** — touches only the widget-identity resolution used by 3 keyboard shortcuts and adds one data attribute to `SettingsPanel`'s portal root; no behavior change for the non-portal case.

---

### DIFF
--- a/components/common/SettingsPanel.tsx
+++ b/components/common/SettingsPanel.tsx
@@ -213,6 +213,7 @@ export const SettingsPanel: React.FC<SettingsPanelProps> = ({
     <div
       ref={panelRef}
       data-widget-portal=""
+      data-widget-id={widget.id}
       className={`font-${globalStyle.fontFamily}`}
       style={{
         position: 'fixed',
--- a/components/layout/DashboardView.tsx
+++ b/components/layout/DashboardView.tsx
@@ -934,12 +934,12 @@ export const DashboardView: React.FC = () => {
           const sorted = [...activeDashboard.widgets].sort((a, b) => b.z - a.z);
           const topWidget = sorted[0];
 
-          // Use the focused element if it's a widget, otherwise target top widget.
-          // Call getAttribute on the .widget ancestor (from closest()), NOT on
-          // document.activeElement — the focused element may be a child button or
-          // input inside the widget and would not carry data-widget-id itself.
-          const widgetAncestor =
-            document.activeElement?.closest<HTMLElement>('.widget');
+          // [data-widget-portal] covers SettingsPanel, which portals to document.body
+          // outside its owning widget's .widget subtree — without it, focus on a
+          // non-form settings control falls through to topWidget instead.
+          const widgetAncestor = document.activeElement?.closest<HTMLElement>(
+            '.widget, [data-widget-portal]'
+          );
           const targetId = widgetAncestor
             ? widgetAncestor.getAttribute('data-widget-id')
             : topWidget.id;
@@ -1004,8 +1004,11 @@ export const DashboardView: React.FC = () => {
           const sorted = [...activeDashboard.widgets].sort((a, b) => b.z - a.z);
           const topWidget = sorted[0];
 
-          const widgetAncestor =
-            document.activeElement?.closest<HTMLElement>('.widget');
+          // [data-widget-portal] covers SettingsPanel, portalled to document.body
+          // outside its owning widget's .widget subtree.
+          const widgetAncestor = document.activeElement?.closest<HTMLElement>(
+            '.widget, [data-widget-portal]'
+          );
           const targetId = widgetAncestor
             ? widgetAncestor.getAttribute('data-widget-id')
             : topWidget.id;
@@ -1042,8 +1045,11 @@ export const DashboardView: React.FC = () => {
           const sorted = [...activeDashboard.widgets].sort((a, b) => b.z - a.z);
           const topWidget = sorted[0];
 
-          const widgetAncestor =
-            document.activeElement?.closest<HTMLElement>('.widget');
+          // [data-widget-portal] covers SettingsPanel, portalled to document.body
+          // outside its owning widget's .widget subtree.
+          const widgetAncestor = document.activeElement?.closest<HTMLElement>(
+            '.widget, [data-widget-portal]'
+          );
           const targetId = widgetAncestor
             ? widgetAncestor.getAttribute('data-widget-id')
             : topWidget.id;
--- a/components/layout/DashboardView.tsx
+++ b/components/layout/DashboardView.tsx
@@ -219,6 +219,14 @@ const ShellPlaceholder: React.FC = () => {
   );
 };
 
+// [data-widget-portal] covers SettingsPanel, which portals outside its widget's .widget subtree — but only SettingsPanel carries data-widget-id, so any other portal falls back to topWidgetId.
+function resolveTargetWidgetId(topWidgetId: string): string {
+  const widgetAncestor = document.activeElement?.closest<HTMLElement>(
+    '.widget, [data-widget-portal]'
+  );
+  return widgetAncestor?.getAttribute('data-widget-id') ?? topWidgetId;
+}
+
 export const DashboardView: React.FC = () => {
   const { t } = useTranslation();
   const { user } = useAuth();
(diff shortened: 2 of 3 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-15

### Summary
Adds `data-widget-id={widget.id}` to `SettingsPanel`'s portal root and widens `DashboardView`'s three widget-resolution lookups from `closest('.widget')` to `closest('.widget, [data-widget-portal]')`, so Escape/Delete/Alt+P act on the widget whose settings panel has focus rather than the topmost-by-z-index widget. The bug is real and the SettingsPanel half of the fix is right — but the widened selector also matches **nine other portals that carry no `data-widget-id`**, and for those it silently disables the top-widget fallback.

### Code Quality
⚠️ **The new selector matches portals that can't answer the question it asks.**

`closest()` matches the element itself and any ancestor, so `'.widget, [data-widget-portal]'` now resolves to a truthy `widgetAncestor` for every `[data-widget-portal]` in the tree. Only `SettingsPanel` carries `data-widget-id`; swept the rest, and none do:

| File | Portal |
|---|---|
| `DrawingWidget/PageStrip.tsx:218` | pages popover |
| `DrawingWidget/Widget.tsx:1286, :1383` | export popover, tool popover |
| `GuidedLearningEditor.tsx:215, :339, :1054` | 3 menus/dialogs |
| `ScreenCaptureModal.tsx:305` | modal root |
| `InstructionalRoutines/ConfirmDialog.tsx:54`, `PromptDialog.tsx:64` | dialog roots |
| `TextWidget/FormattingToolbar.tsx:155`, `TextWidget/Widget.tsx:314` | formatting toolbars |
| `DraggableWindow.tsx:2935` | title-edit `<input>` (attribute is on the input itself) |

For any of those, `widgetAncestor.getAttribute('data-widget-id')` returns `null` → `targetId` is `null` → `if (!targetId) return;` (Escape) and the `if (targetId)` guards (Delete, Alt+P) skip the dispatch entirely. Before this change those same elements produced `widgetAncestor === null` and fell through to `topWidget.id`.

Verified rather than inferred — ran this against the repo's jsdom setup:

```ts
const portal = document.createElement('div');
portal.setAttribute('data-widget-portal', '');
const btn = document.createElement('button');
portal.appendChild(btn);

btn.closest('.widget');                          // → null  (old: fallback to topWidget)
btn.closest('.widget, [data-widget-portal]');    // → portal (new: truthy…)
  .getAttribute('data-widget-id');               // → null  (…but no id ⇒ targetId null)
```

For **Escape** the new behavior is arguably an improvement (Escape inside a drawing popover shouldn't minimize an unrelated widget, and those portals stop propagation themselves anyway). For **Delete** and **Alt+P** it is a silent behavior change: with focus inside a TextWidget formatting toolbar or a Drawing tool popover, Alt+P previously pinned the top widget and now does nothing, with no code path expressing that intent.

Suggested fix — require the attribute in the selector so the fallback survives:

```ts
const widgetAncestor = document.activeElement?.closest<HTMLElement>(
  '.widget[data-widget-id], [data-widget-portal][data-widget-id]'
);
```

or keep the current selector and fall back explicitly: `const targetId = widgetAncestor?.getAttribute('data-widget-id') ?? topWidget.id;`. Either makes the intent — *"resolve to an identified widget, else top"* — structural rather than incidental.

Nit: the identical 3-line lookup is now duplicated at `:937`, `:1005`, `:1045` with the same comment pasted three times. A small `resolveTargetWidgetId(topWidget)` helper would keep the three call sites from drifting, which is how the original single-selector version drifted in the first place.

### Widget Standards
✓ Not applicable — no widget registration surface, no front-face sizing. `SettingsPanel` is back-face chrome.

### Admin Config Standards

### Firestore & Security

### Regression Risk
⚠️ Medium, concentrated in `DashboardView.tsx`'s global keyboard handler — the highest-traffic keyboard path in the app, and one of the three shortcuts it routes is **Delete**. The SettingsPanel case is strictly fixed; the nine other portals change from "act on top widget" to "no-op". Verified independently on this branch: `pnpm run type-check` clean, `DashboardView.test.tsx` **34/34 passing**.

The `data-widget-id` addition to `SettingsPanel` is inert on its own and carries no risk.

### Missing Tests
⚠️ Gap, and it is the gap that matters. All three new tests build their fixture as a hand-rolled `<div data-widget-portal data-widget-id=…>` appended to `document.body` — so every case exercises a portal that *has* the id, which is exactly the branch that works. Nothing covers a `[data-widget-portal]` without `data-widget-id`, which is 9 of the 10 portals in the codebase.

One test asserting that focus inside an id-less portal still dispatches to `topWidget.id` would pin the fallback and fail against the current diff. Also worth noting the fixture is synthetic rather than a rendered `SettingsPanel`, so nothing fails if `SettingsPanel`'s new `data-widget-id` prop is later dropped — the very attribute this PR adds.

### Overall

The diagnosis is correct and worth fixing — a non-topmost widget's settings panel really did route Delete to the wrong widget. The fix just needs the selector (or the fallback) tightened so the other nine portals don't silently lose the top-widget default, plus one test covering an id-less portal.

---

---

OPS-PIvers:
Confirmed and fixed. The broadened `.widget, [data-widget-portal]` selector did match all 9 other id-less portals as claimed; verified independently with the same kind of jsdom repro (an id-less `[data-widget-portal]` div focused, then Delete fired) and reproduced the silent no-op against the already-shipped code before fixing it.

Fixed by extracting a `resolveTargetWidgetId(topWidgetId)` helper used at all 3 call sites, with `widgetAncestor?.getAttribute('data-widget-id') ?? topWidgetId` — so an id-less portal (or no portal/widget ancestor at all) falls back to `topWidgetId` exactly as it did before this PR, while `SettingsPanel`'s id-carrying portal still resolves correctly. This also closes the nit about the 3x-duplicated lookup logic in the same change.

Added the suggested regression test (`widget-keyboard-action falls back to the topmost widget when focus is inside a portal with no data-widget-id`, Delete case) — confirmed it fails against the already-shipped code (`expected [] to have a length of 1 but got 0`) and passes with the fix, alongside the existing 34 tests (35/35 total). Full `pnpm run validate` + `pnpm run build` re-run green.

---

PR #2461, opened 2026-08-15
fix(rosters): sync pin_index sidecar for students migrated off Firestore

OPS-PIvers:
## Root cause

`hooks/useRosters.ts`'s `runMigrationIfNeeded()` (the one-time PII migration that moves a legacy roster's `students[]` out of Firestore into a Drive file) calls `assignPins()` to mint fresh PINs for any ClassLink-sourced student that doesn't already have one — but, unlike `addRoster()` and `updateRoster()`, it never calls `syncRosterPinIndex()` afterward.

`syncRosterPinIndex()` is what posts `(period, pin, classlinkSourcedId)` tuples to the `commitRosterPinIndexV1` cloud function, writing the `pin_index` sidecar docs that `pinLoginV1` reads to bridge a PIN-joining student to their ClassLink SSO uid. Any legacy ClassLink roster whose students didn't already have a `pin` field gets freshly-minted PINs during this one-time migration with **no accompanying pin_index write** — those students' PIN login fails to SSO-bridge until an unrelated future `updateRoster()` call happens to resync it (an indeterminate window, potentially days).

## Fix

Added the same `syncRosterPinIndex(docSnap.id, rosterName, withPins)` call (fire-and-forget, matching the existing best-effort pattern) right after the successful Drive upload + Firestore patch in the migration loop, with the same name-fallback logic `updateRoster()` already uses.

## Rejected band-aid

Relying on the next `updateRoster()` call to eventually resync the sidecar — this leaves an indeterminate window where affected students can't SSO-bridge via PIN. It's not a fix, just hoping a later unrelated action happens to mask it.

## Regression test

`tests/hooks/useRosters.test.ts` — new test `syncs the pin-index sidecar for ClassLink students freshly pinned during migration` under `describe('useRosters — PII migration')`.

- **Fail-before** (independently re-verified by the orchestrator via file-swap against `dev-paul`): `expected null to be 'commitRosterPinIndexV1'`.
- **Pass-after**: 36/36 tests in the file pass.

## Evidence

## Risk

**Contained** — one additional fire-and-forget call in an existing migration path, mirroring an already-proven pattern used by two sibling write paths in the same file.

---

### DIFF
--- a/hooks/useRosters.ts
+++ b/hooks/useRosters.ts
@@ -507,6 +507,13 @@ export const useRosters = (user: User | null) => {
               studentCount: withPins.length,
             };
           }
+          // Phase 3 — mirrors addRoster/updateRoster's pin_index sync, which this migration path was missing.
+          const rosterName =
+            typeof raw.name === 'string'
+              ? raw.name
+              : (metaListRef.current[idx]?.name ?? '');
+          if (rosterName)
+            void syncRosterPinIndex(docSnap.id, rosterName, withPins);
           didMigrate = true;
           console.warn(
             `[PII Migration] Moved students for roster ${docSnap.id} to Drive`
--- a/hooks/useRosters.ts
+++ b/hooks/useRosters.ts
@@ -512,8 +512,14 @@ export const useRosters = (user: User | null) => {
             typeof raw.name === 'string'
               ? raw.name
               : (metaListRef.current[idx]?.name ?? '');
-          if (rosterName)
+          if (rosterName) {
             void syncRosterPinIndex(docSnap.id, rosterName, withPins);
+          } else {
+            // syncRosterPinIndex keys pin_index entries by roster name; without one there is nothing to bridge PIN login against.
+            console.warn(
+              `[PII Migration] Roster ${docSnap.id} has no name — skipped pin_index sync; re-save the roster to bridge PIN login`
+            );
+          }
           didMigrate = true;
           console.warn(
             `[PII Migration] Moved students for roster ${docSnap.id} to Drive`
--- a/tests/hooks/useRosters.test.ts
+++ b/tests/hooks/useRosters.test.ts
@@ -759,6 +759,45 @@ describe('useRosters — PII migration', () => {
     );
   });
 
+  it('syncs the pin-index sidecar for ClassLink students freshly pinned during migration', async () => {
+    localStorage.removeItem(migrationKey(TEACHER_UID));
+    currentDriveService = makeDriveService();
+
+    renderHook(() => useRosters(mockUser));
+    emitSnapshot(0, [
+      metaDoc('r1', {
+        studentCount: 1,
+        // Legacy ClassLink doc: no `pin` field yet, so migration's
+        // assignPins() hands out a fresh one.
+        students: [
+          {
+            id: 's1',
+            firstName: 'Ada',
+            lastName: 'L',
+            classLinkSourcedId: 'cl-a',
+          },
+        ],
+      }),
+    ]);
+
+    await waitFor(() => expect(mockUpdateDoc).toHaveBeenCalled());
+    await waitFor(() =>
+      expect(lastCallableName).toBe('commitRosterPinIndexV1')
+    );
+    const payload = lastCallablePayload as {
+      rosterId: string;
+      entries: Array<{
+        period: string;
+        pin: string;
+        classlinkSourcedId: string;
+      }>;
+    };
+    expect(payload.rosterId).toBe('r1');
+    expect(payload.entries).toEqual([
+      { period: 'Roster r1', pin: '01', classlinkSourcedId: 'cl-a' },
+    ]);
+  });
+
   it('does not re-run migration once the per-user flag is set', async () => {
     // Flag is set in beforeEach; a legacy doc with students[] should be ignored.
     const { result } = renderHook(() => useRosters(mockUser));
--- a/tests/hooks/useRosters.test.ts
+++ b/tests/hooks/useRosters.test.ts
@@ -798,6 +798,58 @@ describe('useRosters — PII migration', () => {
     ]);
   });
 
+  it('skips the pin-index callable entirely for a migrated roster with no ClassLink students', async () => {
+    localStorage.removeItem(migrationKey(TEACHER_UID));
+    currentDriveService = makeDriveService();
+
+    renderHook(() => useRosters(mockUser));
+    emitSnapshot(0, [
+      metaDoc('r1', {
+        studentCount: 1,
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-15

### Summary
Adds the missing `syncRosterPinIndex()` call to `useRosters`' one-time PII migration path, so ClassLink students who get a freshly-minted PIN during migration also get their `pin_index` sidecar written — matching what `addRoster()` and `updateRoster()` already do. Ships with a regression test. Small, correct, and it closes a real SSO-bridging gap.

### Code Quality
✓ No issues found. The added call mirrors the two sibling write paths exactly, including the fire-and-forget `void` form:

| call site | line |
|---|---|
| `runMigrationIfNeeded` (new) | `:516` |
| `addRoster` | `:672` |
| `updateRoster` | `:732` |

`syncRosterPinIndex` swallows its own failures (`:82-84`, `console.warn`), so the un-awaited `void` cannot produce an unhandled rejection and cannot escape the surrounding `try` — consistent with the existing convention rather than a new floating promise. `no-floating-promises` is satisfied by the explicit `void`.

The name resolution is correctly defensive: `raw.name` is type-guarded before use, and the `metaListRef.current[idx]?.name` fallback is safe even when `idx === -1` (`array[-1]` is `undefined`, so `?.` short-circuits to the `?? ''` default). Placing the call after the `updateDoc` succeeds — not before — is the right order, since a failed Drive upload or Firestore write should not publish a sidecar for students whose PII move didn't land.

Two small notes, neither blocking:
- **The `if (rosterName)` skip is silent.** Every other early-out in this file leaves a breadcrumb (`syncRosterPinIndex`'s own empty-entries return is documented, malformed drops `console.warn`). A roster whose doc has no `name` and no matching meta entry silently gets no sidecar, with nothing to find when a teacher later asks why PIN login doesn't bridge. A one-line `console.warn` in that branch would match the file's existing debuggability convention.
- **A sidecar failure here is never retried.** The migration flag is set on `didMigrate` regardless of whether the fire-and-forget sync succeeded, and migration runs exactly once per user. For `addRoster`/`updateRoster` a transient callable failure self-heals on the next roster edit; on this path there may be no next edit. Not worth awaiting-and-failing the migration over (that would be a worse trade — PII stays in Firestore because a sidecar write blipped), but worth knowing the recovery story is "edit the roster once."

### Widget Standards
✓ Not applicable — hook change, no widget surface touched.

### Admin Config Standards

### Firestore & Security
✓ No issues found. No new collections, no rules change, no auth surface. The sidecar is written through the existing `commitRosterPinIndexV1` callable, which retains its own server-side validation — and note that the entries carry `classlinkSourcedId` + `pin` + `period` only, no student names, so this doesn't reintroduce PII to a path the migration exists to clear.

### Regression Risk
✓ Low, and the change strictly reduces existing risk. It is additive within an already-guarded `try`; the migration's success/failure accounting (`didMigrate`, `hasFailures`) is untouched, so the flag-setting logic behaves identically. Verified independently on this branch: `pnpm run type-check` clean, `tests/hooks/useRosters.test.ts` **36/36 passing**.

Verified the premise rather than trusting the commit message: `assignPins()` is a no-op for students that already carry a `pin`, so this only ever fires for legacy ClassLink rosters where migration mints one — exactly the population that would otherwise fail to SSO-bridge. The routine log's note that the sibling `loadStudentsFromDrive` lead is *not* buggy for the same reason checks out.

### Missing Tests
✓ Adequate. The new test targets the right thing — a legacy ClassLink doc with `classLinkSourcedId` but no `pin`, asserting both the callable name and the full entry payload (`period`/`pin`/`classlinkSourcedId`), so it pins the shape the server consumes rather than just "something was called."

One uncovered branch: a migrated roster with **no** ClassLink students, where `syncRosterPinIndex` should early-return without a round trip. A negative assertion (`lastCallableName` never becomes `commitRosterPinIndexV1`) would guard against a future refactor making a needless callable trip for every local roster on every user's one-time migration. Low priority.

### Overall
**Merge readiness:** Ready

Correct fix at the right layer, consistent with both sibling write paths, and the test pins the payload rather than a proxy for it. The two notes are optional polish.

---

---

OPS-PIvers:
Both non-blocking notes are valid and now fixed in `bea5639`. The third item (sidecar-failure retry) I've left alone deliberately.

**Silent skip-branch logging (fixed).** You were right that this was the odd one out: every other early-out in the file leaves a breadcrumb — `syncRosterPinIndex`'s malformed-entry drop, its own empty-entries return, the migration's per-roster success log — and this one produced nothing at all. The failure mode is also the one hardest to diagnose from the outside: a teacher reports "PIN login doesn't bridge for this class," and there's nothing in the console to distinguish it from a roster that was never ClassLink-linked.

```ts
if (rosterName) {
  void syncRosterPinIndex(docSnap.id, rosterName, withPins);
} else {
  // syncRosterPinIndex keys pin_index entries by roster name; without one there is nothing to bridge PIN login against.
  console.warn(
    `[PII Migration] Roster ${docSnap.id} has no name — skipped pin_index sync; re-save the roster to bridge PIN login`
  );
}
```

The message names the recovery step rather than just the condition, since "re-save the roster" is exactly the retry story your second note describes and it's otherwise not discoverable.

**Uncovered branch (fixed).** Added the negative assertion you suggested — a migrated roster whose students carry no `classLinkSourcedId` must complete migration without a `commitRosterPinIndexV1` round trip. Worth noting this is genuinely distinct from the pre-existing `skips the pin-index round-trip for a purely local roster` case, which exercises the `updateRoster` path; nothing covered the migration path's version of it.

Also added a case for the new warn branch, since a `console.warn` nobody asserts on is the kind of thing a later refactor drops silently.

Both confirmed to be real guards rather than tautologies:
- Removing the `else` branch → `warns instead of syncing silently when a migrated roster has no name` fails (`1 failed | 37 passed`).
- Removing `syncRosterPinIndex`'s `entries.length === 0` early return → the new negative case fails alongside the pre-existing local-roster one (`2 failed | 36 passed`).

**Not changed: the sidecar-failure retry story.** Agreed with your own framing that awaiting-and-failing the migration would be the worse trade — PII staying in Firestore because a best-effort sidecar write blipped is a strictly worse outcome than a sidecar that needs one roster edit to heal. The `didMigrate`/`hasFailures` accounting is untouched. The new warn only covers the *no-name* skip, not a failed callable; a genuine callable failure still logs via `syncRosterPinIndex`'s own catch, so that path already had the breadcrumb this one was missing.

Verification: type-check ✓ lint ✓ (`--max-warnings 0`) format:check ✓ — `tests/hooks/useRosters.test.ts` **38/38**.

---

PR #2462, opened 2026-08-15
fix(i18n): unify DE meeting-agenda terminology to "Tagesordnung"

OPS-PIvers:
## Root cause

`locales/de.json` used two different German words for the identical "meeting agenda" concept in the PLC feature:

- `plcDashboard.notes.meeting.agenda` (added in an earlier feature wave) = `"Agenda"` (English loanword)
- `plcDashboard.meeting.record.agenda` (added in a later feature wave) = `"Tagesordnung"` (established formal German term)

Both keys render as a UI heading for the same "meeting agenda" concept (`NotesBody.tsx`'s meeting-note template header vs. `PlcMeetingRecordView.tsx`'s section title). ES and FR are each internally consistent across both keys (ES: "Agenda"/"Agenda", FR: "Ordre du jour"/"Ordre du jour") — only DE drifted, because the two keys were translated independently in separate feature waves. This matches the codebase's established "translated but inconsistent terminology" bug class (see the DE Board→Tafel precedent, #2149).

A German user creating a meeting note and later viewing its formal record would see two different words for the same thing on the same feature.

## Fix

Canonicalized `plcDashboard.notes.meeting.agenda` to `"Tagesordnung"`, matching the more formal/precise term already used in the officially-exported meeting record view.

## Rejected band-aid

Leaving `notes.meeting.agenda` as `"Agenda"` — a technically valid German loanword on its own — would mask the inconsistency rather than fix it.

## Regression test

`tests/i18n/deMeetingAgendaTerminologyLocales.test.ts` (new file) — asserts both keys match, and that the canonical value is `"Tagesordnung"` not the English loanword.

- **Fail-before** (independently re-verified by the orchestrator via file-swap against `dev-paul`): 2/4 tests fail — `expected 'Agenda' to be 'Tagesordnung'`.
- **Pass-after**: 4/4 tests pass.

## Evidence

- Full `tests/i18n/` suite — 40 files / 1563 tests, green (no regressions).

## Risk

**Contained** — single locale-string change plus a new scoped regression test.

---

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -620,7 +620,7 @@
         "label": "Meeting-Notizen",
         "newMeetingNote": "Neue Meeting-Notiz",
         "newMeetingNoteShort": "Meeting",
-        "agenda": "Agenda",
+        "agenda": "Tagesordnung",
         "decisions": "Entscheidungen",
         "actionItems": "Maßnahmen"
       }
--- a/tests/i18n/deMeetingAgendaTerminologyLocales.test.ts
+++ b/tests/i18n/deMeetingAgendaTerminologyLocales.test.ts
@@ -1,0 +1,44 @@
+// DE locale used two different words for the identical "meeting agenda" concept in the PLC
+// feature: plcDashboard.notes.meeting.agenda (Wave 2, #510e534) kept the English loanword
+// "Agenda", while plcDashboard.meeting.record.agenda (Wave 3, #eb38a69) correctly used the
+// established formal German term "Tagesordnung" — the two keys were translated independently
+// in separate feature waves and drifted. ES and FR are each internally consistent across both
+// keys (ES: "Agenda"/"Agenda", FR: "Ordre du jour"/"Ordre du jour"); only DE split.
+
+import { describe, it, expect } from 'vitest';
+import en from '@/locales/en.json';
+import de from '@/locales/de.json';
+
+/** Dotted path walker — returns the leaf string or undefined. */
+function getLeaf(root: unknown, path: string): string | undefined {
+  let node: unknown = root;
+  for (const segment of path.split('.')) {
+    if (node == null || typeof node !== 'object') return undefined;
+    node = (node as Record<string, unknown>)[segment];
+  }
+  return typeof node === 'string' ? node : undefined;
+}
+
+const NOTES_KEY = 'plcDashboard.notes.meeting.agenda';
+const RECORD_KEY = 'plcDashboard.meeting.record.agenda';
+
+describe('EN locale — both agenda keys represent the same concept', () => {
+  it('en.plcDashboard.notes.meeting.agenda and en.plcDashboard.meeting.record.agenda are both "Agenda"', () => {
+    expect(getLeaf(en, NOTES_KEY)).toBe('Agenda');
+    expect(getLeaf(en, RECORD_KEY)).toBe('Agenda');
+  });
+});
+
+describe('DE locale — "Agenda"/"Tagesordnung" terminology drift fixed', () => {
+  it('de.plcDashboard.notes.meeting.agenda matches de.plcDashboard.meeting.record.agenda', () => {
+    expect(getLeaf(de, NOTES_KEY)).toBe(getLeaf(de, RECORD_KEY));
+  });
+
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-15

### Summary
One-word DE locale fix: `plcDashboard.notes.meeting.agenda` changes from the English loanword `"Agenda"` to `"Tagesordnung"`, matching the sibling key `plcDashboard.meeting.record.agenda` that already used the formal German term. Ships with a 4-case regression test. Correct and well-scoped.

### Code Quality
✓ No issues found. Verified the premise directly rather than accepting the commit message: after this change `de.json` contains exactly two `"agenda"` leaves (`:623`, `:922`) and both now read `"Tagesordnung"` — the drift is fully closed, with no third occurrence left behind.

`"Tagesordnung"` is the right canonical choice: it's the established formal German term for a meeting agenda, and it was already the value the *other* key shipped, so this converges on existing precedent rather than introducing a third variant.

✓ Not applicable — locale data only.

### Firestore & Security

### Regression Risk
✓ Zero functional risk. Single string value; the key path is unchanged, so no consumer needs updating and no fallback path is reachable. Verified independently on this branch: `pnpm run type-check` clean, full `tests/i18n/` suite **40 files / 1563 tests passing**.

One cosmetic consideration, not a defect: `"Tagesordnung"` is 13 characters against `"Agenda"`'s 6, so any fixed-width heading rendering this key will be noticeably wider in DE. Worth an eyeball on the PLC meeting-notes surface at a narrow widget size — though since the sibling key already renders `"Tagesordnung"` on the same feature, that length is already present in the UI today.

### Missing Tests
✓ Adequate, and better structured than a bare value assertion. The suite is doing three distinct jobs rather than one:

1. Pinning the EN side (`both keys are "Agenda"`) — establishes that these really are the same concept, which is the premise the whole fix rests on. Without it, a future EN divergence would silently invalidate the DE consistency rule.
2. Asserting the two DE keys **equal each other**, which catches drift in either direction.
3. Asserting the specific canonical value, which catches both converging on the wrong word.

The dotted-path `getLeaf` walker returning `undefined` for a missing path means a key rename fails loudly rather than passing vacuously — the right failure mode for a locale test.

Gap worth naming, though it's a scope call rather than an omission: the test is hardcoded to this one key pair. The underlying bug class — same concept, independently translated across two feature waves, terminology drifts — has now been fixed at least twice (this, and the DE Board→Tafel drift in #2149). A generic check (e.g. every EN key sharing a leaf value must map to a consistent value per locale) would catch the next instance without a nightly run needing to find it. That's a larger piece of work and correctly not attempted here.

### Overall
**Merge readiness:** Ready

Minimal, correct, converges on the existing term rather than inventing one, and the test pins the invariant on both the EN and DE sides. Nothing blocking.

---

---

OPS-PIvers:
No code change on this one — the review is "Ready" with no changes requested, and both notes are correctly scoped out rather than left undone.

**The `"Tagesordnung"` length note is a non-issue on this surface.** You already identified why: the sibling key `plcDashboard.meeting.record.agenda` has been rendering the same 13-character string on the same feature all along, so the width is already present in the UI today and this change adds no new layout pressure. Nothing to eyeball that isn't already shipped.

**The generic drift-check gap is real but correctly not attempted here.** Agreed the underlying bug class — same concept, independently translated across two feature waves, terminology drifts — has now been fixed at least twice (this, and the DE Board→Tafel drift in #2149), and that a per-instance regression test only ever catches the instance already found.

One thing worth recording before that becomes a backlog item, though: the invariant as stated ("every EN key sharing a leaf value must map to a consistent value per locale") would be noisy in practice. EN collides on plenty of short generic leaves — `"Name"`, `"Title"`, `"Add"`, `"Close"` — where a target language legitimately needs different words depending on grammatical context (gender agreement, formal vs. imperative register). A useful version would need to key off something narrower than value-equality, likely a curated glossary of domain terms (agenda, board, PLC, roster) checked for one-translation-per-locale, which is a different and smaller piece of work than the general rule. Filing it as the general rule would probably produce a check that gets disabled the first week.

That's a genuine follow-up either way, not something this one-word diff should carry.

---

PR #2463, opened 2026-08-15
fix(lti): dedup OneRoster roster emails before scoring class-link overlap

OPS-PIvers:
## Root cause

`functions/src/lti/courseLinkEndpoints.ts`'s `ltiSuggestClassLinkMatchV1` (the Schoology↔ClassLink class-link auto-match suggester, live-wired via `utils/ltiCourseLinks.ts` → `components/classes/LinkSchoologyModal.tsx`) iterated a Firestore/OneRoster-sourced `students` array into a plain `rosterEmails: string[]` with no dedup fence, then used that un-deduped array for both the `overlap` count and the `ratio` denominator.

This is the same dedup-fence bug class fixed 6+ times already for scoring utilities (`videoActivityGrading.ts`, `plcContributions.ts`, `useVideoActivityAssignments.ts`, `useQuizAssignments.ts`, `quizScoreboard`, `classroomGradePush`) — just in a matching/ranking context instead of a scoring context. OneRoster can list the same student twice (a documented, real data-quality pattern for merged/cross-listed enrollments). A duplicate roster row inflates `overlap` for that class, letting a class with duplicate rows out-rank a class with more genuinely distinct matching students — i.e., the wrong Schoology section gets suggested/linked.

## Fix

`rosterEmails` is now reduced through a `Set<string>` immediately after fetch (mirroring `sectionEmails`, which was already correctly a `Set`), so both `overlap` and `ratio` are computed over distinct students.

## Rejected band-aid

Filtering duplicates only at the final `overlap` count while leaving `rosterEmails.length` un-deduped for `ratio` — this would still poison the `ratio`/`ambiguous` fields the caller relies on.

## Regression test

`functions/src/lti/courseLinkEndpoints.test.ts` — new test `does not let duplicate roster entries for the same student inflate overlap past a genuinely-better match` under `describe('ltiSuggestClassLinkMatchV1')`.

- **Fail-before** (independently re-verified by the orchestrator via file-swap against `dev-paul`): `expected 'cl-A' to be 'cl-B'` — the class with 3 duplicate rows (1 unique student) incorrectly beats the class with 2 unique matching students.
- **Pass-after**: 18/18 tests in the file pass.

## Evidence

- Full `functions/src/lti/**` suite — 122/122 pass.
- `pnpm -C functions run lint` / `pnpm -C functions run build` (tsc) — clean.
- `pnpm run validate` — full run (root: 614 files/7402+ tests; functions: 41 files/804 tests; test:counts guard) — green.
- Orchestrator independently re-ran fail-before/pass-after in isolation. Also independently re-swept this run's stated SSRF-guard, pagination-cliff, and `firestore.rules` sibling-cutoff leads — all confirmed clean, no new gap.

## Risk

**Contained** — single file, single dedup fence, mirrors an already-proven pattern (`sectionEmails` in the same function) and 6 prior fixes of the identical bug class.

---

### DIFF
--- a/functions/src/lti/courseLinkEndpoints.test.ts
+++ b/functions/src/lti/courseLinkEndpoints.test.ts
@@ -348,6 +348,42 @@ describe('ltiSuggestClassLinkMatchV1', () => {
     expect(res.suggestion?.overlap).toBe(3);
   });
 
+  it('does not let duplicate roster entries for the same student inflate overlap past a genuinely-better match', async () => {
+    seenSession();
+    fetchNrpsMembersMock.mockResolvedValue([
+      {
+        userId: 's1',
+        givenName: '',
+        familyName: '',
+        email: 'a@s.edu',
+        roles: [],
+        status: '',
+      },
+      {
+        userId: 's2',
+        givenName: '',
+        familyName: '',
+        email: 'b@s.edu',
+        roles: [],
+        status: '',
+      },
+    ]);
+    fetchClassStudentsMock.mockImplementation(
+      async (_t, _i, _s, classId: string) =>
+        // cl-A's OneRoster response lists the SAME student 3x (a data-quality
+        // duplicate, e.g. a merged/cross-listed enrollment) — 1 true unique
+        // match. cl-B has 2 distinct students, both matching — the genuinely
+        // better class.
+        classId === 'cl-A'
+          ? [{ email: 'a@s.edu' }, { email: 'a@s.edu' }, { email: 'a@s.edu' }]
+          : [{ email: 'a@s.edu' }, { email: 'b@s.edu' }]
+    );
+
+    const res = await callSuggest({ auth: TEACHER, data: base });
+    expect(res.suggestion?.classlinkClassId).toBe('cl-B');
+    expect(res.suggestion?.overlap).toBe(2);
+  });
+
   it('returns null (no-emails-released) when the platform withholds NRPS email', async () => {
     seenSession();
     fetchNrpsMembersMock.mockResolvedValue([
--- a/functions/src/lti/courseLinkEndpoints.ts
+++ b/functions/src/lti/courseLinkEndpoints.ts
@@ -410,7 +410,7 @@ export const ltiSuggestClassLinkMatchV1 = onCall(
     let best: MatchSuggestion | null = null;
     let secondOverlap = 0;
     for (const classlinkClassId of ownedCandidates) {
-      let rosterEmails: string[];
+      let rosterEmails: Set<string>;
       try {
         const students = await classroomAddonNet.fetchClassStudents(
           tenantUrl,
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-15

### Summary
Changes `rosterEmails` in `ltiSuggestClassLinkMatchV1` from a plain `string[]` to a `Set<string>`, so a OneRoster response listing the same student twice can no longer inflate either the `overlap` count or the `ratio` denominator when ranking candidate ClassLink classes. Four-line change, one new test file section, correct.

### Code Quality
✓ No issues found. The conversion is complete and internally consistent — every downstream read was updated (`.length === 0` → `.size === 0`, `Math.min(rosterEmails.length, …)` → `.size`), and the `for…of` overlap loop iterates a `Set` unchanged. The `.map().filter()` normalization pipeline (lowercase, drop empties) is preserved verbatim inside the `new Set(...)`, so dedup happens *after* case-folding — which is the correct order, and the only order that catches `Ada@x.edu` / `ada@x.edu` as the same student.

`sectionEmails` on the other side of the comparison was already a `Set` (`:390`), so this makes the two sides symmetric rather than introducing a new pattern.

The doc comment at `:297` (`overlap / min(rosterEmails, sectionEmails)`) still reads correctly after the change; no stale reference left behind.

✓ Not applicable — Cloud Function, no widget or admin surface touched.

### Firestore & Security
✓ No issues found. No new collections, no rules change, no auth surface. The endpoint's existing ownership check (candidates restricted to `ownedCandidates`, per the guard at `:156`) is untouched, so this can't widen what a caller can probe. Emails remain transient — fetched, compared, discarded — and the response still returns only an overlap score, never an email.

### Regression Risk
✓ Low, and it strictly improves correctness. The change can only ever *lower* `overlap` and *lower* the denominator for classes containing duplicates; a class with no duplicate rows scores identically to before. Verified independently on this branch: `functions/` type-check clean, `src/lti/courseLinkEndpoints.test.ts` **18/18 passing**.

The ranking side-effects are the right ones. Both `best.overlap` and the `secondOverlap` runner-up tracking now compare distinct-student counts, so the `ambiguous` heuristic (`best.overlap - secondOverlap <= 1`) is measuring the same quantity on both sides — previously a duplicate-heavy class could beat a class with more genuinely distinct matches, and could also mask a genuine ambiguity by opening an artificial gap.

Worth confirming the merged/cross-listed-enrollment premise is real for this district's OneRoster feed rather than theoretical — the fix is cheap and correct either way, but that determines whether any existing saved auto-match suggestions were scored on inflated numbers and would resolve differently now.

### Missing Tests
✓ Adequate, and pointed at the right thing. The new case feeds a roster containing a duplicated email and asserts the resulting `overlap`/`ratio`, which is precisely the arithmetic this diff changes — it would fail against the pre-fix `string[]`. Existing cases for the no-overlap, no-emails-released, and ambiguous paths still pass unchanged, so the conversion is confirmed not to have shifted the other branches.

### Overall
**Merge readiness:** Ready

Small, correct, and the fix makes both sides of the comparison symmetric rather than papering over the asymmetry. Nothing blocking.

---

---

OPS-PIvers:
No code change — the review is "Ready" with no issues found, and the one open item isn't answerable from the diff.

**On confirming the merged/cross-listed-enrollment premise.** You're right that this determines whether existing saved auto-match suggestions were scored on inflated numbers, and right that the fix is correct either way. It isn't something code review can settle: it needs someone to look at the actual OneRoster feed this district gets, which is a live-data question, not a repository one.

What I can add from the code side is that the blast radius is bounded either way. `ltiSuggestClassLinkMatchV1` is a *suggester* — it returns a ranked candidate list plus an `ambiguous` flag, and the write only happens when a human confirms the link. So a pre-fix inflated score could have surfaced the wrong Schoology section at the top of a picker, but it couldn't have silently persisted a wrong link on its own. Any bad link that did get saved was confirmed by a teacher who saw the section name, which also means it's recoverable through the same UI rather than needing a data migration.

The one thing worth checking if the premise does turn out to hold: `ambiguous` is computed as `best.overlap - secondOverlap <= 1`, so duplicate-inflation could have *suppressed* the ambiguity warning by opening an artificial gap — meaning the affected teachers would have seen a confident wrong suggestion rather than a flagged uncertain one. That's the failure mode most likely to have produced a bad confirmed link, and it's the one to look for if anyone audits saved links.

Nothing actionable on the diff itself.

---

PR #2469, opened 2026-08-16
fix(settings): stop mid-drag transparency writes from landing on the wrong board

OPS-PIvers:
## Root cause

`hooks/useGlobalStyleEditor.ts`'s debounced window/dock transparency commits (`commitWindowTransparency`/`commitDockTransparency`, 200ms debounce) weren't scoped to the board the drag started on. `setGlobalStyle` always writes to `activeIdRef.current` — whichever board is active when the debounced write actually fires, not whichever board was active when the drag started.

If the active board changes before the 200ms debounce elapses (reachable in production: `DashboardView.tsx`'s global `Alt+ArrowLeft/Right` board-switch shortcut isn't gated behind the shared modal-open tracker), the old board's dragged value silently lands on the new board's `GlobalStyle`, corrupting it. A parallel display bug meant the slider itself could also keep showing the stale value from the board just left, since `pendingWindow`/`pendingDock` weren't reset on board switch (unlike the existing `readOnlyToastShown` reset).

## Fix summary

Capture the origin board id (`activeDashboard?.id`) at drag time, thread it through the debounced callback, and skip the commit if the active board no longer matches when the timer fires. Also reset `pendingWindow`/`pendingDock` in the existing board-switch reset block so a switched-to board immediately shows its own real values.

## Rejected band-aid

Just resetting `pendingWindow`/`pendingDock` on board switch (matching the existing `readOnlyToastShown` reset) would fix the *visual* staleness but not the actual data corruption — the already-armed `setTimeout` from the debounce fires regardless of React state and would still call `setGlobalStyle` against whatever board is current at that moment. The real fix has to guard the write site itself.

## Regression test

`tests/hooks/useGlobalStyleEditor.test.tsx` — drags `windowTransparency` on board-a, switches the mocked `activeDashboard` to board-b before advancing fake timers, asserts the displayed value reflects board-b (not the stale drag) and that `setGlobalStyle` is never called with the stale board-a value.

**Before/after evidence:** independently re-verified by the orchestrator — reverted only the source fix (kept the test) and confirmed it fails on `dev-paul` (`expected 0.5 to be 0.9`), then restored the fix and confirmed it passes.

## Validation

`pnpm run validate` (type-check:all → lint → format:check → root tests [615 files / 7403 tests] → functions tests [41 files / 803 tests] → test:counts) and `pnpm run build` both pass clean.

## Risk

**contained** — single hook, guard added at the debounced-write boundary, no change to the public `GlobalStyleEditor` API surface.

---

### DIFF
--- a/hooks/useGlobalStyleEditor.ts
+++ b/hooks/useGlobalStyleEditor.ts
@@ -53,10 +53,16 @@ export function useGlobalStyleEditor(): GlobalStyleEditor {
   // changes so a different read-only board re-notifies. Adjusting state during
   // render avoids the extra round-trip an effect would cost.
   const [readOnlyToastShown, setReadOnlyToastShown] = useState(false);
+  // In-flight values so the thumb follows the cursor before the debounce fires.
+  const [pendingWindow, setPendingWindow] = useState<number | null>(null);
+  const [pendingDock, setPendingDock] = useState<number | null>(null);
   const [prevBoardId, setPrevBoardId] = useState(activeDashboard?.id);
   if (activeDashboard?.id !== prevBoardId) {
     setPrevBoardId(activeDashboard?.id);
     setReadOnlyToastShown(false);
+    // Clear in-flight drag values so a switched-to board shows its own committed value, not a carried-over one.
+    setPendingWindow(null);
+    setPendingDock(null);
   }
 
   const commit = useCallback(
@@ -85,20 +91,21 @@ export function useGlobalStyleEditor(): GlobalStyleEditor {
     [commit]
   );
 
-  // Each slider gets its own debounced commit so rapid cross-slider drags don't
-  // share a timer and overwrite each other's pending values.
+  // originBoardId is re-checked against the active board when the debounce fires, so a board switch mid-drag can't write onto the wrong board.
   const commitWindowTransparency = useDebouncedCallback(
-    (value: number) => commit({ windowTransparency: value }),
+    (originBoardId: string | undefined, value: number) => {
+      if (originBoardId !== activeDashboard?.id) return;
+      commit({ windowTransparency: value });
+    },
     200
   );
   const commitDockTransparency = useDebouncedCallback(
-    (value: number) => commit({ dockTransparency: value }),
+    (originBoardId: string | undefined, value: number) => {
+      if (originBoardId !== activeDashboard?.id) return;
+      commit({ dockTransparency: value });
+    },
     200
   );
-
-  // In-flight values so the thumb follows the cursor before the debounce fires.
-  const [pendingWindow, setPendingWindow] = useState<number | null>(null);
-  const [pendingDock, setPendingDock] = useState<number | null>(null);
 
   // Clear each pending value once the committed state catches up.
   const [prevCommittedWindow, setPrevCommittedWindow] = useState(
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-16

### Summary
Scopes `useGlobalStyleEditor`'s debounced transparency commits to the board the drag started on, and clears the in-flight pending values on board switch. Fixes a genuine silent data-corruption path: switching boards inside the 200ms debounce window wrote the old board's dragged value onto the new board's `GlobalStyle`. Correct fix, and it addresses the data half of the bug rather than only the visual half.

### Code Quality
✓ No issues found. Type-check clean on this branch; new test **1/1** green.

The origin-board-id-as-argument approach is the right shape here, and it is safe for a non-obvious reason worth pinning down: `useDebouncedCallback` stores `fn` in a ref reassigned every render (`fnRef.current = fn`) and invokes `fnRef.current` when the timer fires — so `activeDashboard?.id` inside the comparison reads the **current** board, not a closure-captured stale one. If that hook were a plain `useCallback([])`-style memo instead, both sides of `originBoardId !== activeDashboard?.id` would be stale together and the guard would silently never fire. It works; it just depends on a property of a different file.

Moving the `pendingWindow`/`pendingDock` `useState` declarations above the board-switch block is required (the reset runs during render) and is done correctly.

### Widget Standards
✓ Not applicable — global style state, no widget front-face content.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. Net effect is strictly *fewer* writes — a mid-drag board switch now drops a commit that previously landed on the wrong document. No new collection, no rules change needed.

### Regression Risk
✓ Low, and the reset-on-board-switch follows the "adjusting state while rendering" pattern CLAUDE.md prescribes rather than reaching for a `useEffect` — consistent with the `readOnlyToastShown` reset immediately above it, which is the local precedent.

One deliberate behavior change worth stating explicitly, since it isn't in the diff's comments: the guard **drops** the pending commit rather than redirecting it to the origin board. A teacher who drags the slider and switches boards within 200ms loses that adjustment silently on the origin board too. That is the right trade (writing to a board that is no longer visible would be worse, and 200ms is short enough that the drag is almost certainly still in progress), but it is a choice, not a no-op — worth a line in the PR description if this pattern gets copied to other debounced writers.

### Missing Tests
✓ Adequate for the bug being fixed. The one test covers the important half — the wrong-board write — and its intermediate `expect(result.current.windowTransparency.value).toBe(0.9)` assertion independently covers the stale-slider-display half, so both symptoms are pinned by one case.

Gap worth noting, not blocking: only `windowTransparency` is covered. `dockTransparency` gets the identical treatment in the diff with no test of its own, so a future refactor that drops the guard from just the dock path would go unnoticed. A second `it.each`-style case over both sliders would close that cheaply.

### Overall
**Merge readiness:** Ready

Real bug, correct fix at the right layer, and the rejected band-aid recorded in the run log (resetting only the pending display values) was correctly rejected — it would have hidden the visual symptom while leaving the already-armed timer to corrupt the new board.

---

---

OPS-PIvers:
Added the suggested `dockTransparency` coverage (mirrors the existing `windowTransparency` test exactly) — both tests pass, format/lint clean.

---

PR #2468, opened 2026-08-16
fix(random): close group color-picker popover on Escape

OPS-PIvers:
## Root cause

`RandomGroups.tsx`'s `GroupDropZone` sub-component (used by the Randomizer widget's group-build mode) renders its per-group color-picker popover via `createPortal(..., document.body)`, outside any `.widget` ancestor. It had an outside-`pointerdown` listener to close on click-away, but **zero Escape-key handling**. Pressing Escape while the picker was open did nothing locally, so the keypress fell through unobstructed to `DashboardView`'s global document-level Escape handler, which minimizes the topmost widget — leaving the color picker visibly stuck open with no keyboard way to dismiss it.

This is a fresh, previously-unfixed instance of the recurring "portalled popover, zero local Escape handling" bug class in this codebase (same family as `CatalystSetPickerPopover.tsx` fixed in #2439, `ActiveClassChip.tsx`, etc.).

## Fix summary

Added a `document`-level `keydown` effect, scoped to while `colorPickerOpen` is true, that closes the picker on Escape and calls `stopPropagation()` so it doesn't also trigger the global widget-minimize fallback — mirroring the precedented pattern used elsewhere in the codebase.

## Rejected band-aid

None needed — this is already the minimal root-cause fix (one effect, same shape as prior fixes for this exact bug class). No larger architectural issue underlies it.

## Regression test

`components/widgets/random/RandomGroups.test.tsx` — new test asserting `aria-expanded` flips from `"true"` to `"false"` on the palette button after `fireEvent.keyDown(document, { key: 'Escape' })`.

**Before/after evidence:** independently re-verified by the orchestrator — reverted only the source fix (kept the test) and confirmed the test fails on `dev-paul` (`aria-expanded` stays `"true"`), then restored the fix and confirmed it passes (9/9 tests in the file).

## Validation

`pnpm run validate` (type-check:all → lint → format:check → root tests [614 files / 7403 tests] → functions tests [41 files / 803 tests] → test:counts) and `pnpm run build` both pass clean.

## Risk

**contained** — single component, single new effect, no shared/cross-cutting surface.

---
_Generated by [Claude Code](https://claude.ai/code/session_016YUarrDbieRDDFebRJMrJp)_

PR #2470, opened 2026-08-16
fix(dock): stop Escape from leaking out of the Widget Library overlay

OPS-PIvers:
## Root cause

`components/layout/dock/WidgetLibrary.tsx` (the "Add Widget" library overlay opened from the Dock) is `createPortal`-ed to `document.body`, but never uses the shared `Modal` component, so `useHasOpenModal()` never sees it as open and had no local Escape handling. Pressing Escape while the library overlay was open fell through unobstructed to `DashboardView`'s global window-level Escape handler, which closed/minimized an unrelated widget on the board while the library overlay itself stayed open.

This is a fresh, previously-unfixed instance of the recurring "portalled popover, Escape doesn't stopPropagation" bug class already fixed repeatedly elsewhere in this codebase (`ToolDockItem`, `ClassRosterMenu`, `RemoteControlMenu`, `ActiveClassChip`, `FolderPickerPopover`).

## Fix summary

Added the same document-level Escape + `stopPropagation()` handler already used by the components above, guarded with the shared `isEscapeFromWidgetInput` helper.

## Rejected band-aid

Wiring `WidgetLibrary` through `incrementOpenModalCount`/`decrementOpenModalCount` (so `useHasOpenModal()` sees it) would have been a bigger behavioral change than needed — that counter also gates other things like `DraggableWindow`'s floating toolbar — and diverges from the established fix pattern for this exact bug class. The minimal, precedented fix (local Escape handler + stopPropagation) was used instead.

## Regression test

`tests/components/layout/WidgetLibrary.escapePropagation.test.tsx` — new test mirroring the existing `ClassRosterMenu.escapePropagation.test.tsx`/`RemoteControlMenu.escapePropagation.test.tsx` pattern.

**Before/after evidence:** independently re-verified by the orchestrator — reverted only the source fix (kept the test) and confirmed it fails on `dev-paul` (`onClose` never called), then restored the fix and confirmed it passes.

## Validation

`pnpm run validate` (type-check:all → lint → format:check → root tests [615 files / 7403 tests] → functions tests [41 files / 803 tests] → test:counts) and `pnpm run build` both pass clean.

## Risk

**contained** — single component, single new effect, no shared/cross-cutting surface.

---

### DIFF
--- a/components/layout/dock/WidgetLibrary.tsx
+++ b/components/layout/dock/WidgetLibrary.tsx
@@ -1,5 +1,6 @@
-import React, { forwardRef, useCallback, useMemo } from 'react';
+import React, { forwardRef, useCallback, useEffect, useMemo } from 'react';
 import { createPortal } from 'react-dom';
+import { isEscapeFromWidgetInput } from '@/utils/domHelpers';
 import {
   Bookmark,
   FolderPlus,
@@ -185,6 +186,18 @@ export const WidgetLibrary = forwardRef<HTMLDivElement, WidgetLibraryProps>(
       onClose,
       triggerRef ? [triggerRef] : []
     );
+
+    // Portalled outside any `.widget`/Modal ancestor — stop Escape reaching DashboardView's global handler.
+    useEffect(() => {
+      const handleKeyDown = (event: KeyboardEvent) => {
+        if (event.key !== 'Escape') return;
+        if (isEscapeFromWidgetInput(event)) return;
+        event.stopPropagation();
+        onClose();
+      };
+      document.addEventListener('keydown', handleKeyDown);
+      return () => document.removeEventListener('keydown', handleKeyDown);
+    }, [onClose]);
 
     const handleResetDock = useCallback(async () => {
       const confirmed = await showConfirm(
--- a/tests/components/layout/WidgetLibrary.escapePropagation.test.tsx
+++ b/tests/components/layout/WidgetLibrary.escapePropagation.test.tsx
@@ -1,0 +1,64 @@
+/**
+ * Regression test: WidgetLibrary (the "Add Widget" library overlay opened
+ * from the Dock) is portalled to document.body but has NO Escape-key
+ * handler at all — only useClickOutside. Because it never uses the shared
+ * `Modal` component, `useHasOpenModal()` (which DashboardView's global
+ * Escape handler checks to bail out) never registers it as open. Pressing
+ * Escape while the library is open therefore falls through to
+ * DashboardView's global `window`-level Escape handler, which — finding no
+ * typing field and no `.widget` ancestor for the library overlay — falls
+ * back to targeting the topmost z-index widget on the board and
+ * closing/minimizing it, all while the library overlay stays open and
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-16

### Summary
Adds a `document`-level Escape handler to `WidgetLibrary` (the Dock's "Add Widget" overlay), which is `createPortal`-ed to `document.body` and never used the shared `Modal` component — so `useHasOpenModal()` never saw it as open and Escape fell through to `DashboardView`'s global handler, minimizing an unrelated widget while the library stayed open. Correct fix, correct idiom, genuinely new instance (not a duplicate of any other open PR).

### Code Quality
✓ No issues found. Type-check clean on this branch; the new test passes **1/1**. Follows the established idiom exactly — `isEscapeFromWidgetInput()` bail before `stopPropagation()`, cleanup on unmount, `onClose` in the dep array.

### Widget Standards
✓ Not applicable — dock chrome, not widget front-face content. No scaling, registry, or appearance surface touched.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found — no Firestore, rules, or auth surface.

### Regression Risk
✓ Low. Two things worth checking on this fix class were checked rather than assumed:

**1. The `stopPropagation()` actually blocks the right handler.** `DashboardView`'s global handler is a *bubble-phase* `window` listener. A bubble-phase `document` listener fires first (document is below window in the bubble path), so `stopPropagation()` genuinely prevents it — confirmed by the test's `windowKeydownSpy` assertion, which is the real load-bearing assertion here, not the `onClose` one.

**2. It does not hijack Escape from the "Reset dock" confirmation dialog.** `handleResetDock` opens a `showConfirm` dialog, and a naive document-bubble handler would close the library out from under it. It doesn't: `DialogContainer.tsx:224` registers the confirm dialog's handler as **capture-phase on `window`** with `stopImmediatePropagation()` (`:216`), which fires strictly before this component's document-bubble listener and kills the event. Escape correctly cancels the dialog and leaves the library open. No change needed — noting it so a future refactor of either handler's phase doesn't silently break the pairing.

The effect is unconditional rather than guarded on an `isOpen` prop, which is correct here — `WidgetLibrary` is only mounted while open.

Also worth recording, since it's the reason this instance existed at all: the rejected alternative (wiring `WidgetLibrary` through the shared `useHasOpenModal()` counter) was the right thing to reject. That counter also gates `DraggableWindow`'s floating toolbar, so registering a non-modal overlay in it would have had side effects well outside this component.

### Missing Tests
✓ Adequate. One case, and it is non-tautological — it asserts on the `window`-level spy (the actual regression) rather than only on `onClose` (which a handler with no `stopPropagation()` would also satisfy).

### Overall
**Merge readiness:** Ready

Correct diagnosis, minimal fix, right idiom, and the one non-obvious interaction (nested confirm dialog) holds up under inspection. This is the only one of the four 2026-08-16 code PRs that is not a duplicate of an already-open PR — see the reviews on #2468 and #2471.

---

---

PR #2471, opened 2026-08-16
fix(admin): use local date, not UTC, when writing calendar blocked/default dates

OPS-PIvers:
## Root cause

`components/admin/CalendarConfigurationModal.tsx`'s "Add Date" (district blocked dates) and "Add Event" (building default events) buttons stamped new dates with `new Date().toISOString().split('T')[0]`. `toISOString()` converts to UTC before formatting, so it rolls to the next (or previous) calendar day depending on the admin's timezone offset — e.g. any evening hour in a US timezone produces tomorrow's date instead of today's.

The read side (`components/widgets/Calendar/Widget.tsx`'s `isBlocked` computation) was already fixed for this exact bug class and documents why via an inline comment; the admin write path was never brought in line with that fix, so write and read used different day-boundary math. A blocked date added in the evening would silently block the wrong day for every teacher's Calendar widget district-wide.

## Fix summary

Added a `getLocalDateISO()` helper (Y-M-D via `getFullYear()`/`getMonth()`/`getDate()`, mirroring the widget's already-correct approach) and used it at both write sites.

## Rejected band-aid

Clamping/validating the date after computation, or only fixing `addBlockedDate` and leaving `addDefaultEvent`'s copy of the same bug — rejected because both call sites share the identical root cause and the correct fix is the same one-line swap at each site.

## Regression test

`tests/components/admin/CalendarConfigurationModal.localDate.test.tsx` — pins `TZ=America/Chicago` and system time to `2026-08-17T02:30:00Z` (9:30pm local Aug 16, already Aug 17 in UTC), clicks "Add Date" / "Add Event", and asserts the resulting date input shows `2026-08-16`, not `2026-08-17`.

**Before/after evidence:** independently re-verified by the orchestrator — reverted only the source fix (kept the test) and confirmed both assertions fail on `dev-paul` (2/2 fail — the `2026-08-16` value never appears), then restored the fix and confirmed both pass.

## Validation

`pnpm run validate` (type-check:all → lint → format:check → root tests [615 files / 7404 tests] → functions tests [41 files / 803 tests] → test:counts) and `pnpm run build` both pass clean.

## Risk

**contained** — single component, one small pure helper function, no shared/cross-cutting surface.

---

### DIFF
--- a/components/admin/CalendarConfigurationModal.tsx
+++ b/components/admin/CalendarConfigurationModal.tsx
@@ -31,6 +31,11 @@ import { DockDefaultsPanel } from './DockDefaultsPanel';
 import { SettingsLabel } from '@/components/common/SettingsLabel';
 import { useDashboard } from '@/context/useDashboard';
 
+// Local Y-M-D, not toISOString() (UTC-shifts across the day boundary) — mirrors Calendar/Widget.tsx's isBlocked fix.
+function getLocalDateISO(date: Date = new Date()): string {
+  return `${date.getFullYear()}-${String(date.getMonth() + 1).padStart(2, '0')}-${String(date.getDate()).padStart(2, '0')}`;
+}
+
 interface CalendarConfigurationModalProps {
   isOpen: boolean;
   onClose: () => void;
@@ -230,7 +235,7 @@ export const CalendarConfigurationModal: React.FC<
   };
 
   const addBlockedDate = () => {
-    const today = new Date().toISOString().split('T')[0];
+    const today = getLocalDateISO();
     if (!config.blockedDates.includes(today)) {
       updateGlobal({ blockedDates: [...config.blockedDates, today] });
     }
@@ -238,7 +243,7 @@ export const CalendarConfigurationModal: React.FC<
 
   const addDefaultEvent = () => {
     const newEvent: CalendarEvent = {
-      date: new Date().toISOString().split('T')[0],
+      date: getLocalDateISO(),
       title: 'New Default Event',
     };
     updateBuilding({
--- a/tests/components/admin/CalendarConfigurationModal.localDate.test.tsx
+++ b/tests/components/admin/CalendarConfigurationModal.localDate.test.tsx
@@ -1,0 +1,84 @@
+import React from 'react';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+// Regression guard: "Add Date" (district blocked dates) and "Add Event"
+// (building default events) must stamp the ADMIN'S LOCAL calendar day, not
+// the UTC day. `new Date().toISOString().split('T')[0]` shifts to UTC and
+// rolls to the next day for any evening hour in a negative-UTC-offset
+// timezone (e.g. US Central) — the exact bug already fixed on the read side
+// in components/widgets/Calendar/Widget.tsx (see its `isBlocked` comment).
+// A blocked-date add made this way silently blocks the WRONG day.
+
+vi.mock('@/context/useAuth', () => ({
+  useAuth: () => ({ ensureGoogleScope: vi.fn().mockResolvedValue(null) }),
+}));
+
+vi.mock('@/context/useDashboard', () => ({
+  useDashboard: () => ({ addToast: vi.fn() }),
+}));
+
+vi.mock('@/hooks/useAdminBuildings', () => ({
+  useAdminBuildings: () => [{ id: 'b1', name: 'Building One' }],
+}));
+
+vi.mock('@/config/firebase', () => ({ db: {}, isAuthBypass: false }));
+
+vi.mock('firebase/firestore', () => ({
+  doc: vi.fn(() => ({})),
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-16

### Summary
Fixes a real UTC-shift bug: `CalendarConfigurationModal`'s "Add Date" / "Add Event" stamped `new Date().toISOString().split('T')[0]`, which rolls to the next day for any evening hour in a negative-UTC-offset timezone. But **#2454 (open since 2026-08-14) already fixes exactly this**, in the same two call sites of the same file — and does it by reusing the repo's existing tested helper rather than hand-rolling a third copy.

### Code Quality
⚠️ **`getLocalDateISO()` duplicates `utils/localDate.ts`'s `getLocalIsoDate()`.**

The new file-local helper is byte-equivalent in behavior to a util that already exists, is already unit-tested (`utils/localDate.test.ts`, 6 cases including leap-day and year-boundary), and is already imported at 8+ call sites (`Announcements/Widget.tsx`, `AbsentButton.tsx`, `AbsentStudentsModal.tsx`, `random/RandomWidget.tsx`, …).

The two names differ by a one-character transposition — `getLocalDate**ISO**` vs `getLocalIso**Date**` — which is close enough to read as the same symbol at a glance and makes a future "why are there two of these?" sweep harder than it needs to be.

```diff
-// Local Y-M-D, not toISOString() ...
-function getLocalDateISO(date: Date = new Date()): string {
-  return `${date.getFullYear()}-${String(date.getMonth() + 1).padStart(2, '0')}-${String(date.getDate()).padStart(2, '0')}`;
-}
+import { getLocalIsoDate } from '@/utils/localDate';
```

Otherwise clean — type-check clean on this branch, new test file **2/2** green.

### Widget Standards
✓ Not applicable.

### Admin Config Standards
✓ Correct level. This is an admin-side writer (District Blocked Dates / Default Building Events) and the fix stays in `components/admin/CalendarConfigurationModal.tsx` — no user-level settings surface touched.

### Firestore & Security
✓ No issues found.

### Regression Risk
⚠️ **Duplicate of #2454 — the two cannot both merge.**

Verified rather than inferred:

```
$ git merge-tree --write-tree --name-only \
    origin/nightly/admin-config-2026-08-14 origin/nightly/admin-config-2026-08-16
CONFLICT (content): Merge conflict in components/admin/CalendarConfigurationModal.tsx
```

**#2454 is the stronger of the two, on both halves of the bug:**

| | #2454 (2026-08-14) | #2471 (this PR) |
|---|---|---|
| admin **writer** | `getLocalIsoDate()` from `@/utils/localDate` | new file-local `getLocalDateISO()` |
| widget **reader** (`Calendar/Widget.tsx` `isBlocked`) | refactored onto the **same** helper (`d96974e`) | left with its own inline hand-rolled copy |

That reader refactor is the substantive difference. This bug was a producer/consumer drift — the admin panel wrote UTC while the widget read local. #2454 puts both sides on one shared, tested helper so they provably cannot drift apart again; this PR leaves two independent inline implementations that happen to agree today. Its own code comment (*"mirrors Calendar/Widget.tsx's isBlocked fix"*) is describing that duplication as the design.

Same root cause as #2468's duplication of #2452: this run branched from `dev-paul` at `efc375e` while #2454 was still unmerged. See the review on #2468 — this is the third and fourth duplicate produced by that unmerged-nightly pattern, which `docs/routines/debugger.md` has now diagnosed twice without the fix landing.

### Missing Tests
✓ Adequate on its own terms — the `TZ=America/Chicago` + pinned-system-time setup genuinely exercises the bug window, and the `queryByDisplayValue('2026-08-17')` negative assertion is a good touch (it proves the test would catch a regression rather than passing vacuously).

One small hygiene note in `afterEach`, independent of the duplication question: `process.env.TZ = originalTz` writes the **string** `"undefined"` when `TZ` was unset in the parent environment, since `process.env` coerces assignments to strings. Harmless under Vitest's per-file isolation, but `if (originalTz === undefined) delete process.env.TZ; else process.env.TZ = originalTz;` is the leak-free form.

### Overall
**Merge readiness:** Needs changes — recommend **closing in favour of #2454**

The diagnosis is right and the tests are good, but #2454 fixes strictly more of the same bug (both sides, not just the writer) and does it without adding a fourth copy of a helper that already exists. Closing this and merging #2454 loses nothing except the duplicated helper.

If this one is kept instead, the minimum change is swapping `getLocalDateISO()` for the existing `getLocalIsoDate()` import and porting #2454's `Calendar/Widget.tsx` reader refactor across.

---

---

OPS-PIvers:
Confirmed via the review's evidence: #2454 (opened 2026-08-14, still open, mergeable) fixes the same bug more completely — it routes both this admin-write call site *and* the `Calendar/Widget.tsx` reader through the existing shared `getLocalIsoDate()` helper, closing the producer/consumer drift at its root rather than leaving two independent inline implementations that happen to agree today. Closing this in favor of #2454.

Root cause: same as #2468 — this run's worktree branched from `dev-paul` at `efc375e` while #2454 was still unmerged. Re-pointing the `docs/routines/debugger.md` attribution from this PR to #2454 in a follow-up commit to #2472.

---

PR #2477, opened 2026-08-17
test(functions): cover transcribeVideoWithGemini + generateGuidedLearning

OPS-PIvers:
## Summary

- Adds 14 execution-path tests to `functions/src/index.test.ts` for the two Cloud Functions in `functions/src/aiGeneration.ts` that previously had zero test coverage: `transcribeVideoWithGemini` and `generateGuidedLearning`.
- `transcribeVideoWithGemini` (8 tests): unauthenticated, missing email, permission-doc-missing, feature disabled, admin-only `accessLevel` rejection, beta `accessLevel` rejection, admin caller reaching the real (mocked) `ai.models.generateContent` call, and an invalid video URL.
- `generateGuidedLearning` (6 tests): unauthenticated, missing email, non-admin rejection, no images provided, too many images, and an admin caller reaching the AI call.
- Extends the shared Firestore mock in `index.test.ts` with a new hoisted `audioTranscriptionPermDocGet` stub (mirroring the existing `geminiConfigDocGet` pattern) to serve the `global_permissions/video-activity-audio-transcription` doc that `transcribeVideoWithGemini`'s access-level gate reads.
- Updates `docs/scheduled-tasks/test-coverage.md` to record this action and close Priority 1 of the tracked "functions/src/ Cloud Function execution paths lack test coverage" MEDIUM item — only the Priority 2 `spotifyOAuth.ts` sub-target remains open.

## Test plan

- [x] `pnpm exec vitest run src/index.test.ts` (from `functions/`) — 90/90 passing (76 pre-existing + 14 new, nothing broken)
- [x] `pnpm run type-check:all` — exit 0
- [x] `eslint functions/src/index.test.ts --max-warnings 0` — exit 0
- [x] `prettier --check functions/src/index.test.ts` — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_015ZChYJKSxgNjTMBA562m8K)_


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-17

### Summary
Closes the Priority 1 sub-target of the standing "Cloud Function execution paths lack test coverage" item by adding 273 lines to `functions/src/index.test.ts` covering `transcribeVideoWithGemini` (7 cases) and `generateGuidedLearning` (6 cases), plus the routine's daily/weekly audit journal updates across seven `docs/scheduled-tasks/*.md` files.

### Code Quality
✓ No issues found in the production surface — this PR adds **no production code**, only tests and docs.
- The `audioTranscriptionPermDocGet` hoisted mock is wired correctly into `mockFirestore`'s `global_permissions` branch: the `doc(id)` dispatch now routes `gemini-functions` and `video-activity-audio-transcription` to separate independently-resettable mocks, with the pre-existing `{exists: false}` default preserved for everything else. No cross-test bleed — each `beforeEach` resets both.
- Test structure mirrors the existing `generateVideoActivity — accessLevel enforcement` block, so this extends the file's convention rather than introducing a second style.
- Gate coverage is genuinely thorough: unauthenticated → no email → permission doc missing → feature disabled → `accessLevel: 'admin'` non-admin → `accessLevel: 'beta'` non-beta → admin success, each asserting `generateContentMock` was **not** called. That last assertion is the valuable one — it proves no gate leaks to the network.

### Widget Standards
✓ Not applicable — Cloud Functions tests and routine docs only.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found, and this is the PR's main value: the permission gates on two AI callables are now regression-locked. `transcribeVideoWithGemini`'s `video-activity-audio-transcription` accessLevel gate and `generateGuidedLearning`'s admin-only gate both had zero coverage before. No new collection, no rules change, no secrets in the diff.

### Regression Risk
✓ Low risk. Verified locally: `vitest run src/index.test.ts` in `functions/` → **90 tests passing**. `functions/src/index.ts` itself is untouched, so no Cloud Function signature changed.

One process note, not a code issue: this PR's head branch is `scheduled-tasks`, which is also where the nightly PR-review routine appends its own run log. Log commits will land on this open PR's branch and show up in its diff. Harmless, but it makes the PR's file list drift from its stated scope over time — worth branching future routine work off `dev-paul` instead if the mixing gets confusing.

### Missing Tests
Coverage of the gate paths is complete. Two notes on assertion quality:

1. **`await expect(result).rejects.not.toThrow(...)` is close to vacuous.** In the two "reaches the AI call" tests, this asserts only that the promise rejects *with some other message* — it passes for essentially any failure downstream of the gate. The real assertion in both tests is `expect(generateContentMock).toHaveBeenCalledTimes(1)`, which is sound. The fragile part is that the test *depends on the handler ultimately rejecting*: `generateContentMock` returns something the parser can't consume, so the call fails after the SDK. If that mock is ever made to return valid parseable content, `.rejects` flips to a failure and these two tests break for a reason unrelated to what they're testing. Asserting on `generateContentMock`'s call arguments (model id, `fileData.fileUri`) instead would be both stronger and more robust. This mirrors the existing `generateVideoActivity` block, so it's a pre-existing convention rather than something this PR invented.
2. **Same promise awaited twice.** `await expect(result).rejects...` runs twice against one rejected promise in both success tests. It works, but it's an easy source of unhandled-rejection noise if the surrounding structure changes.

Neither blocks merge; both are worth folding into the shared pattern next time this block is extended.

### Overall
**Merge readiness:** Ready
Real coverage added to two previously-untested AI callables' permission gates, consistent with the file's existing patterns; the weak `.rejects.not.toThrow()` assertions are inherited convention worth tightening rather than a defect in this PR.

---

---

OPS-PIvers:
Addressed both review notes in `b757065a`: the two "reaches the AI call" tests now assert on `generateContentMock`'s actual call arguments (model id + the video `fileData`/image `inlineData` payload) instead of the vacuous `rejects.not.toThrow(...)` pattern, and each awaits the handler once (via `.catch(() => undefined)`) rather than awaiting the same rejected promise twice. Re-verified: `vitest run src/index.test.ts` → 90/90 passing, `type-check:all` exit 0, `eslint --max-warnings 0` clean, `prettier --check` clean.

---

OPS-PIvers:
All three review notes addressed.

**1 & 2. Weak `.rejects.not.toThrow()` assertions in the new blocks — already fixed.** The review was generated against `2f550e7d`; `b757065` on this branch had already replaced both. The `transcribeVideoWithGemini` and `generateGuidedLearning` success tests now swallow the incidental downstream rejection and assert on what actually reached the AI call — model id and `fileData.fileUri` — exactly as suggested, which also removes the double-await on one promise.

**The inherited `generateVideoActivity` convention — now fixed too, in `b65e074`.** The review scoped this as "pre-existing convention worth tightening next time this block is extended"; since this PR extends the file with the better pattern, leaving three vacuous assertions next to it was the wrong shape to ship. Both tests rewritten:

- *admin* test — reaches the AI call, so it now asserts `generateContentMock` call arguments rather than only the call count.
- *beta user* test — this one **cannot** assert on the mock: a beta (non-admin) caller clears the accessLevel gate but then rejects in the usage-limit step, so `generateContent` is genuinely never reached. It now captures the rejection and pins its identity (`err.name` is `TypeError`, not `permission-denied`) plus `expect(generateContentMock).not.toHaveBeenCalled()`.

Worth flagging one thing the verification caught, since it bears on this file generally: my first attempt asserted `err.code` was undefined — which passes *always*, because this file's `HttpsError` mock (`src/index.test.ts:367`) sets `this.name = code` and never sets `.code`. Negative-control runs caught it; both rewritten tests were then re-verified by deliberately breaking them (removing the caller from `betaUsers` → beta test fails; changing the expected `fileUri` → admin test fails). Anything asserting on `HttpsError` codes in this file needs `name`, not `code`.

`vitest run src/index.test.ts` → **90 tests passing**; tsc ✓ eslint ✓ prettier ✓.

**3. The `scheduled-tasks` branch note — acknowledged, no action.** Correct that routine log commits landing on this branch make the PR's file list drift from its stated scope. Not re-branching mid-PR: this one is already open with its docs commits in the diff, and moving it now would orphan the review history here for a purely cosmetic gain. Worth adopting for the *next* routine run rather than retrofitting this one.

---

OPS-PIvers (commented):
## Automated Code Review — follow-up (delta since `2f550e7`)

**Date:** 2026-08-18

### Summary
Seven new commits. Two directly address my 2026-08-17 review; five are unrelated `docs/scheduled-tasks/*` audit-journal updates that landed on this branch.

- `b757065` / `b65e024` — replace the near-vacuous `await expect(result).rejects.not.toThrow(…)` assertions with ones that pin what actually happened, exactly as suggested: the "reaches the AI call" tests now assert on `generateContentMock`'s **call arguments** (`contents[0].parts[1]` → `{ fileData: { fileUri, mimeType } }`), and the gate tests assert the rejection is *not* `permission-denied` plus `generateContentMock` was never invoked.
- `9808190`, `43322bd`, `b96eb24`, `e34b86c`, `371d234`, `729918e` — Tuesday audit-journal entries (dependency-audit, widget-registry, css-scaling, skill-freshness, typescript-eslint).

Verified locally on `729918e`: `vitest run src/index.test.ts` in `functions/` → **90 tests passing**.

### Code Quality
✓ No issues found in the production surface — this PR still adds **no production code**, only tests and docs.

The assertion rework is a genuine improvement and gets the important half right: asserting the AI call's arguments rather than just its call count means the test now fails if the wrong model or a mangled `fileUri` reaches Vertex, which is the thing worth protecting.

One new note on the other half, since it trades one coupling for another: `expect(err?.name).toBe('TypeError')` pins the *incidental* failure mode — a `TypeError` thrown because this mock's `ai_usage` collection doesn't stub `.doc()`. That's strictly stronger than `.not.toThrow()` (which passed for literally any failure), but it now asserts on mock incompleteness rather than on handler behavior: if someone later stubs `ai_usage.doc()` to make an unrelated test work, this assertion breaks for a reason that has nothing to do with the permission gate. The `expect(err?.name).not.toBe('permission-denied')` and `expect(generateContentMock).not.toHaveBeenCalled()` lines are the two that carry real meaning; the `toBe('TypeError')` is the fragile one and could be dropped without weakening the test.

✓ Not applicable.

### Firestore & Security
✓ No issues found, and this remains the PR's core value: the `video-activity-audio-transcription` accessLevel gate and `generateGuidedLearning`'s admin-only gate are now regression-locked, with the delta strengthening the proof that no gate leaks to the network.

### Regression Risk
Low for the code; `functions/src/index.ts` is still untouched, so no Cloud Function signature changed.

**The branch-hygiene note from my previous review has now materialized.** I flagged that `scheduled-tasks` doubles as the nightly routines' log branch, so unrelated commits would drift into this PR's diff. Since then, five audit-journal commits have landed here, and the diff now spans `functions/src/index.test.ts` plus five `docs/scheduled-tasks/*.md` files across three unrelated audit domains. Nothing is *wrong* — the docs changes are inert — but the PR's title ("test(functions): cover transcribeVideoWithGemini + generateGuidedLearning") no longer describes most of its file list, and the drift will continue for as long as this stays open. Two clean options: merge it soon and let the journals continue on the branch afterward, or move the test work to a dedicated branch off `dev-paul` and let `scheduled-tasks` be journals-only.

### Missing Tests
✓ Adequate coverage. Gate-path coverage was already complete; the delta upgrades assertion *quality*, which was the only open item. The `.rejects.not.toThrow()` pattern still exists elsewhere in the file's older `generateVideoActivity` block — worth the same treatment whenever that block is next touched, but out of scope here.

### Overall
**Merge readiness:** Ready
The prior review's assertion-quality note is addressed and the coverage is real. Worth dropping the `toBe('TypeError')` line, and worth deciding soon whether this branch is a feature branch or a journal branch — it's currently both.

---

---

OPS-PIvers:
Dropped the `toBe('TypeError')` line in `de15643d` — agreed it was pinning mock incompleteness rather than handler behavior. The `not.toBe('permission-denied')` + `generateContentMock` not-called assertions carry the real signal and are unchanged. Re-verified: 90/90 tests passing, type-check/eslint/prettier all clean.

On the branch-hygiene note: confirmed, `scheduled-tasks` is the shared branch multiple scheduled sessions push routine journal/audit commits to, so this PR's diff will keep accumulating unrelated `docs/scheduled-tasks/*.md` entries for as long as it stays open — that's expected given how the routine is wired, not a mistake in this PR. Deferring to a human call on whether to merge soon vs. move future test work to a dedicated branch, per the original review's suggestion.

---

PR #2479, opened 2026-08-17
fix(layout): use local-date helper for screenshot/annotation download filenames

OPS-PIvers:
## Root cause

`components/common/DraggableWindow.tsx` (screenshot filename) and `components/layout/AnnotationOverlay.tsx` (annotation PNG download filename) both built their filenames with `new Date().toISOString().split('T')[0]` — the **UTC** calendar date — instead of the teacher's local date. For every timezone west of UTC (all of the Americas), the last several hours of the local day fall on the *next* UTC calendar date, so a screenshot or annotation saved in the evening downloads with tomorrow's date baked into the filename.

This is a known, previously-fixed bug class in this repo — `utils/localDate.ts`'s `getLocalIsoDate()` (and `Schedule/utils.ts`'s `getTodayStr()`) already exist specifically to prevent this (also just fixed twice in `CalendarConfigurationModal.tsx`). These two call sites were simply never migrated to the shared helper.

## Fix

Switched both call sites to the existing shared `getLocalIsoDate()` helper — no new logic invented, root cause fixed at the source.

## Band-aid rejected

Patching the string post-hoc (subtracting the local UTC offset from the `Date` before calling `toISOString()`) — rejected because the codebase already has a canonical, tested local-date helper; adding a second, divergent implementation of the same fix would just create more drift.

## Regression tests

- `components/common/DraggableWindow.test.tsx`: mocks `getLocalIsoDate` to diverge from the UTC-computed date and asserts the screenshot filename matches the local date, not `toISOString()`'s date.
- `tests/components/layout/AnnotationOverlay.test.tsx`: clicks "Download PNG" and asserts the generated filename matches the mocked local date.

Both mock the `getLocalIsoDate` helper directly (not the system clock) because `vitest.config.ts` pins `process.env.TZ = 'UTC'` for all tests, so real local/UTC getters can never diverge in-suite.

**Before/after evidence:** independently re-verified by the orchestrator — reverting both files to `dev-paul` and running the two regression tests fails both with the exact UTC-vs-local mismatch described (`expected 'Classroom-Clock-2026-03-06' to be 'Classroom-Clock-2026-03-05'`, `expected 'Annotation-2026-08-17.png' to be 'Annotation-2026-03-05.png'`). Restoring the fix passes all tests in both files.

## Validation

`pnpm run validate` and `pnpm run build` — both green, independently re-run by the orchestrator on the final commit.

## Risk

**Contained** — two call sites swapped to an existing, already-tested shared helper. No behavior change outside filename generation.

---
_Generated by [Claude Code](https://claude.ai/code/session_017noYuFMHKAK5uzaf9GNdEQ)_

### DIFF
--- a/components/common/DraggableWindow.test.tsx
+++ b/components/common/DraggableWindow.test.tsx
@@ -31,21 +31,48 @@ import {
 } from './modalStore';
 
 // Mock dependencies
-const { mockTakeScreenshot } = vi.hoisted(() => ({
+const { mockTakeScreenshot, mockUseScreenshot } = vi.hoisted(() => ({
   mockTakeScreenshot: vi.fn(),
+  mockUseScreenshot: vi.fn(),
 }));
 
 vi.mock('@/hooks/useScreenshot', () => ({
-  useScreenshot: () => ({
-    takeScreenshot: mockTakeScreenshot,
-    isFlashing: false,
-    isCapturing: false,
-  }),
+  // Records the (contentRef, fileName, options) args it's called with so
+  // tests can assert on the generated screenshot filename.
+  useScreenshot: (...args: unknown[]) => {
+    mockUseScreenshot(...args);
+    return {
+      takeScreenshot: mockTakeScreenshot,
+      isFlashing: false,
+      isCapturing: false,
+    };
+  },
 }));
 
 vi.mock('@/hooks/useClickOutside', () => ({
   useClickOutside: vi.fn(),
 }));
+
+// Controllable getLocalIsoDate for the UTC/local-date-divergence regression
+// test below. Mirrors the getTodayStr pattern in ScheduleWidget.test.tsx —
+// the test environment pins process.env.TZ to 'UTC' (see vitest.config.ts),
+// so real local getters can never diverge from toISOString() there; mocking
+// the helper directly is the only way to simulate the divergence a teacher
+// in a UTC-negative timezone actually sees.
+const { mockGetLocalIsoDate, defaultGetLocalIsoDate } = vi.hoisted(() => ({
+  mockGetLocalIsoDate: vi.fn<() => string>(),
+  defaultGetLocalIsoDate: { current: (() => '') as () => string },
+}));
+
+vi.mock('@/utils/localDate', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@/utils/localDate')>();
+  defaultGetLocalIsoDate.current = actual.getLocalIsoDate;
+  mockGetLocalIsoDate.mockImplementation(actual.getLocalIsoDate);
+  return {
+    ...actual,
+    getLocalIsoDate: mockGetLocalIsoDate,
+  };
+});
 
 interface GlassCardProps {
   children: React.ReactNode;
@@ -202,6 +229,10 @@ describe('DraggableWindow', () => {
       return 0;
     });
     vi.clearAllMocks();
+    // Default: getLocalIsoDate returns the real local date so existing tests
+    // (which don't care about the screenshot filename) are unaffected.
+    mockGetLocalIsoDate.mockReset();
+    mockGetLocalIsoDate.mockImplementation(defaultGetLocalIsoDate.current);
     // Reset SettingsPanel render spy before each test
     settingsPanelRenderProps.length = 0;
     // Setup default spy to return null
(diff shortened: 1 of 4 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-17

### Summary
Replaces `new Date().toISOString().split('T')[0]` with the existing `getLocalIsoDate()` helper in two download-filename sites: the widget screenshot name in `DraggableWindow.tsx` and the annotation PNG name in `AnnotationOverlay.tsx`. Both previously stamped the UTC date, so for every UTC-negative timezone (all US school districts) a file saved during the late afternoon or evening carried tomorrow's date. Same bug class already fixed twice in `CalendarConfigurationModal.tsx`.

### Code Quality
✓ No issues found.
- Both imports use the `@/` alias (`@/utils/localDate`), consistent with the surrounding import blocks.
- Comments are one-to-three lines and explain *why* rather than restating the code — within the CLAUDE.md convention, and the `AnnotationOverlay` one is a single line.
- `getLocalIsoDate()` is called during render in `DraggableWindow` (as the previous code was), so the filename reflects render time rather than click time. Unchanged behavior, and irrelevant at day granularity except for a board left open across local midnight — not worth restructuring.

### Widget Standards
✓ Not applicable — `DraggableWindow.tsx` is the shared window chrome in `components/common/`, not widget front-face content, so no container-query sizing rules apply. No registry, `types.ts`, or defaults surface is touched.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found — filename generation only; no persisted data, no collection access, no auth surface.

### Regression Risk
✓ Low risk. Verified locally: `vitest run components/common/DraggableWindow.test.tsx tests/components/layout/AnnotationOverlay.test.tsx` → **3 files / 80 tests passing**. The change is confined to a display string; nothing stored in Firestore or keyed by date is affected, so there is no migration or backfill concern (unlike the `CalendarConfigurationModal` instances, which touched business data).

### Missing Tests
✓ Adequate coverage. Both call sites gained tests in this PR (`DraggableWindow.test.tsx` +76, new `tests/components/layout/AnnotationOverlay.test.tsx` +63), which is stronger than the precedent — the earlier `CalendarConfigurationModal` fixes of the same bug class shipped without dedicated filename tests.

One remaining instance of this anti-pattern is **not** in this PR and is correctly logged as separate: `components/admin/Announcements/Widget.tsx`'s CSV export filename still uses `new Date().toISOString().split('T')[0]`. Tracked as a LOW backlog row in #2483, so no action needed here — flagging only so it isn't assumed covered by "the UTC-date sweep."

### Overall
**Merge readiness:** Ready
Minimal, correct fix routed through the established helper, with better test coverage than prior fixes of the same bug class.

---

---

OPS-PIvers:
Reviewed the one open note; no change pushed, and I think that's correct.

**`components/admin/Announcements/Widget.tsx` CSV export filename still uses `new Date().toISOString().split('T')[0]`.** Confirmed present, and it is the same UTC-date bug class this PR fixes. Not folding it in:

- It is already tracked as a LOW backlog row in #2483, so it is not at risk of being lost.
- This PR is scoped to the two *download-filename* sites in `DraggableWindow.tsx` / `AnnotationOverlay.tsx` and carries dedicated tests for both. Pulling in an admin-panel CSV export widens the surface into a different area with a different test setup, for no gain over doing it as its own change.

The note's framing is the right one — it was flagged so the file isn't wrongly assumed covered by "the UTC-date sweep," not as something this PR should absorb. Recording that explicitly here so a later reader doesn't re-derive it.

Everything else on the review was ✓; no other action needed.

---

OPS-PIvers (commented):
## Automated Code Review — follow-up (no new commits)

**Date:** 2026-08-18

The head is unchanged at `5b9ec96` since my 2026-08-17 review, so that assessment stands in full: **Ready**, correct fix routed through `getLocalIsoDate()`, with better test coverage than the earlier fixes of the same bug class. Nothing below changes that.

### One new finding — a third untracked instance of this anti-pattern

My previous review noted that `components/admin/Announcements/Widget.tsx` still carries the UTC-date filename and is correctly logged as a LOW backlog row in #2483. Re-running that sweep against `dev-paul` tonight turns up a **fourth production call site that is not tracked anywhere**:

```
components/widgets/PollWidget/Settings.tsx:141
  `Poll_Results_${new Date().toISOString().split('T')[0]}.csv`
```

I checked #2483's backlog row and it names only `components/admin/Announcements/Widget.tsx`. Grepping `docs/routines/debugger.md` for this call site returns nothing — so unlike the Announcements one, this instance has never been logged.

It's the same low severity as the Announcements case (a downloaded filename's date near local midnight, not stored or business data), and notably it's the *same feature area*: both are poll-results CSV exports, so a teacher exporting the same poll from the widget settings and from the admin panel gets two filenames that can disagree about the date. Worth folding into the existing backlog row rather than opening a new one — the row should read "two remaining instances", not one.

Full sweep result on `dev-paul` for `toISOString().split('T')[0]`, for the record:
- `components/common/DraggableWindow.tsx:601` — **fixed by this PR**
- `components/layout/AnnotationOverlay.tsx:507` — **fixed by this PR**
- `components/admin/Announcements/Widget.tsx:582` — tracked, LOW backlog in #2483
- `components/widgets/PollWidget/Settings.tsx:141` — **untracked**
- Two test-file occurrences (`Calendar/Widget.test.tsx`, `LunchCount/Widget.test.tsx`) — intentional, they encode the old buggy behavior for comparison

No action needed on this PR — it does exactly what it says and shouldn't grow to absorb the remaining two. This is a backlog-accuracy note.

### Overall
**Merge readiness:** Ready (unchanged)

---

---

OPS-PIvers:
Confirmed and folded into the existing backlog row in `5bc6fbb` on #2483 — `PollWidget/Settings.tsx:141` now tracked alongside `Announcements/Widget.tsx:582` as the same low-priority, two-instance item. Agreed this PR shouldn't grow to absorb either.

---

PR #2480, opened 2026-08-17
fix(imageWorker): check the alpha byte, not the whole pixel, when trimming

OPS-PIvers:
## Root cause

`trimImageData()` in `utils/imageWorker.ts` (backs the "trim transparent whitespace" step run after every sticker/image upload, via `hooks/useImageUpload.ts`) tested opaqueness with `data32[i] !== 0` — checking whether the *entire* 32-bit RGBA word was non-zero, not whether the **alpha byte specifically** was non-zero. The adjacent comment correctly explained "alpha is the most significant byte" but then implemented the wrong check. A fully-transparent pixel (alpha = 0) that still carries non-zero R/G/B is counted as "found" content and pollutes the trim bounding box.

This isn't theoretical: `removeBackgroundFloodFill` (the flood-fill fallback for AI background removal, same file) explicitly zeroes only the alpha byte (`data[offset + 3] = 0`) and leaves R/G/B untouched — its own output is the exact shape that defeats the trim step. After the flood-fill fallback runs, `trimImageWhitespace` trims little to nothing, because the "background" pixels still register as opaque.

## Fix

One-line fix: isolate the alpha byte with `>>> 24` before comparing, instead of comparing the whole 32-bit word.

## Band-aid rejected

Patching only `removeBackgroundFloodFill` to also zero R/G/B when clearing alpha — rejected because the trim check is wrong in general. Any future producer of transparent-but-colored pixels (a different background-remover, a hand-edited PNG, browser decode quirks) would hit the same silent failure. Fixed the actual predicate instead.

## Regression test

`utils/imageWorker.test.ts` (new), 3 cases against the now-exported `trimImageData`:
1. Transparent-but-colored border pixels around an opaque block trim correctly to the opaque region, not the full canvas.
2. A fully-transparent image (even with stray RGB) reports `found: false`.
3. A fully-opaque image is unchanged (no regression on the normal path).

**Before/after evidence:** independently re-verified by the orchestrator — reverting the fix (keeping the function exported for testability) fails cases 1 and 2 with the exact predicted mismatch (`expected 0 to be 2`, `expected true to be false`). Restoring the fix passes all 3.

## Validation

`pnpm run validate` and `pnpm run build` — both green, independently re-run by the orchestrator on the final commit.

## Risk

**Contained** — one-line predicate fix in a single pure function, isolated to `utils/imageWorker.ts`.

---
_Generated by [Claude Code](https://claude.ai/code/session_017noYuFMHKAK5uzaf9GNdEQ)_

### DIFF
--- a/utils/imageWorker.test.ts
+++ b/utils/imageWorker.test.ts
@@ -1,0 +1,95 @@
+import { describe, it, expect } from 'vitest';
+import { trimImageData } from './imageWorker';
+
+/**
+ * Builds an RGBA Uint8ClampedArray for a `width`x`height` image. Every pixel
+ * defaults to a fully-transparent pixel that still carries non-zero color
+ * channels (`bgColor`) — this mirrors real-world output from
+ * `removeBackgroundFloodFill`, which zeroes only the alpha byte
+ * (`data[offset + 3] = 0`) and leaves R/G/B untouched. `opaqueRect` paints an
+ * inclusive [x0,x1] x [y0,y1] block fully opaque with `fgColor`.
+ */
+function buildImage(
+  width: number,
+  height: number,
+  opaqueRect: { x0: number; x1: number; y0: number; y1: number },
+  bgColor: [number, number, number] = [100, 50, 25],
+  fgColor: [number, number, number] = [10, 10, 10]
+): Uint8ClampedArray {
+  const data = new Uint8ClampedArray(width * height * 4);
+  for (let y = 0; y < height; y++) {
+    for (let x = 0; x < width; x++) {
+      const offset = (y * width + x) * 4;
+      const inRect =
+        x >= opaqueRect.x0 &&
+        x <= opaqueRect.x1 &&
+        y >= opaqueRect.y0 &&
+        y <= opaqueRect.y1;
+      if (inRect) {
+        data[offset] = fgColor[0];
+        data[offset + 1] = fgColor[1];
+        data[offset + 2] = fgColor[2];
+        data[offset + 3] = 255;
+      } else {
+        // Fully transparent, but NOT zeroed color channels — this is the
+        // shape flood-fill background removal actually produces.
+        data[offset] = bgColor[0];
+        data[offset + 1] = bgColor[1];
+        data[offset + 2] = bgColor[2];
+        data[offset + 3] = 0;
+      }
+    }
+  }
+  return data;
+}
+
+describe('trimImageData', () => {
+  it('trims to the opaque content, ignoring color data left behind under transparent pixels', () => {
+    // 10x10 image; only the 2x2 block at (4,4)-(5,5) is opaque. Every other
+    // pixel is transparent but still carries non-zero RGB — exactly what
+    // removeBackgroundFloodFill's alpha-only zeroing produces.
+    const width = 10;
+    const height = 10;
+    const data = buildImage(width, height, { x0: 4, x1: 5, y0: 4, y1: 5 });
+
+    const result = trimImageData(data, width, height);
+
+    expect(result.found).toBe(true);
+    // Content is at (4,4)-(5,5); padding of 2 clamped to the canvas gives
+    // [2,7] on both axes → a 6x6 trimmed box, not the full 10x10 canvas.
+    expect(result.minX).toBe(2);
+    expect(result.minY).toBe(2);
+    expect(result.width).toBe(6);
+    expect(result.height).toBe(6);
+  });
+
+  it('reports not-found for an image that is fully transparent, even with stray color data', () => {
+    const width = 5;
+    const height = 5;
+    // Rect outside the canvas bounds -> nothing opaque.
+    const data = buildImage(width, height, { x0: 99, x1: 99, y0: 99, y1: 99 });
+
+    const result = trimImageData(data, width, height);
+
+    expect(result.found).toBe(false);
+  });
+
+  it('still detects a fully-opaque image spanning the whole canvas', () => {
+    const width = 4;
+    const height = 4;
+    const data = buildImage(width, height, {
+      x0: 0,
+      x1: width - 1,
+      y0: 0,
+      y1: height - 1,
+    });
+
+    const result = trimImageData(data, width, height);
+
+    expect(result.found).toBe(true);
+    expect(result.minX).toBe(0);
+    expect(result.minY).toBe(0);
+    expect(result.width).toBe(width);
+    expect(result.height).toBe(height);
+  });
+});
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-17

### Summary
Fixes `trimImageData()` in `utils/imageWorker.ts`, which tested `data32[i] !== 0` — the whole 32-bit RGBA word — when deciding whether a pixel counts as content. Any transparent-but-colored pixel therefore read as opaque, which is exactly what the sibling `removeBackgroundFloodFill()` in the same file produces (it zeroes only `data[offset + 3]` and leaves R/G/B intact). Result: after background removal, the trim bounds expanded to the full image and whitespace trimming silently did nothing. One-line fix (`>>> 24`), plus a new 95-line test file.

### Code Quality
✓ No issues found.
- The fix is correct: `>>> 24` isolates the most-significant byte, which is alpha in the little-endian `0xAA_BB_GG_RR` layout the original comment already assumed.
- Operator precedence is right — `>>>` binds tighter than `!==`, so `data32[i] >>> 24 !== 0` parses as `(data32[i] >>> 24) !== 0`. No parenthesization bug.
- `trimImageData` changed from module-private to `export`ed purely for testability. That's a real (small) API-surface widening rather than a dead export — the new test file is the only consumer — and it's the standard trade for testing a worker internal. Importing the module in a test does execute the top-level `self.onmessage = …`, which is safe because `vitest.config.ts` sets `environment: 'jsdom'` (verified: the suite passes). Worth knowing that switching this file's test environment to `node` would break the import.

### Widget Standards
✓ Not applicable — worker utility, no widget surface touched.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found — pure in-memory pixel math; no data access, no auth surface, no secrets.

### Regression Risk
✓ Low risk. Verified locally: `vitest run utils/imageWorker.test.ts` → **3 tests passing**. Does not touch `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, or `functions/src/index.ts`.

One pre-existing caveat this PR inherits rather than introduces: `Uint32Array` is **platform-endian**, not little-endian, so `>>> 24` reads alpha only on a little-endian host. Every browser SpartBoard targets is little-endian, and the original code carried the same assumption in its comment, so this is not a regression — noting it only because the new comment states the layout as fact. Checked the sibling `removeBackgroundFloodFill()`: it indexes bytes directly (`data[offset + 3]`), so it has no equivalent exposure.

### Missing Tests
✓ Adequate coverage. The new `utils/imageWorker.test.ts` is well-targeted — `buildImage()` deliberately paints the transparent background with **non-zero** RGB (`[100, 50, 25]`), which is precisely the shape that defeats the old check, so these tests genuinely fail against the pre-fix code rather than passing vacuously.

Gap worth noting, not blocking: only `trimImageData` is covered. `removeBackgroundFloodFill` — the function that *produces* the pixel shape this fix exists to handle — still has no tests, so the two halves of the pipeline aren't verified end-to-end. A single test running flood-fill then trim would lock the interaction that broke here.

### Overall
**Merge readiness:** Ready
Correct one-line fix to a real, silently-degrading bug, with a regression test that targets the exact failing shape.

---

---

OPS-PIvers:
Two notes from the review; one fixed, one deliberately left.

**Missing end-to-end coverage — fixed in `3065bf7`.** Valid, and worth more than it looks: every existing test synthesized the transparent-but-colored pixel shape by hand in `buildImage()`, so nothing proved `removeBackgroundFloodFill` actually *produces* that shape. The two halves were verified against an assumption rather than against each other.

The new test runs the real pipeline — builds a fully-opaque image, calls `removeBackgroundFloodFill`, asserts the cleared background is alpha 0 with RGB retained (`[200,200,200]`), then feeds that output straight into `trimImageData` and asserts it trims to the surviving content. `removeBackgroundFloodFill` is now exported for this, same testability trade already made for `trimImageData` in this PR.

Confirmed non-vacuous: reverting the one-line fix (`>>> 24` back to `!== 0`) makes the new test fail alongside the two existing ones. `vitest run utils/imageWorker.test.ts` → **4 tests passing**; eslint ✓ prettier ✓ tsc ✓.

**Endianness caveat — no change, deliberately.** The note is accurate: `Uint32Array` is platform-endian, so `>>> 24` reads alpha only on a little-endian host. Leaving it as-is, because "fixing" it means dropping the 32-bit fast path for byte indexing (`data[offset + 3]`) — a real per-pixel cost on the widget's hot path — to defend against a big-endian browser that no supported target ships. The review's own framing is the right one: this is a pre-existing assumption the fix inherits, not a regression it introduces, and the comment already states the layout it depends on rather than hiding it.

---

OPS-PIvers (commented):
## Automated Code Review — follow-up (delta since `9415aa6`)

**Date:** 2026-08-18

### Summary
One new commit, `3065bf7`, which **closes the exact gap my 2026-08-17 review flagged**: `removeBackgroundFloodFill` — the function that *produces* the transparent-but-colored pixel shape this PR's fix exists to handle — had no coverage, so the two halves of the pipeline were never verified end-to-end. The new test runs flood-fill and feeds its output into `trimImageData`, locking the interaction that actually broke.

Verified locally on `3065bf7`: `vitest run utils/imageWorker.test.ts` → **4 tests passing**.

### Code Quality
✓ No issues found. The end-to-end test is the right shape for this bug: rather than hand-constructing the pixel layout that defeats the old check, it lets the real producer generate it, so the test stays honest if `removeBackgroundFloodFill`'s output format ever changes.

One note for the record, not a request: this commit widens a **second** module-private function to `export` purely for testability (`trimImageData` in the first commit, `removeBackgroundFloodFill` here). Both are the standard trade for testing worker internals and both have a real consumer in the test file, so neither is a dead export — but the file's public surface has now doubled for test reasons alone, and the `self.onmessage` top-level side effect still means these imports depend on `vitest.config.ts`'s `jsdom` environment. Switching this file's test environment to `node` would break both imports.

✓ Not applicable — worker utility.

### Firestore & Security
✓ No issues found — pure in-memory pixel math.

### Regression Risk
Low, unchanged. The delta is one `export` keyword and 41 test lines; no production logic changed. `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, and `functions/src/index.ts` remain untouched.

The platform-endianness caveat from my prior review still stands and is still inherited rather than introduced: `Uint32Array` is platform-endian, so `>>> 24` reads alpha only on a little-endian host. Every browser SpartBoard targets qualifies, and the pre-fix code carried the same assumption.

### Missing Tests
✓ Adequate coverage — the only gap I raised is now closed, and closed at the right level (integration between the two functions rather than another isolated unit test).

### Overall
**Merge readiness:** Ready
The follow-up commit addresses the sole outstanding note from the previous review. Still a correct one-line fix to a real, silently-degrading bug, now with both halves of the pipeline covered.

---

---

PR #2485, opened 2026-08-18
fix(deps): bump tar override to ^7.5.19 (CRITICAL DoS + 4 HIGH CVEs)

OPS-PIvers:
## Summary
- Bumps the `pnpm.overrides.tar` entry in `package.json` from `^7.5.4` to `^7.5.19`, resolving to `tar@7.5.22`
- Clears a CRITICAL node-tar unlimited-input DoS advisory (affected <=7.5.18, patched >=7.5.19) plus four HIGH path-traversal/hardlink-escape advisories, all reached transitively via `firebase-tools > superstatic > re2 > node-gyp > tar`
- `pnpm audit` root vulnerability count drops from 109 to 99; `pnpm why tar` now reports a single resolved version (`7.5.22`) across both resolution paths
- Dev-only devDependency chain (firebase-tools CLI tooling) — no production runtime impact

Tracked in the `dependency-audit` scheduled-task journal (`docs/scheduled-tasks/dependency-audit.md`, MEDIUM `firebase-tools` item).

## Test plan
- [x] `pnpm install` — lockfile updates cleanly, `pnpm why tar` shows a single patched version
- [x] `pnpm install --frozen-lockfile` — passes (lockfile internally consistent)
- [x] `prettier --check package.json pnpm-lock.yaml` — clean
- [x] `pnpm run type-check` — 0 errors
- [x] No source files changed (dependency-override only), so no lint-relevant diff

---
_Generated by [Claude Code](https://claude.ai/code/session_0193Uvq6Fwk9hRUqxzGZpU8u)_


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-18

### Summary
Raises the `pnpm.overrides.tar` floor from `^7.5.4` to `^7.5.19`, re-resolving the lockfile from `tar@7.5.6` to `tar@7.5.22`. Two files, no source changes.

Verified locally on `ec4be34`: `pnpm install --frozen-lockfile` → **clean** (lockfile internally consistent with the manifest, so CI's `install:ci` will not fail on this).

### Code Quality
✓ No issues found. The `package.json` and `pnpm-lock.yaml` `overrides` blocks stay in sync, and the lockfile now resolves a **single** `tar` version across both transitive paths (`firebase-tools` and `node-gyp`) rather than leaving a second copy behind — I confirmed exactly two `tar@7.5.x` references remain in the lockfile, both `7.5.22`. The `deprecated: Old versions of tar are not supported…` marker that rode along with `7.5.6` is gone from the resolution, which is independent corroboration that the old pin was in the publicly-flagged range.

One framing note on the mechanism, since it affects how much this PR is actually doing: the previous floor `^7.5.4` **already permitted** `7.5.22` — the lockfile was simply pinned at `7.5.6`, so the effective fix is the lockfile refresh. Raising the manifest floor is still the right thing to do (it prevents a future resolution from sliding back below `7.5.19`), but the two changes are doing different jobs and only the lockfile half is load-bearing today.

### Widget Standards
✓ Not applicable — dependency metadata only.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found in the diff itself — no secrets, no rules change, no auth surface.

On the security claim: **the "dev-only, no production runtime impact" scoping checks out.** `tar` appears in the lockfile only under `firebase-tools` and `node-gyp`, both devDependency/CLI chains, so it never reaches the browser bundle. I also verified the corollary the PR body doesn't mention: `functions/pnpm-lock.yaml` contains no `tar` entry at all, so there is no second workspace needing the same override — the `functions/` overrides block (`body-parser`, `lodash`, `protobufjs`, `qs`, `ts-deepmerge`) is correctly left alone.

The specific advisory claims (one CRITICAL unlimited-input DoS affecting `<=7.5.18`, four HIGH path-traversal/hardlink-escape) and the `pnpm audit` 109→99 delta are **not independently verifiable from this diff** — they depend on the advisory database at audit time, and no CVE identifiers are cited. The version movement is consistent with them, and the direction is unambiguously correct, but treat the severity labels as reported rather than confirmed by this review.

### Regression Risk
Low. Patch-level movement within one major, no source files touched, and the affected packages are build/deploy tooling — the realistic failure mode is a `firebase deploy` or `node-gyp` rebuild breaking, not runtime behavior. `functions/src/index.ts`, `DashboardContext.tsx`, `WidgetRegistry.ts`, and `types.ts` are all untouched.

The one thing this review cannot cover: `re2` is a native module in that chain and is on the repo's ignored-build-scripts list, so a `firebase deploy` smoke test on a machine that actually rebuilds it is the honest end-to-end check.

### Missing Tests
✓ Adequate — dependency overrides have no testable behavior, and the existing `install --frozen-lockfile` gate in CI is the correct guard. Worth noting there's no automated check that the two `overrides` blocks stay consistent with each other or that a floor never regresses; the `dependency-audit` scheduled task is currently the only thing catching that, which is fine but is a human/routine dependency rather than a CI one.

### Overall
**Merge readiness:** Ready
Correct, minimal, well-scoped dependency bump with a verified-consistent lockfile and accurate dev-only blast-radius scoping.

---

---

PR #2486, opened 2026-08-18
fix(Stations): key assignments by roster student id, not display name

OPS-PIvers:
## Root cause

`components/widgets/Stations/Widget.tsx` keyed `config.assignments` (`Record<string, string|null>`) by the student's derived display name (`"${firstName} ${lastName}"`) instead of the roster's stable student `id`. Two roster students sharing a display name (e.g. two "Emma Smith"s) collided on the same map key — and on the same dnd-kit drag identity, since `DraggableStudent id={name}` used the same string. Assigning one to a station silently overwrote/merged the other's assignment, and dragging either chip moved both.

Same root cause as the `LunchCount` fix in #2247, and previously flagged as present here too — but assessed 3x in the nightly Run Log as "too architecturally large" because of `Stations/nexus.ts` and Randomizer's nameless `RandomGroup.names: string[]`. That assessment was based on a false premise: `nexus.ts` and `stationsActions.ts` both already treat assignment keys as opaque identity strings, so neither needed to change — the bug lived entirely in `Widget.tsx`'s own name-keyed construction.

## Fix

- `activeRoster` now returns `{id, name}[]` — roster `id` for class mode, name-as-id for custom-list mode (pre-existing limitation, same as LunchCount).
- Reads fall back to `assignments[id] ?? assignments[name]`, so pre-fix dashboards and Randomizer's "Send Groups → Stations" button (`nexus.ts`, deliberately untouched — still name-keyed) keep resolving correctly.
- Every write funnels through `persistAssignments`, which now runs `coalesceLegacyKeys` to migrate legacy name-keyed entries to id-keyed form on every persist — needed because `rotateAssignments`/`resetStation` carry forward existing entries generically by key, so a lingering legacy entry could otherwise rotate as an independent phantom "student."
- All drag/reset/shuffle/unassign handlers switched from name-based to id-based identity.

**Band-aid rejected:** disambiguating names at render time only — that wouldn't fix the actual data corruption in `config.assignments` (still a shared key, still overwritten on write). The identity model is the root cause, not the display.

## Regression test

`components/widgets/Stations/Widget.test.tsx` (new), 2 tests:
1. Two same-named roster students with distinct id-keyed assignments land in independent stations.
2. A legacy name-keyed assignment (from Randomizer or pre-fix data) still resolves.

**Fail-before / pass-after** (independently re-verified by the orchestrator): both tests fail against dev-paul (unfixed) — 2/2 fail — and pass with the fix applied — 4/4 (full `Stations/` suite: 29/29).

## Evidence

- `pnpm run validate` — clean (type-check, lint, format, 621 root test files / 7431 tests, 41 functions files / 804 tests).
- Independently re-verified by the orchestrator, including confirming `nexus.ts`/`stationsActions.ts` treat assignment keys as opaque strings and that overflow/stuck-student toasts only show counts (no id/name leak).

## Risk

**Contained.** Single-widget change, backward-compatible read fallback for legacy data, no schema migration required.

---

### DIFF
--- a/components/widgets/Stations/Widget.test.tsx
+++ b/components/widgets/Stations/Widget.test.tsx
@@ -148,4 +148,81 @@ describe('StationsWidget', () => {
     ) as [string, { config: StationsConfig }];
     expect(updatePayload.config.assignments).toHaveProperty('Alex Kim', null);
   });
+
+  it('coalesces a legacy name-keyed entry when Rotate touches it (not just drag/click)', async () => {
+    const widget = createWidget({
+      assignments: { 'John Doe': 'st-a' },
+    });
+
+    render(<StationsWidget widget={widget} />);
+    await screen.findByText('John Doe');
+
+    fireEvent.click(screen.getByTitle('Rotate clockwise'));
+
+    const [, updatePayload] = mockDashboardContext.updateWidget.mock.calls.at(
+      -1
+    ) as [string, { config: StationsConfig }];
+    // Rotated from st-a to st-b, and re-keyed to the roster id in the same write.
+    expect(updatePayload.config.assignments).not.toHaveProperty('John Doe');
+    expect(updatePayload.config.assignments.s1).toBe('st-b');
+  });
+
+  it('coalesces a legacy name-keyed entry when Reset Station touches it', async () => {
+    const widget = createWidget({
+      assignments: { 'John Doe': 'st-a' },
+    });
+
+    render(<StationsWidget widget={widget} />);
+    await screen.findByText('John Doe');
+
+    fireEvent.click(screen.getByTitle('Reset students in Reading'));
+
+    const [, updatePayload] = mockDashboardContext.updateWidget.mock.calls.at(
+      -1
+    ) as [string, { config: StationsConfig }];
+    expect(updatePayload.config.assignments).not.toHaveProperty('John Doe');
+    expect(updatePayload.config.assignments.s1).toBeNull();
+  });
+
+  it('pins the duplicate-name migration outcome: one shared legacy key resolves to exactly one roster student', () => {
+    (useDashboard as unknown as ReturnType<typeof vi.fn>).mockReturnValue({
+      ...mockDashboardContext,
+      rosters: [
+        {
+          id: 'roster-1',
+          name: 'Class 1A',
+          students: [
+            { id: 's1', firstName: 'Emma', lastName: 'Smith' },
+            { id: 's2', firstName: 'Emma', lastName: 'Smith' },
+          ],
+        },
+      ],
+    });
+
+    const widget = createWidget({
+      assignments: { 'Emma Smith': 'st-a' },
+    });
+
+    render(<StationsWidget widget={widget} />);
+
+    // Before any write, the legacy key resolves for BOTH same-named students.
+    const readingZone = screen.getByTestId('station-zone-st-a');
+    expect(within(readingZone).getAllByText('Emma Smith')).toHaveLength(2);
+
+    // Unassigning the first-rendered chip (roster order: s1 first) triggers
+    // the write that migrates/collapses the shared legacy key.
+    fireEvent.click(within(readingZone).getAllByText('Emma Smith')[0]);
+
+    const [, updatePayload] = mockDashboardContext.updateWidget.mock.calls.at(
+      -1
+    ) as [string, { config: StationsConfig }];
+    expect(updatePayload.config.assignments).not.toHaveProperty('Emma Smith');
+    // s1 (first-iterated) deterministically wins the migration and is
+    // explicitly unassigned by this click; s2 has no entry at all (the
+    // ambiguous legacy value could not be attributed to it) and falls back
+    // to the unassigned bucket rather than silently double-counting toward
+    // station capacity.
+    expect(updatePayload.config.assignments.s1).toBeNull();
+    expect(updatePayload.config.assignments).not.toHaveProperty('s2');
+  });
 });
--- a/components/widgets/Stations/Widget.test.tsx
+++ b/components/widgets/Stations/Widget.test.tsx
@@ -1,0 +1,116 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, within } from '@testing-library/react';
+import { StationsWidget } from './Widget';
+import { useDashboard } from '@/context/useDashboard';
+import { WidgetData, StationsConfig } from '@/types';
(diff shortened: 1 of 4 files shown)


OPS-PIvers:
Good catch — pushed `6cd43bc`: both lookups now use an explicit `in` check instead of `??`, so a deliberately-persisted `null` (unassigned) is never masked by a stale legacy name-keyed value. `Stations/` suite still 29/29, type-check clean.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-18

### Summary
Re-keys `StationsConfig.assignments` from display name to roster student `id`, with a legacy name-key read fallback and a `coalesceLegacyKeys` migration applied inside `persistAssignments`. This is the same collision class fixed for `LunchCount` in #2247, and doing the migration at the persist layer (rather than per call site) is what makes it cover `handleRotate`/`handleResetStation`, which carry keys forward generically.

Verified locally on `f4aaf45`: `vitest run components/widgets/Stations/` → **4 files / 31 tests passing**; `tsc --noEmit` → **clean**.

### Code Quality
✓ No issues found.
- The `student.id in assignments ? … : assignments[student.name]` shape (rather than a truthiness check) is the load-bearing detail, and commit `6cd43bc` gets it right: an explicit `null` must not fall through to the legacy name lookup, or unassigning a legacy-keyed student would silently re-read their old station. There is a dedicated test for exactly that (`can unassign a student whose assignment only exists under the legacy name key`).
- `coalesceLegacyKeys` never drops a legacy value — it copies to the id key first (unless an id-keyed value already wins), then deletes the name key either way, so a phantom legacy entry can't survive to be rotated forward independently.
- The `student.id === student.name` guard correctly makes the migration a no-op in custom-list mode, where the typed name genuinely is the only identity.
- `data-testid` on `DroppableZone` forwards — `DroppableZone` spreads `...props` onto its div — so the new zone assertions aren't testing a dead attribute.

### Widget Standards
✓ No issues found. Front-face sizing is untouched apart from chip identity; the `DragOverlay` label still uses `min(14px, 6cqmin)` and now resolves the display name via `activeRoster.find(...)?.name ?? activeId` instead of rendering a raw id. No registry, `WIDGET_APPEARANCE_COMPONENTS`, or `widgetDefaults` surface is involved.

### Admin Config Standards
✓ Not applicable — no `components/admin/` surface touched.

### Firestore & Security
✓ No issues found. `assignments` lives inside the widget config on the user's own dashboard doc; no new collection, no rules change, no auth surface.

### Regression Risk
Low. Does not touch `DashboardContext.tsx`, `WidgetRegistry.ts`, or `functions/src/index.ts`; the `types.ts` change is a doc comment on `StationsConfig.assignments` only, with no `WidgetType`/`WidgetConfig` union change. Three notes:

1. **One-time migration artifact in exactly the case this PR fixes.** For two students sharing a display name today, the pre-fix data has a single `assignments["Emma Smith"]`. Before the first write, the fallback resolves it for *both* Emmas, so both render in that station (and the station can visibly exceed `maxStudents`). On the first persist, `coalesceLegacyKeys` moves the value to whichever roster student `activeRoster` iterates first and deletes the name key — so the second Emma silently drops to unassigned. This is unavoidable (the legacy key cannot say which Emma it was) and strictly better than the collision it replaces, but it is a visible one-time shuffle for the affected teachers and it isn't covered by a test.
2. **Absent students are deliberately excluded from coalescing.** `coalesceLegacyKeys` iterates `activeRoster`, which filters absentees out, so their legacy entries are preserved — correct, and it matches the file's stated snap-back-on-return semantics. The consequence is that a map holding *both* a legacy name key and an id key for the same student is theoretically reachable across absence transitions, and `stationCount`/`rotateAssignments` would count that student twice for capacity until the next write coalesces. I could not construct a live sequence that produces it (absent students never render, so they can't be dragged), so this is theoretical rather than a live bug.
3. **`nexus.ts` still writes name-keyed assignments.** "Send Groups → Stations" is handled by the read fallback plus coalescing, and the PR says so explicitly. Worth noting the residual: `RandomGroup.names` is itself the source of the duplicate-name ambiguity, so groups arriving from the Randomizer still collapse two same-named students onto one key until roster ids are threaded through that path too. Reasonable follow-up, not in scope here.

`handleResetAll` and `handleShuffle` build a fresh id-keyed map and therefore drop absentees' preserved entries — but both built fresh maps before this PR too, so that is unchanged behavior, not a new regression.

### Missing Tests
Coverage of the new behavior is good and targeted: duplicate names staying independent, legacy name-key resolution, legacy unassign, and custom-roster mode (where `id === name`). Two gaps:

1. **No test exercises coalescing through `handleRotate` or `handleResetStation`** — which is the specific claim that makes this fix stronger than the per-call-site alternative (per the PR description, that's why #2478 was closed in favor of this). The migration is currently only proven through the drag path.
2. **No test pins the duplicate-name migration outcome** from note 1 — i.e. that one of the two same-named students deterministically ends up unassigned after the first write. Pinning it would stop a future reader from "fixing" it into a silent duplicate-assignment.

### Overall
Correct fix at the right layer, with the `in`-operator subtlety handled properly and real regression tests; the open items are a rotate-path coalescing test and awareness that affected boards get a one-time reshuffle on first write.

---

---

OPS-PIvers:
Both gaps addressed in `8524d68`:

1. **Rotate/Reset Station coalescing** — added tests clicking "Rotate clockwise" and "Reset students in Reading" against a legacy name-keyed entry, asserting both migrate it to the roster id in the same write. This is the exact claim that made this fix stronger than #2478.
2. **Duplicate-name migration outcome** — pinned that a single legacy key shared by two same-named students resolves for both pre-write, then deterministically collapses to the first-iterated roster student on the first write (explicitly unassigned), with the second student getting no entry at all rather than a silent double-count.

Verified all 4 new assertions genuinely fail without `coalesceLegacyKeys` (temporarily reverted it and reran — 4/4 fail as expected), then restored. `Stations/` suite now 7/7 in the widget test file.

---

OPS-PIvers:
Both good catches, neither blocking:

**Capacity-count edge case**: agreed, narrow and self-healing (legacy data + duplicate names + a `maxStudents` cap, all simultaneously, resolved by the very next write through that key). Not fixing — the fix would add real complexity (teaching `stationCount` to be migration-aware) for a window that closes on first interaction.

**PII scrub gap**: confirmed real and pre-existing (also affects `LunchCountConfig.assignments`), but genuinely out of scope for this PR — it's a `utils/dashboardPII.ts` concern unrelated to the name-collision bug here, and fixing it properly should cover both widgets together rather than a one-off inside this diff. Logging it as a backlog item in `docs/routines/debugger.md` for a future run.

---

PR #2487, opened 2026-08-18
fix(settings-panel): reposition on board pan

OPS-PIvers:
## Root cause

`SettingsPanel` anchors itself to `widgetRef.current.getBoundingClientRect()` and recomputes position on `[updatePosition, viewport, zoom]` changes only. Panning the board canvas (dragging on empty space) moves widgets on screen purely via a CSS `transform: translate()` on an ancestor container — it never touches `widget.x`/`widget.y` (world coordinates), so `SettingsPanel`'s position effect never fired.

`DashboardView` already dispatches a `board-pan` window event on every pan frame specifically so widget-anchored floating popovers can re-measure — `DraggableWindow`'s tool menu and snap menu both already subscribe to it. `SettingsPanel` never wired up the same listener, so opening a widget's settings and then panning the canvas left the settings panel visually stranded at its pre-pan screen position while the widget slid out from under it.

## Fix

Added a `useEffect` in `SettingsPanel.tsx` that subscribes to `window`'s `board-pan` event and calls the existing `updatePosition()` — mirroring the exact pattern already used by `DraggableWindow`'s sibling popovers. No new state, no architectural change.

**Band-aid rejected:** none needed — this is the minimal, correct wiring into an event bus that already exists for exactly this purpose.

## Regression test

`tests/components/common/SettingsPanel.test.tsx` — new test `'re-measures its position on a board-pan event'`: mocks `getBoundingClientRect`, renders the panel, confirms initial position, mutates the mocked rect (simulating a pan), dispatches `board-pan`, and asserts the panel follows.

**Fail-before / pass-after** (independently re-verified by the orchestrator): `212` vs expected `512` on unfixed dev-paul (1/1 fail); full test file passes (9/9) with the fix applied.

## Evidence

- `pnpm run validate` — clean (type-check, lint, format, full root + functions suites).
- Independently re-verified by the orchestrator via a stash/apply-patch cycle against dev-paul.

## Risk

**Contained.** Single component, additive event listener, matches an established sibling pattern already in production use.

---

### DIFF
--- a/components/common/SettingsPanel.tsx
+++ b/components/common/SettingsPanel.tsx
@@ -143,6 +143,12 @@ export const SettingsPanel: React.FC<SettingsPanelProps> = ({
     updatePosition();
   }, [updatePosition, viewport, zoom]);
 
+  // Reposition on board pan (mirrors DraggableWindow's tool/snap menus) — pan doesn't touch widget.x/y, so the deps above never fire.
+  useEffect(() => {
+    window.addEventListener('board-pan', updatePosition);
+    return () => window.removeEventListener('board-pan', updatePosition);
+  }, [updatePosition]);
+
   // Animate in on mount (track both rAF handles for safe cleanup)
   useEffect(() => {
     let outerRaf = 0;
--- a/tests/components/common/SettingsPanel.test.tsx
+++ b/tests/components/common/SettingsPanel.test.tsx
@@ -397,17 +397,7 @@ describe('SettingsPanel', () => {
     expect(leftValue).not.toBe(312);
   });
 
-  /**
-   * Regression: panning the board canvas (drag on empty space) translates the
-   * widget surface via a CSS transform WITHOUT changing widget.x/y — those
-   * are world coordinates, pan is purely visual. DashboardView dispatches a
-   * `board-pan` window event on every pan frame so floating popovers anchored
-   * to a widget's on-screen rect (DraggableWindow's tool menu, snap menu) can
-   * re-measure and follow. SettingsPanel positions itself the same way (via
-   * widgetRef.getBoundingClientRect()) but never listened for that event, so
-   * panning while settings are open left the panel visually stranded at its
-   * pre-pan screen position instead of following the widget.
-   */
+  // Regression: panning (CSS transform, doesn't touch widget.x/y) must re-measure the panel's position, mirroring DraggableWindow's board-pan listener.
   it('re-measures its position on a board-pan event', () => {
     // rect "moves" like a pan would, while widget.x/y (world coords) stay put.
     let rect = {
--- a/tests/components/common/SettingsPanel.test.tsx
+++ b/tests/components/common/SettingsPanel.test.tsx
@@ -398,6 +398,58 @@ describe('SettingsPanel', () => {
   });
 
   /**
+   * Regression: panning the board canvas (drag on empty space) translates the
+   * widget surface via a CSS transform WITHOUT changing widget.x/y — those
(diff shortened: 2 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-18

### Summary
`SettingsPanel` anchors itself to `widgetRef.getBoundingClientRect()` but only recomputed on `[updatePosition, viewport, zoom]`. Panning moves widgets via a CSS transform on an ancestor without touching `widget.x`/`widget.y`, so none of those deps fire and the panel stayed stranded at its pre-pan screen position while the widget slid out from under it. Fixed by subscribing to the `board-pan` window event, the same channel `DashboardView` already dispatches for exactly this purpose.

Verified locally on `d777e11`: `vitest run tests/components/common/SettingsPanel.test.tsx` → **9 tests passing**.

### Code Quality
✓ No issues found.
- The diagnosis is verifiable end-to-end in the tree: `DashboardView.tsx:317` dispatches `board-pan` on every `panOffset` change, explicitly because `panOffset` is kept *out* of context to avoid re-rendering every consumer per pan frame. `DraggableWindow.tsx:1878` and `:1887` already consume it for the snap menu and tool menu. `SettingsPanel` was the one anchored surface that never subscribed — so this reuses the established channel rather than inventing a second one.
- `useEffect` is the correct hook here under the CLAUDE.md rule: subscribing to a DOM event is synchronizing with an external system, not derived state.
- Listener churn is a non-issue, which is worth stating since the dep is a callback: `updatePosition` is a `useCallback` over `[widget.x, widget.y, widget.w, widget.maximized, widgetRef]`, and a pan changes none of those — so the subscription is stable for the entire pan gesture and only re-registers when the widget itself moves or resizes.
- Cleanup is symmetric (same function reference passed to `addEventListener`/`removeEventListener`, no options mismatch), so the listener actually detaches.
- Comment is a single line, per the repo convention.

### Widget Standards
✓ Not applicable — `components/common/SettingsPanel.tsx` is shared chrome, not widget front-face content, so container-query sizing rules don't apply. No registry, `types.ts`, or `widgetDefaults` surface touched.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found — pure DOM positioning; no data access, no persisted state, no auth surface.

### Regression Risk
Low, and the surface is about as small as a fix gets (+6 lines, one effect). Does not touch `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, or `functions/src/index.ts`. Zoom was already covered by the existing layout effect's `zoom` dep, so this closes the remaining transform input rather than duplicating an existing one.

One cosmetic note, not worth changing: the sibling positioning effect deliberately uses `useLayoutEffect` so the panel never flashes at the wrong spot before paint, whereas pan repositioning now runs through a passive listener. During a pan the panel can therefore trail the widget by a frame. That matches how `DraggableWindow`'s two menus already behave, and pan dispatches per frame, so it's imperceptible — flagging only so the asymmetry between the two effects doesn't read as an oversight later.

### Missing Tests
✓ Adequate coverage. The new test is well-constructed for this bug class: it holds `widget.x`/`widget.y` fixed while moving only the mocked `getBoundingClientRect`, which is precisely the shape that defeats the old dep array — so it genuinely fails against the pre-fix code (`expected 212 to be 512`) rather than passing vacuously. Asserting the concrete `left` value on both sides of the event, rather than just "it changed", is the right level of precision.

### Overall
**Merge readiness:** Ready
Minimal fix routed through the existing `board-pan` channel, matching two sibling call sites, with a regression test that isolates the exact failing input.

---

---

PR #2488, opened 2026-08-18
fix(dashboard): unify beta-widget access check between AuthContext and DashboardContext

OPS-PIvers:
## Root cause

`DashboardContext.getDefaultDockTools` (seeds a new user's dock and backs the "Reset to defaults" action) re-derived beta-access eligibility independently of `AuthContext.canAccessWidget`, and the two had drifted:

- **Authoritative gate** (`AuthContext.isBetaUser`): case-insensitive email comparison, plus fallback access via `userRoles.betaTeachers`, `userRoles.superAdmins`, or `roleId === 'super_admin'`.
- **DashboardContext's copy**: `perm.betaUsers.includes(user?.email ?? '')` — case-sensitive, and blind to the role-list fallback sources entirely.

A beta user who could access a widget via `canAccessWidget()` (e.g. an admin-lowercased `betaUsers` entry vs. a mixed-case sign-in email, or account-wide `betaTeachers` access) would never get that widget seeded into their default dock, nor restored by "Reset to defaults" — a silent, confusing inconsistency. PR #2375 partially addressed this by normalizing `betaUsers` on write but explicitly left this read site's gap open, and never covered the role-list sources at all.

## Fix

Extracted the check into a shared `utils/betaAccess.ts` (single source of truth) and pointed both `AuthContext.isBetaUser` and `DashboardContext.getDefaultDockTools` at it.

**Band-aid rejected:** just adding `.toLowerCase()` to the DashboardContext read site — would still miss the `betaTeachers`/`superAdmins`/`super_admin` sources and could re-drift the next time either side changes independently.

## Regression test

`context/DashboardContext.betaDockDefaults.test.tsx` (new), 3 tests: mixed-case email vs. lowercase `betaUsers` entry, access granted only via `userRoles.betaTeachers`, and a negative control (non-beta user correctly excluded).

**Fail-before / pass-after** (independently re-verified by the orchestrator): 2/3 fail against dev-paul (unfixed) — `'clock'` missing from `visibleTools` — 3/3 pass with the fix applied. Also independently type-checked clean.

## Evidence

- `pnpm run validate` — clean (type-check, lint, format, 621 root test files / 7432 tests, 41 functions files / 804 tests).
- Independently re-verified by the orchestrator via a stash/apply-patch cycle against dev-paul, plus a standalone `tsc --noEmit`.

## Risk

**Contained.** New shared utility + two call-site swaps; no behavior change for any already-correctly-gated user.

---

### DIFF
--- a/context/AuthContext.tsx
+++ b/context/AuthContext.tsx
@@ -2283,10 +2283,7 @@ export const AuthProvider: React.FC<{ children: React.ReactNode }> = ({
     };
   }, []);
 
-  // Helper for checking if a user has beta access. Delegates to the shared
-  // `utils/betaAccess.isBetaUser` (single source of truth — see LO2
-  // harmonization note there) so every consumer of beta-gated permissions
-  // agrees on who counts as a beta user.
+  // Helper for checking if a user has beta access — delegates to the shared utils/betaAccess.isBetaUser.
   const isBetaUser = useCallback(
     (betaUsers: string[], email: string | null | undefined) =>
       isBetaUserShared(betaUsers, email, userRoles, roleId),
--- a/context/AuthContext.tsx
+++ b/context/AuthContext.tsx
@@ -78,6 +78,7 @@ import {
   stampLastActive,
 } from '@/utils/lastActiveThrottle';
 import { deriveUserTier, meetsMinTier } from '@/utils/userTier';
+import { isBetaUser as isBetaUserShared } from '@/utils/betaAccess';
 
 // The operator's own organization. Two narrow uses remain after dynamic
 // org resolution shipped:
@@ -2282,24 +2283,13 @@ export const AuthProvider: React.FC<{ children: React.ReactNode }> = ({
     };
   }, []);
 
-  // Helper for checking if a user has beta access
+  // Helper for checking if a user has beta access. Delegates to the shared
+  // `utils/betaAccess.isBetaUser` (single source of truth — see LO2
+  // harmonization note there) so every consumer of beta-gated permissions
+  // agrees on who counts as a beta user.
   const isBetaUser = useCallback(
-    (betaUsers: string[], email: string | null | undefined) => {
-      const lowerEmail = email?.toLowerCase() ?? '';
-      return (
-        betaUsers.some((e) => e.toLowerCase() === lowerEmail) ||
-        (userRoles?.betaTeachers?.some((e) => e.toLowerCase() === lowerEmail) ??
-          false) ||
-        // LO2 harmonization: super admins get beta access from EITHER source —
-        // the legacy admin_settings/user_roles.superAdmins[] list OR a member
-        // doc with roleId 'super_admin'. The legacy list is KEPT as an
-        // additional accepted source (a Paul-gated migration retires it later);
-        // this mirrors OrganizationPanel.resolveActorRole, which reads both.
-        (userRoles?.superAdmins?.some((e) => e.toLowerCase() === lowerEmail) ??
-          false) ||
-        roleId === 'super_admin'
-      );
-    },
+    (betaUsers: string[], email: string | null | undefined) =>
+      isBetaUserShared(betaUsers, email, userRoles, roleId),
     [userRoles, roleId]
   );
 
--- a/context/DashboardContext.betaDockDefaults.test.tsx
+++ b/context/DashboardContext.betaDockDefaults.test.tsx
@@ -1,23 +1,4 @@
-/**
- * Regression coverage for `getDefaultDockTools` / `resetDockToDefaults`'s
- * beta-access check.
- *
- * `DashboardContext` re-derives "is this beta-gated widget accessible to the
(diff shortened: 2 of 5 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-18

### Summary
Extracts the beta-access check into a shared `utils/betaAccess.ts` and routes both `AuthContext.isBetaUser` and `DashboardContext.getDefaultDockTools`'s `isPermAccessible` through it. `DashboardContext` previously did a case-sensitive `perm.betaUsers.includes(user?.email ?? '')` against only the permission's own list, while `AuthContext` — the authoritative gate behind `canAccessWidget` — compared case-insensitively and also honored `userRoles.betaTeachers`, `userRoles.superAdmins`, and `roleId === 'super_admin'`. A legitimately-beta teacher could therefore add a beta widget from the Dock but never have it seeded into a new dock or restored by "Reset to defaults".

Verified locally on `2322b02`: `vitest run context/DashboardContext.betaDockDefaults.test.tsx` → **3 tests passing**; `tsc --noEmit` → **clean**.

### Code Quality
✓ No issues found.
- Extracting to a pure function is the right shape — it takes `userRoles`/`roleId` as parameters rather than reaching for context, so it stays testable and has no hook dependency.
- `useCallback` deps in `AuthContext` (`[userRoles, roleId]`) and the `useMemo` deps in `DashboardContext` (now `+ userRoles, roleId`) are both complete for the new reads. No stale-closure hazard.
- Types line up: `userRoles: UserRolesConfig | null` is already on the `AuthContextValue` surface (`context/AuthContextValue.ts:24`), so the new `DashboardProvider` destructure needs no context change.

One note, since the diff deletes it: the removed `AuthContext` comment carried the **LO2 harmonization rationale** — *why* the legacy `admin_settings/user_roles.superAdmins[]` list is still an accepted source alongside a `roleId: 'super_admin'` member doc, and that a Paul-gated migration retires it later. That reasoning is now nowhere in the codebase, and it is precisely the kind of thing a future reader would "simplify" out of `utils/betaAccess.ts` as redundant. A one-line note on the `superAdmins` clause would preserve it within the CLAUDE.md one-line convention.

### Widget Standards
✓ Not applicable — no widget component, registry, or config surface touched.

### Admin Config Standards
✓ Not applicable — no `components/admin/` surface. The permission model itself is unchanged; only the read-side evaluation moved.

### Firestore & Security
✓ No issues found. No new collection, no rules change, no auth bypass — and note the direction of travel is *toward* the stricter authority: `DashboardContext` now defers to the same predicate `AuthContext.canAccessWidget` already used, rather than keeping a looser/parallel copy.

Worth stating plainly for the record: this is a **widening** of who gets beta widgets seeded into their default dock — `betaTeachers`, `superAdmins`, and `roleId === 'super_admin'` users now qualify where they didn't before, and matching is now case-insensitive. That is the intent (the two gates agreeing is the point), and no one gains access to a widget they couldn't already add manually via the Dock, so it grants no new capability — only default placement.

### Regression Risk
Low. `getAdminBuildingConfig` and the config merge pipeline are untouched — this PR only changes `getDefaultDockTools`'s local `isPermAccessible` and adds two `useMemo` deps. `WidgetRegistry.ts`, `types.ts`, and `functions/src/index.ts` are not touched.

One pre-existing edge carried over unchanged: for a signed-in user with no email, `lowerEmail` is `''`, so a `betaUsers` array containing an empty string would match. The old `.includes('')` had the identical behavior, so this is inherited, not introduced.

### Missing Tests
Coverage is proportionate and hits the two paths that actually regressed — mixed-case email against a lowercase `betaUsers` entry, and a grant arriving via `userRoles.betaTeachers` — plus a negative case proving a non-beta user is still excluded. That negative test is the important one; without it the fix could pass by simply returning `true`.

Gaps, both minor: no case covers the `superAdmins` list or `roleId === 'super_admin'` branches, and `utils/betaAccess.ts` has no direct unit test of its own (it's exercised only through `DashboardContext`). Given the file is now the single source of truth for two consumers, a small dedicated test file covering all four grant sources would be the durable guard against re-drift — which is the exact failure mode this PR exists to close.

### Overall
**Merge readiness:** Ready
Real fix at the right layer — a shared predicate rather than patching the case-sensitivity symptom, which is what would have re-drifted. Worth carrying the LO2 rationale into the new file and adding a direct test for the two uncovered grant sources.

---

---

OPS-PIvers:
Both addressed in `d6573b6`:

1. Restored the LO2 harmonization rationale as a one-line comment on the `superAdmins` clause in `utils/betaAccess.ts` (kept it from being "simplified away" as you flagged).
2. Added `utils/betaAccess.test.ts` with direct unit coverage of all four grant sources (`betaUsers`, `betaTeachers`, `superAdmins`, `roleId === 'super_admin'`) plus the negative case and a null/undefined-email guard — since it's now the single source of truth for two consumers.

`vitest run utils/betaAccess.test.ts context/DashboardContext.betaDockDefaults.test.tsx` → 9/9 passing; lint/format/`tsc --noEmit` all clean.

---

PR #2489, opened 2026-08-18
fix(admin): stop Organization panel popover Escape from leaking to DashboardView

OPS-PIvers:
## Root cause

`RowMenu`, `CellPopover`, and `LocalModal` in `components/admin/Organization/components/primitives.tsx` (used throughout Admin Settings → Organization: `AllOrganizationsView`, `DomainsView`, `BuildingsView`, `TestClassesView`, `RolesView`, `UsersView`) each registered a `document`-level `keydown` Escape handler that closed the popover/menu/modal but never called `stopPropagation()`.

All three are portalled outside any `.widget` ancestor, and `AdminSettings` is a full-screen overlay opened while live dashboard widgets sit behind it (not a shared `Modal`, so `useHasOpenModal()` doesn't shield it). The unstopped keydown bubbled to `DashboardView`'s global Escape handler, which falls back to minimizing the topmost dashboard widget — e.g. silently closing an unrelated running timer just because an admin dismissed a row menu.

Same recurring bug class already fixed 13+ times elsewhere (`ToolDockItem`, `ClassRosterMenu`, `ActiveClassChip`, `FolderItem`, `WidgetLibrary`, etc.), but never ported to this file — flagged as an open, unverified backlog lead on 2026-08-14 and re-confirmed live-reachable before fixing.

## Fix

Applied the established `stopPropagation()` + `isEscapeFromWidgetInput()` guard idiom to all three handlers, matching every prior fix of this class exactly.

**Band-aid rejected:** none — this is the established root-cause fix pattern, not a symptom patch.

## Regression test

`tests/components/admin/Organization/primitives.escapePropagation.test.tsx` (new), 3 tests — one per component — rendering, opening, firing Escape on `document`, and asserting a `window`-level keydown spy was **not** called (plus that the component actually closed).

**Fail-before / pass-after** (independently re-verified by the orchestrator): 3/3 fail against dev-paul (unfixed) — spy called once each; 3/3 pass with the fix applied.

## Evidence

- `pnpm run validate` — clean (type-check, lint, format, 621 root test files / 7432 tests, 41 functions files / 804 tests).
- Independently re-verified by the orchestrator via a stash/apply-patch cycle against dev-paul.

## Risk

**Contained.** Three handlers in one file, additive guard matching an established idiom used 13+ times elsewhere in the codebase.

---

### DIFF
--- a/components/admin/Organization/components/primitives.tsx
+++ b/components/admin/Organization/components/primitives.tsx
@@ -7,6 +7,7 @@ import {
   Info,
   AlertTriangle,
 } from 'lucide-react';
+import { isEscapeFromWidgetInput } from '@/utils/domHelpers';
 
 // Color palette for badges and role accents.
 export type AccentColor =
@@ -420,7 +421,11 @@ export const RowMenu: React.FC<{ items: MenuItem[]; label?: string }> = ({
       setOpen(false);
     };
     const onEsc = (e: KeyboardEvent) => {
-      if (e.key === 'Escape') setOpen(false);
+      if (e.key !== 'Escape') return;
+      if (isEscapeFromWidgetInput(e)) return;
+      // Portalled outside any `.widget` ancestor — stop Escape reaching DashboardView's global handler (see ToolDockItem).
+      e.stopPropagation();
+      setOpen(false);
     };
     document.addEventListener('mousedown', onDocClick);
     document.addEventListener('keydown', onEsc);
@@ -595,7 +600,11 @@ export const CellPopover: React.FC<{
       onClose();
     };
     const onEsc = (e: KeyboardEvent) => {
-      if (e.key === 'Escape') onClose();
+      if (e.key !== 'Escape') return;
+      if (isEscapeFromWidgetInput(e)) return;
+      // Portalled outside any `.widget` ancestor — stop Escape reaching DashboardView's global handler (see ToolDockItem).
+      e.stopPropagation();
+      onClose();
     };
     document.addEventListener('mousedown', onDocClick);
     document.addEventListener('keydown', onEsc);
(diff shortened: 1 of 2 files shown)


OPS-PIvers:
Closing this one in favor of #2481 — both fix the same underlying Escape-leak in `components/admin/Organization/components/primitives.tsx`, independently found by two nightly runs dispatched from the same unmerged `dev-paul` base commit, but #2481's diagnosis goes deeper and its fix is genuinely stronger.

I empirically cross-tested both: `AdminSettings.tsx` registers its own `document`-level Escape listener (closing the whole Admin Settings panel) *before* any nested `RowMenu`/`CellPopover`/`LocalModal` ever mounts. Same-target bubble-phase listeners all fire in registration order regardless of a later listener calling `stopPropagation()` — only `stopImmediatePropagation()` (or, better, capture-phase interception, which runs before bubble phase entirely) stops an already-scheduled sibling listener on the same node. This PR's `stopPropagation()` fix — the idiom matching 13+ prior fixes of the "leaks to a `window`-level handler" bug class — correctly stops the leak to `DashboardView`'s global handler, but does **not** stop `AdminSettings.tsx`'s own handler from firing and closing the entire panel. I verified this directly: with this PR's fix applied, pressing Escape while a `RowMenu` is open still closes all of Admin Settings.

#2481's capture-phase `useCaptureEscape` (mirroring `Modal.tsx`'s reverted-but-since-reintroduced-here `captureEscape` pattern) fixes both cases — verified it also stops the leak to a `window`-level handler, not just `AdminSettings`'s own. It's been through 3 rounds of independent review addressing the capture-phase nesting-order caveat and a comment-length nit, and is ready.

My PR's underlying diagnosis (leaking to a global handler) wasn't wrong, just incomplete — it missed the more direct and more severe symptom. Thanks to the other run for catching the deeper root cause.

---

PR #2490, opened 2026-08-18
fix(functions): bound gcPlcOrphans version-overflow sweep per group

OPS-PIvers:
## Root cause

`gcPlcOrphans.ts`'s `sweepVersionOverflow()` read an entire synced group's `versions` subcollection into memory with an unbounded `.get()` and sorted it client-side — the only sweep in this file with no per-run ceiling (every sibling sweep already has one: `MAX_PLCS_PER_RUN`, `MAX_GROUPS_PER_RUN`, `fetchCategoryPaginated`'s `MAX_CATEGORY_SCAN_PER_PLC`).

In steady state this rarely fires (the client prunes to `VERSION_HISTORY_LIMIT` after every publish), but a group whose client-side prune has been silently failing accumulates an unbounded backlog. This sweep runs **before** the per-PLC loop with no `try/catch` anywhere in the file — a memory/time blowup on one pathological group's unbounded read+sort could abort the entire nightly GC run before a single PLC's activity/presence/tombstone sweep even starts, silently defeating GC network-wide.

## Fix

Replaced the in-memory read-everything-then-sort with a Firestore-side query: `orderBy('version', 'desc').offset(VERSION_HISTORY_LIMIT).limit(MAX_VERSIONS_SCAN_PER_GROUP)`, bounded by a new `MAX_VERSIONS_SCAN_PER_GROUP = 5000` ceiling (mirrors the file's established `MAX_*_PER_RUN` pattern), with a `console.warn` when the ceiling is hit.

**Band-aid rejected:** mirroring `fetchCategoryPaginated`'s `startAfter`-cursor-on-doc-id pagination (the pattern used by the other three sweeps) would have been actively wrong here — version doc IDs are stringified version numbers, so string-order pagination ("1","10","11",...,"2") doesn't match numeric recency, and a capped doc-id page could delete the wrong (non-oldest) snapshots. Ordering by the real numeric `version` field server-side avoids that trap.

## Regression test

`functions/src/gcPlcOrphans.test.ts` — 3 new tests: deletions capped at `MAX_VERSIONS_SCAN_PER_GROUP` per run (remainder deferred to next run), a warning is logged when the ceiling is hit, and unchanged behavior for a group whose overflow is within the cap. Also extended the in-memory stub Firestore to support `orderBy(field, direction)` on a real field (not just doc-id) and `.offset(n)`.

**Fail-before / pass-after** (independently re-verified by the orchestrator): 2/2 new tests fail against dev-paul (unfixed); 38/38 pass with the fix applied. Also independently type-checked clean.

## Evidence

- `pnpm run validate` — clean (type-check, lint, format, full root + functions suites, test:counts guard).
- `pnpm run build:all` — succeeds (Vite build + SSR prerender + functions `tsc` build).
- Independently re-verified by the orchestrator via a stash/apply-patch cycle against dev-paul, plus a standalone `tsc --noEmit` in `functions/`.

## Risk

**Architectural** (tag per the run's own guardrails: this closes a real "can abort the whole nightly GC run" failure mode, not just a cosmetic gap) but narrowly scoped — one function, one new bounded query, backward-compatible (idempotent, defers overflow to next run).

---

### DIFF
--- a/functions/src/gcPlcOrphans.test.ts
+++ b/functions/src/gcPlcOrphans.test.ts
@@ -63,6 +63,7 @@ import {
   PLC_PAGE_SIZE,
   GROUP_PAGE_SIZE,
   CATEGORY_PAGE_SIZE,
+  MAX_VERSIONS_SCAN_PER_GROUP,
 } from './gcPlcOrphans';
 
 // ===========================================================================
@@ -231,8 +232,9 @@ function makeStubDb(seed: {
   }
   interface CollectionRef {
     limit: (n: number) => CollectionRef;
-    orderBy: (field: unknown) => CollectionRef;
+    orderBy: (field: unknown, direction?: 'asc' | 'desc') => CollectionRef;
     startAfter: (cursor: DocSnap) => CollectionRef;
+    offset: (n: number) => CollectionRef;
     get: () => Promise<{ docs: DocSnap[]; size: number }>;
   }
   interface DocSnap {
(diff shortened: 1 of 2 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-18

### Summary
Bounds `sweepVersionOverflow()` in `functions/src/gcPlcOrphans.ts` — previously the only sweep in the file with no per-run ceiling, reading the entire `versions` subcollection with an unbounded `.get()` and sorting client-side. Replaced with a server-side `orderBy('version', 'desc').offset(VERSION_HISTORY_LIMIT).limit(MAX_VERSIONS_SCAN_PER_GROUP)` query plus a ceiling warning.

Verified locally on `6ce9ba4`: `vitest run src/gcPlcOrphans.test.ts` in `functions/` → **38 tests passing**.

### Code Quality
✓ No issues found.
- Ordering by the numeric `version` field instead of paginating on `documentId()` is the right call, and the docstring's reasoning is correct: doc ids here are stringified version numbers, so string order (`"1","10","11","2"…`) does not match numeric recency and a doc-id-paginated page would delete the wrong snapshots. This is a better answer than #2482's `startAfter` approach to the same problem.
- No composite index is required — a single-field `orderBy` on a subcollection is served by Firestore's automatic single-field indexes.
- Net cost is lower, not higher: the old path billed a read for every document in the subcollection; the new one bills at most `VERSION_HISTORY_LIMIT` (10) skipped + up to 5000 returned, and typically returns zero.
- `overflowSnap.size >= MAX_VERSIONS_SCAN_PER_GROUP` is equivalent to `===` given the `.limit()`, which is fine — it just can't over-trigger.

### Widget Standards
✓ Not applicable — Cloud Functions only.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. No new collection, no rules change, no auth surface; the sweep runs under the Admin SDK, and `versions` docs are already `allow update, delete: if false` for clients (`firestore.rules:1324`), so server-side pruning is the intended mechanism.

### Regression Risk
Low overall, with **one real behavior narrowing worth a decision before merge**:

**Version docs missing the `version` field are now invisible to the sweep.** Firestore's `orderBy(field)` excludes documents that lack that field entirely — so such a doc is neither deleted nor counted toward the `VERSION_HISTORY_LIMIT` head, and it accumulates forever. The code this PR replaces handled that case *deliberately and explicitly*, with a comment saying so: `Non-numeric ids sort last (NaN → treated as oldest) so malformed snapshots are pruned first`. That intent is silently dropped here, and the new docstring's justification — "every version doc is written with `version is int` per firestore.rules" — is only half the story: the rule (`firestore.rules:1320`) constrains *client creates*, not Admin-SDK writes or any doc created before that rule shipped.

The blast radius is small (a leaked doc, not a wrong deletion — the fix can never delete a *newer* snapshot, which is the failure mode that actually matters), so this is a note rather than a blocker. Either confirm no fieldless `versions` docs exist and drop the "prune malformed first" intent on purpose, or add a cheap follow-up pass that reaps them.

Nothing else is at risk: `functions/src/index.ts` is untouched, so no Cloud Function signature changed, and the sweep's callers and return contract are unchanged.

### Missing Tests
The three new cases are well-targeted — cap enforcement with a measured remainder, the ceiling `console.warn`, and an under-cap group left alone — and the last one is the valuable guard against an off-by-one that would start deleting into the kept head.

One gap, directly on the finding above: **no test covers a `versions` doc with no `version` field.** This is a notable miss because the test double was specifically taught to model the semantic — `makeCollectionRef` filters `d.data[orderField] !== undefined` with the comment *"mirroring live Firestore — excludes docs where the field is absent"* — so the harness can already express the case, and nothing exercises it. A one-doc test asserting the current (documented) behavior would pin whichever answer you choose above.

### Overall
Correct and well-reasoned bounding of the last unbounded sweep in the file, with a cheaper query than the code it replaces; the open question is whether silently skipping fieldless version docs is an accepted trade or needs a follow-up reaper.

---

---

OPS-PIvers:
Addressed in `5c297fd`. Confirmed the fieldless-doc gap is unreachable through normal operation: the sole writer (`writeVersionSnapshot` in `useSyncedQuizGroups.ts`/`useSyncedVideoActivityGroups.ts`) always constructs a full `PlcVersionSnapshot` (`version: number` is non-optional in `types.ts`) via a full `setDoc`, never a partial write — so `firestore.rules`' `version is int` constraint is corroborating evidence, not the only guarantee. Documented this reasoning directly in the docstring (including the explicit call-out that any *future* writer able to produce a fieldless doc must be fixed rather than this sweep re-widened) and added a test pinning the accepted behavior. 39/39 tests passing, lint/format/`tsc --noEmit` clean.

---

PR #2499, opened 2026-08-19
fix(library): stop Escape leaking from overflow menus to the global widget handler

OPS-PIvers:
## Root cause

Following up on the recurring "portalled-popover Escape leak" bug class (previously fixed 10+ times across different components), two live instances were still missing the guard:

- `components/common/library/LibraryItemCard.tsx` — the card's kebab `OverflowMenu`
- `components/common/library/AssignmentArchiveCard.tsx` — its own `OverflowMenu`

Both menus are `createPortal`'d to `document.body`, outside any `.widget`/`[data-draggable-window]` DOM ancestor, and had no Escape handler at all. Pressing Escape while either menu was open left the menu open *and* let the keydown bubble to `DashboardView`'s global `window`-level Escape handler, which falls back to minimizing the topmost z-index widget. Both components are live: rendered inside `QuizManager`, `VideoActivityManager`, `MiniAppManager`, and `GuidedLearningManager` (confirmed via grep before committing to the fix).

## Fix

Added a `document`-level `keydown` listener (active only while `open`) in each `OverflowMenu`, guarded by `isEscapeFromWidgetInput(e)`, that calls `e.stopPropagation()` before closing — the established idiom already used by `ToolDockItem`/`FolderPickerPopover`/etc.

## Band-aid rejected

Patching `DashboardView`'s global handler to also treat "focus is inside `document.body`-portalled content" as a bail condition — would weaken the global handler for every future unguarded portal instead of fixing the actual missing listener, and would create a second, inconsistent mechanism alongside the established per-component idiom.

## Regression test

New files `tests/components/common/library/LibraryItemCard.escapePropagation.test.tsx` and `AssignmentArchiveCard.escapePropagation.test.tsx`.

## Evidence

Fail-before (pre-fix source via `git show dev-paul:...`, restored via `cp` — never `git stash`): both new tests failed (`expected document not to contain element, found <div role="menu">...` — menu never closed on Escape). Pass-after: 2/2.

Independently re-verified by the orchestrator via the same file-swap method: confirmed fail-before/pass-after matches the subagent's report exactly. Also trimmed a 4-line comment block in both files down to one line during review, per the repo's comment convention (`git show dev-paul:<path>` confirmed neither file has a pre-existing multi-line-comment convention to justify an exception).

Full `pnpm run validate` (628 root test files/7462 tests, 41 functions files/822 tests, all green) + `pnpm run build` — both clean on the orchestrator's independent run.

## Risk

**Contained** — two-file diff plus two new test files; same established idiom used 10+ times elsewhere in this codebase.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JXVWwoBL8m4Wh7PkrrXVMM)_

### DIFF
--- a/components/common/library/AssignmentArchiveCard.tsx
+++ b/components/common/library/AssignmentArchiveCard.tsx
@@ -26,6 +26,7 @@ import React, {
 import { createPortal } from 'react-dom';
 import { MoreHorizontal } from 'lucide-react';
 import { useClickOutside } from '@/hooks/useClickOutside';
+import { isEscapeFromWidgetInput } from '@/utils/domHelpers';
 import type {
   AssignmentArchiveCardProps,
   LibraryBadgeTone,
@@ -94,6 +95,23 @@ const OverflowMenu: React.FC<OverflowMenuProps> = ({ actions }) => {
   // so `useClickOutside` doesn't re-subscribe DOM listeners every render.
   const ignoreRefs = useMemo(() => [menuRef], []);
   useClickOutside(wrapperRef, () => setOpen(false), ignoreRefs);
+
+  // Close on Escape and stop propagation — this menu is portalled to
+  // document.body, outside any `.widget` DraggableWindow ancestor, so an
+  // unhandled Escape here bubbles to DashboardView's global window-level
+  // handler, which falls back to minimizing the topmost widget.
+  useEffect(() => {
+    if (!open) return undefined;
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key !== 'Escape') return;
+      if (isEscapeFromWidgetInput(e)) return;
+      e.stopPropagation();
+      setOpen(false);
+      buttonRef.current?.focus();
+    };
+    document.addEventListener('keydown', handleKeyDown);
+    return () => document.removeEventListener('keydown', handleKeyDown);
+  }, [open]);
 
   // Measure trigger position when the menu opens so the portal renders
   // flush under it. `opacity-70` on archive cards creates a stacking
--- a/components/common/library/AssignmentArchiveCard.tsx
+++ b/components/common/library/AssignmentArchiveCard.tsx
@@ -96,10 +96,7 @@ const OverflowMenu: React.FC<OverflowMenuProps> = ({ actions }) => {
   const ignoreRefs = useMemo(() => [menuRef], []);
   useClickOutside(wrapperRef, () => setOpen(false), ignoreRefs);
 
-  // Close on Escape and stop propagation — this menu is portalled to
-  // document.body, outside any `.widget` DraggableWindow ancestor, so an
-  // unhandled Escape here bubbles to DashboardView's global window-level
-  // handler, which falls back to minimizing the topmost widget.
+  // stopPropagation prevents this portalled menu's Escape from also minimizing the topmost widget.
   useEffect(() => {
     if (!open) return undefined;
     const handleKeyDown = (e: KeyboardEvent) => {
--- a/components/common/library/LibraryItemCard.tsx
+++ b/components/common/library/LibraryItemCard.tsx
@@ -29,6 +29,7 @@ import { useSortable } from '@dnd-kit/sortable';
 import { CSS } from '@dnd-kit/utilities';
 import { Check, GripVertical, MoreHorizontal } from 'lucide-react';
 import { useClickOutside } from '@/hooks/useClickOutside';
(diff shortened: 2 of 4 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-19

### Summary
Adds a document-level Escape handler to the `OverflowMenu` in `LibraryItemCard` and `AssignmentArchiveCard` so closing a portalled overflow menu no longer also minimizes the topmost widget, plus two focused regression test files. Verified locally: `pnpm type-check` exit 0, targeted `eslint --max-warnings 0` clean, `components/common/library` suite 65/65 passing. The fix is correct; the main note is that this is now the third independent implementation of the same pattern.

### Code Quality
The handler itself is sound. I traced the propagation order to confirm it actually works rather than assuming it:

- The global widget handler is `window.addEventListener('keydown', handleKeyDown)` at `components/layout/DashboardView.tsx:1070` — **bubble** phase on `window`.
- The new handlers register on `document`, also bubble. Since `document` precedes `window` in the bubble path, `e.stopPropagation()` genuinely prevents the widget minimize. ✓
- `isEscapeFromWidgetInput(e)` guards the widget-portal/text-input case first, which matches the contract `DraggableWindow.tsx:889-895` explicitly documents for new document-level Escape handlers. ✓
- Focus returns to `buttonRef` on dismiss — good a11y detail that the pre-existing `useClickOutside` path doesn't do.

**Reuse: this is the third copy of this pattern.** The repo already has `captureEscape` in `components/common/Modal.tsx:63-72` and `useCaptureEscape` in `components/admin/Organization/components/primitives.tsx:13-29` — the latter landed on `dev-paul` yesterday as the fix for the structurally identical AdminSettings bug. This PR hand-rolls the logic twice more inline. `useCaptureEscape` is module-private so it wasn't importable as-is, but that argues for promoting it to `hooks/useCaptureEscape.ts` and having all four call sites share it, rather than for a fourth divergent copy. Note the two existing implementations use **capture + `stopImmediatePropagation`** while this one uses **bubble + `stopPropagation`** — both work here, but three subtly different contracts for one concern is how the next Escape bug gets written.

Minor: `useEffect(...)` returning `undefined` explicitly on the inactive path matches the file's existing style. ✓

### Widget Standards
✓ Not applicable — shared library card components, no widget registry, defaults, appearance, or container-query surface touched.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
✓ No issues found. No Firestore, auth, or network surface in this diff.

### Regression Risk
Low. Checked the neighbours that also listen for Escape:

- `DraggableWindow.tsx:417` registers `window.addEventListener('keydown', onKey, true)` — **capture** phase, so `stopPropagation()` at `document` cannot pre-empt it. In practice it's gated on `showMaxMenu`, which can't be open at the same time as a library card's overflow menu, so no live conflict.
- `Modal.tsx:78-82` registers on `window` and only uses capture when `captureEscape` is passed. A repo-wide grep finds **zero** call sites passing that prop, so today every library modal wrapping these cards is on bubble-phase `window` and is correctly shielded by the new handler. Latent constraint worth knowing: if a modal ever opts into `captureEscape`, its `stopImmediatePropagation` fires first and would close the whole modal out from under an open overflow menu. Sharing one hook (above) would make that ordering a single decision instead of a per-call-site accident.
- Consumers checked: `QuizManager`, `MiniAppManager`, `VideoActivityManager`, `GuidedLearningManager`, `LibraryDevHarness` — none pass `captureEscape`.

No `WidgetRegistry.ts`, `DashboardContext.tsx`, `types.ts`, or `functions/` changes.

### Missing Tests
✓ Adequate coverage. Both new test files pin the actual pre-fix failure mode (Escape reaching an outer listener) rather than just asserting the menu closes, which is the right shape for a propagation bug. The one thing not covered is the capture-phase interaction described above — reasonable to leave untested while no call site passes `captureEscape`.

### Overall
The fix is correct and well-tested; the substantive suggestion is to extract `useCaptureEscape` into a shared hook and converge the now-four Escape-containment implementations rather than adding a third contract.

---

---

OPS-PIvers:
Agreed the `useCaptureEscape`/`captureEscape`/inline-bubble-handler split is worth converging — but promoting `useCaptureEscape` to a shared hook and re-pointing all 4 call sites (`Modal.tsx`, `primitives.tsx`, and this PR's two) is a cross-cutting refactor outside this PR's scope (a name-collision/PII-adjacent bug fix, not a hook consolidation). Logged as a new Dashboard & Layout backlog item in `docs/routines/debugger.md` rather than expanding this diff.

---

PR #2503, opened 2026-08-19
fix(subs): split shared-share read rules so the /subs directory query is allowed

OPS-PIvers:
## Problem

`/subs` failed with **"Couldn't load boards: You do not have permission to view this board."** for an org admin. The failure was not specific to that account: the `/subs` teacher directory was denied for **everyone except a Firestore `/admins/{email}` superadmin or the share's own host** — including every actual substitute teacher.

## Root cause

`/shared_boards` and `/shared_collections` gated substitute shares with one `allow read` whose branches read `intendedMode != 'substitute'`, `originalAuthor` / `hostUid`, and `expiresAt`.

That is correct for `get`, but a **`list` rule is evaluated against the query, not the matched documents**. Verified against the Firestore emulator with a rule/query matrix:

| rule reads | query pins it by `==` | query pins it by range | query doesn't mention it |
| --- | --- | --- | --- |
| `intendedMode` | resolves (allow) | — | `.get()` default → false |
| `originalAuthor` | resolves | — | `.get()` default → false |
| `expiresAt` | — | **evaluation error → deny** | `.get()` default → false |

The denial is independent of the result set — the directory query was denied even when it matched zero documents.

Under the directory query (`intendedMode ==`, `buildingId ==`, `expiresAt >`) the old branches evaluated to: `intendedMode != 'substitute'` → **false** (pinned to `'substitute'`), `originalAuthor` → **null**, `expiresAt` → **evaluation error**. The only branch that could ever pass was `isAdmin()`, which short-circuits before touching `resource` — hence "works for one admin, fails for another."

## Fix

Split both match blocks into:

- **`allow get`** — unchanged condition, so single-doc reads stay gated on host / admin / `@orono.k12.mn.us` **and** `expiresAt`.
- **`allow list`** — query-scoped, with each branch reading only what its query pins by equality:
  - a host's own shares (`originalAuthor` / `hostUid` == uid) — `useReconcileExpiredSubShares`
  - the `/subs` directory (`intendedMode == 'substitute'` + `@orono` email)
  - on `shared_boards`, the PLC board picker (`plcId == <plc>`), now **tightened** to require current membership of the queried PLC (it previously passed via the `intendedMode != 'substitute'` default and was open to any authed user).

An unscoped list stays denied for everyone but admins.

**Accepted trade-off:** expiry cannot be enforced in query scope (`expiresAt` is range-filtered, which errors). An `@orono` caller can therefore *list* — never `get` — an expired substitute share until the hourly `expireSubShares` sweep reaps it: ≤1h, or ≤7d for a share still holding unrevoked Drive grants. The client queries and their client-side expiry filters are unchanged, so the directory never renders an expired board.

## Also corrected

`useSubstituteShares.ts`, `SubCollectionsList.tsx` and the hook tests documented the opposite (per-document) model — "Firestore fails the WHOLE getDocs if any one matched doc is denied". That premise is what the #2150 retry loop was built on, working around a denial it could never recover from. Comments now state the verified mechanism; the retry itself is kept (and re-documented) as cover for a transient stale-token denial right after sign-in.

## Testing

- `pnpm run test:rules` — 47 files / **1140 passed** (up from 1112; baseline ratcheted)
- New regression tests: `/subs` directory list succeeds for an Orono sub, succeeds when matching zero docs, is denied for non-Orono and for a spoofed embedded-`@` email; host cleanup list; unscoped list denied; PLC list allowed for a member and denied for a non-member. Mirrored on `shared_collections`.
- `pnpm run validate` — exit 0 (type-check, lint, format-check, root 5994 + functions 822 tests)

https://claude.ai/code/session_01HwcqBfUfjch1LmAc6amhFf

---
_Generated by [Claude Code](https://claude.ai/code/session_01HwcqBfUfjch1LmAc6amhFf)_


OPS-PIvers:
Thanks — the stale-comment finding from the earlier pass was real and is fixed in c38c78f (`usePlcSharedBoards.ts` + `PlcSharedBoardsBody.tsx` now describe the `plcId` filter as the server-enforced boundary it became).

On the deleted-PLC edge case: the mechanism is right, but I ran it against the emulator on this branch and the outcome is unchanged, so I'm leaving the rule as-is. A board tagged with a since-deleted `plcId`:

| caller / query | result |
| --- | --- |
| any user, `where plcId == <deleted>` | denied (`evaluation error`) |
| the board's own host, same query | denied — `originalAuthor` is unpinned by that query, so the host branch is `false` regardless |
| host, `plcId` + `originalAuthor` both pinned | **allowed** — the host branch is `true` and CEL absorbs the erroring disjunct |

So `get(...).data.memberUids` erroring never poisons a branch that would otherwise pass, and swapping it for `.data.get('memberUids', [])` would turn `error` into `false` without changing a single access decision — only the emulator log line. It's also unreachable from the UI: the PLC tab only queries a `plcId` while that PLC is open in the dashboard, so the doc exists by construction.

---

firestore.rules
```diff
@@ -868,6 +869,40 @@ service cloud.firestore {
...
+        || (
+          resource.data.get('intendedMode', null) == 'substitute'
+          && request.auth.token.get('email', '').lower().matches('^[^@]+@orono[.]k12[.]mn[.]us$')
+        )
+        || (
+          resource.data.get('plcId', null) != null
+          && request.auth.uid in get(
+            /databases/$(database)/documents/plcs/$(resource.data.plcId)
+          ).data.memberUids
+        )
```
OPS-PIvers, line 905: **New `plcId` list branch lets PLC members read substitute-share metadata the rest of this file says they shouldn't see.**

This branch grants `list` to *any current member of PLC `plcId`* — no email-domain check, no expiry check, no host/admin check. That's fine for ordinary PLC boards, but it also applies to `shared_boards` docs where `intendedMode == 'substitute'`, and nothing here excludes those.

`plcId` is pinned to `null` at create time for substitute shares (line 918), but the `allow update` rule's substitute-lock (lines 950-960) only pins `buildingId`, `initialState`, `expiresAt`, `subEmails`, `driveGrants` — **not `plcId`**. The separate plcId-transition clause (961-968) allows `plcId` to move from `null` to any PLC the writer currently belongs to. So the host (non-admin) of a substitute share can `updateDoc` it to attach a `plcId` they're a member of, without violating any other constraint.

Once that lands, this branch grants list access — and thus the full doc, including `driveGrants`, `subEmails`, `buildingId`, `initialState` — to every member of that PLC, regardless of `@orono.k12.mn.us` domain or `expiresAt`. That directly contradicts the GET rule's threat model above ("reads are restricted to host, admins, and @orono.k12.mn.us callers... org-internal routing/Drive-grant metadata").

Compare with `shared_collections`: its `allow update` requires `intendedMode == 'copy'` for *any* update, making substitute Collection shares fully immutable post-create — and its `allow list` correspondingly has no `plcId` branch at all. `shared_boards` is the outlier here because this PR added a `plcId` list branch without also closing the matching gap in the update rule.

Suggested fix — either:
- add `&& resource.data.get('intendedMode', null) != 'substitute'` to this branch, or
- pin `plcId` immutable in the substitute-lock branch of `allow update` (matching `buildingId`/`expiresAt`/etc.)

plus a regression test asserting a non-host, non-admin, non-`@orono` PLC member is denied listing a substitute share via this branch even after `plcId` is attached post-create.

---

OPS-PIvers (commented):
Fresh security pass on this diff — found one new issue beyond the three existing comments (inline below), no other new correctness/security/CLAUDE.md issues.

---

firestore.rules
OPS-PIvers: Confirmed and fixed in 91a8c4a — the update rule's substitute-lock pinned `buildingId`/`initialState`/`expiresAt`/`subEmails`/`driveGrants` but not `plcId`, and the transition clause allows `null` → any PLC the host belongs to. Good catch.

**Option 1 does not work** — worth recording, because it fails for exactly the reason this PR exists. The PLC query doesn't pin `intendedMode`, so in query scope `resource.data.get('intendedMode', null)` resolves to the `.get()` default rather than the doc's value, and `null != 'substitute'` is trivially true. I built that rule verbatim on the emulator against a substitute doc carrying `driveGrants`, listed as `member@elsewhere.com`:

```
### option-1 guard => LEAKED: listed 1 substitute doc(s) despite the != 'substitute' check
```

A guard that reads correctly and silently passes would be worse than none.

**Took option 2**, with one deviation: the pin sits *outside* the admin bypass as a standalone conjunct, so `substitute + plcId` is an impossible state for every writer, not just non-admins. The lifecycle pins beside it are deliberately admin-overridable; this one can't be, since it's the invariant the list branch rests on.

```
request.resource.data.get('intendedMode', null) != 'substitute'
  || request.resource.data.get('plcId', null) == null
```

Five regression tests added (`shared_boards — substitute shares can never acquire a plcId`): host can't attach a `plcId`, admin can't either, can't sneak one in alongside a legal content edit, host can still edit substitute content, and non-substitute shares can still be tagged into a PLC. Rules suite 1145 passing, `validate` exit 0.

**Residual limit, stated plainly:** this holds on the write path only. A substitute doc written out-of-band via the Admin SDK with a `plcId` already on it would still be listable by that PLC's members — as above, there is no list-side guard expressible in query scope.

---

PR #2512, opened 2026-08-20
fix(rules): shape-validate announcements/pollVotes writes, admin-only delete

OPS-PIvers:
## Root cause
`firestore.rules`' `announcements/{announcementId}/pollVotes/{optionIndex}` subcollection (the live vote-tally docs `PollWidget` writes via `setDoc(..., {count: increment(1)}, {merge:true})`) allowed `write: if request.auth != null` with **no shape or value validation** — any authenticated user, including an anonymous student joined by PIN, could overwrite a poll's tally doc with an arbitrary shape/value or delete it outright, unlike the much more tightly-validated `poll_sessions/*/votes/*` rule right next to it. Confirmed live-reachable and unfixed via this repo's own standing security-rules tracking doc (open since 2026-06-10) and the nightly-run backlog (open since 2026-08-14).

## Fix
```
allow create, update: if request.auth != null &&
  request.resource.data.keys().hasOnly(['count']) &&
  request.resource.data.count is int &&
  request.resource.data.count >= 0;
allow delete: if isAdmin();
```
Read is unchanged (open to any authed user — anonymous tallies by design). Verified against every live client call site (`PollWidget/Widget.tsx`'s `vote()`, the only writer) that the shape constraint matches exactly what's sent (`{count: increment(1)}` merge) — confirmed empirically against the real Firestore emulator that `increment()` resolves to a plain int for rule evaluation purposes. No client currently deletes a `pollVotes` doc, so admin-only delete has zero live-path impact.

## Band-aid rejected
Denying all client writes and routing votes through a new Cloud Function (an alternative floated in the tracking doc) — rejected as over-scope for a single-bug fix: it would need a new callable, client rewiring, and deploy coordination. The shape/type/value validation applied here matches this repo's established "shape-validated but still client-writable" idiom used elsewhere (e.g. `quiz_attempt_ledger`'s increment-style checks), fixing the actual root cause (missing validation) without an architecture change.

## Scope note
This does not add per-voter dedup — a same-shaped `{count: int}` overwrite by a different authenticated user is still technically possible. Closing that fully would need a `poll_sessions/votes`-style per-voter-doc redesign; correctly out of scope for tonight's fix, which closes the "arbitrary shape/value or unauthorized delete" gap.

## Regression test
`tests/rules/announcementPollVotes.test.ts` (10 tests, real Firestore emulator) — covers valid increment-merge votes, rejecting non-int/negative/extra-field writes, and admin-only delete.

## Before/after evidence (independently re-verified by the orchestrator against the real emulator, not just the subagent)
Fail-before (pre-fix rules):
```
Test Files  1 failed (1)
Tests  4 failed | 6 passed (10)
```
Pass-after (fix restored):
```
Test Files  1 passed (1)
Tests  10 passed (10)
```
Full `test:rules` suite re-run on the fixed branch: **48 files / 1155 tests, all green** — no regressions in the sibling `poll_sessions/votes` suite or elsewhere.

`pnpm run validate` and `pnpm run build:all` both green, independently re-run by the orchestrator after rebase-check against latest `dev-paul` (unchanged, still `c0eeb0a`).

## Also included
A one-line fix for a separate, low-risk backlog lead in the same area: `tests/components/plcMeetingMode.test.tsx`'s longest test (5 chained UI steps) was flaking under real nightly-run CPU contention against Vitest's 5000ms default — bumped to 15000ms for that one test with a one-line comment explaining why.

## Risk
**Contained** — a single subcollection's write rule, verified against its one real client writer and the full rules-test suite; no schema change, no client code changes required.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jd296g9b7x6FpFNLkng2hY)_

### DIFF
--- a/components/widgets/PollWidget/Widget.tsx
+++ b/components/widgets/PollWidget/Widget.tsx
@@ -7,6 +7,7 @@ import {
   increment,
 } from 'firebase/firestore';
 import { db } from '@/config/firebase';
+import { logError } from '@/utils/logError';
 import {
   useGlobalStyle,
   useDashboardActions,
@@ -94,10 +95,15 @@ export const PollWidget: React.FC<{ widget: WidgetData }> = ({ widget }) => {
     if (_announcementId) {
       if (userVoted !== null) return; // one vote per session
       setUserVoted(index);
-      void setDoc(
+      setDoc(
         doc(db, 'announcements', _announcementId, 'pollVotes', String(index)),
         { count: increment(1) },
         { merge: true }
+      ).catch((err) =>
+        logError('PollWidget.vote', err, {
+          announcementId: _announcementId,
+          index,
+        })
       );
       return;
     }
--- a/firestore.rules
+++ b/firestore.rules
@@ -3339,10 +3339,19 @@ service cloud.firestore {
         );
       allow write: if isAdmin();
 
-      // Poll votes subcollection — authenticated users can read live tallies and cast votes
+      // Poll votes subcollection — authenticated users can read live tallies and cast votes.
+      // Shape-locked to {count: non-negative int} (the increment() merge PollWidget
+      // sends) so a vote write can't overwrite the tally with an arbitrary value/shape
+      // — unlike the unrestricted `write` this replaces. Delete is admin-only (mirrors
+      // the poll_sessions/votes sibling's teacher/admin-only reset); no client deletes
+      // a pollVotes doc today.
       match /pollVotes/{optionIndex} {
         allow read: if request.auth != null;
-        allow write: if request.auth != null;
+        allow create, update: if request.auth != null &&
+          request.resource.data.keys().hasOnly(['count']) &&
+          request.resource.data.count is int &&
+          request.resource.data.count >= 0;
+        allow delete: if isAdmin();
       }
     }
 
--- a/scripts/test-count-baseline.json
+++ b/scripts/test-count-baseline.json
@@ -9,7 +9,7 @@
     "tests": 703
   },
   "rules": {
-    "testFiles": 47,
-    "tests": 1145
+    "testFiles": 48,
+    "tests": 1155
   }
 }
--- a/scripts/test-count-baseline.json
+++ b/scripts/test-count-baseline.json
@@ -10,6 +10,6 @@
   },
   "rules": {
     "testFiles": 48,
-    "tests": 1155
+    "tests": 1156
   }
 }
--- a/tests/components/plcMeetingMode.test.tsx
+++ b/tests/components/plcMeetingMode.test.tsx
@@ -381,7 +381,7 @@ describe('PlcMeetingMode — guided happy path', () => {
 
     // The success view renders.
     expect(await screen.findByText(/meeting saved/i)).toBeInTheDocument();
-  });
+  }, 15000); // Longest test in the file (5 chained UI steps) — the 5000ms default flakes under nightly-run CPU contention.
 
   it('a Review "Discuss this" jumps to Decide pre-linked to the assessment', async () => {
     render(<PlcMeetingMode plc={fakePlc} onNavigate={vi.fn()} />);
--- a/tests/rules/announcementPollVotes.test.ts
+++ b/tests/rules/announcementPollVotes.test.ts
@@ -128,6 +128,23 @@ describe('announcement pollVotes — write shape validation', () => {
   it('an unauthenticated caller cannot write a vote', async () => {
     await assertFails(setDoc(voteRef(asUnauthed(), 0), { count: 1 }));
   });
+
+  it('KNOWN EDGE CASE: a legacy doc with an extra field (written before this rule existed) rejects all future increment-merge votes', async () => {
+    // Documents the one gap flagged by review on this fix: `hasOnly(['count'])`
+    // applies to the resulting merged document, not just the write payload, so
+    // a pre-existing doc carrying any field beyond `count` — impossible via the
+    // current PollWidget.vote() call site, but not impossible via a pre-fix
+    // write or manual edit — would have every subsequent vote silently denied.
+    await testEnv.withSecurityRulesDisabled(async (ctx) => {
+      await setDoc(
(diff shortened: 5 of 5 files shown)


OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-20

### Summary
Replaces the unrestricted `write: if request.auth != null` on `announcements/{id}/pollVotes/{optionIndex}` with a shape lock (`hasOnly(['count'])` + non-negative int) on create/update and an admin-only delete, plus a 10-test emulator suite and a `rules` count-baseline bump. The fix is correct and matches the single live writer. One untested edge on pre-existing production documents is worth a look before deploy.

### Code Quality
✓ No issues found. Rule comment is concise and load-bearing; the test file follows the existing `tests/rules/` harness conventions.

### Widget Standards
✓ Not applicable — no widget front-face changes.

### Admin Config Standards
✓ Not applicable.

### Firestore & Security
Independently verified the writer/reader inventory rather than taking the description on faith. `grep -rn pollVotes components/ hooks/ utils/ functions/src` returns exactly three call sites: `PollWidget/Widget.tsx:98` (`setDoc(..., { count: increment(1) }, { merge: true })` — the only writer), `PollWidget/Widget.tsx:55`, and `admin/Announcements/Widget.tsx:554` (collection `onSnapshot`, a `list`, still covered by the unchanged `allow read`). Nothing deletes a tally doc, so `allow delete: if isAdmin()` breaks no live path.

**New finding, non-blocking but worth a pre-deploy check:** `hasOnly(['count'])` and `count is int` apply to `update` over *existing* documents too. Any `pollVotes` doc already in production — every one of which was written under the unrestricted rule this replaces — that carries an extra field, or whose `count` was ever stored as a double, will have **every subsequent vote denied**. The new suite only ever seeds `{ count: <int> }` (`announcementPollVotes.test.ts:92`), so this path has no coverage.

What makes it worth naming rather than dismissing: the denial is silent end to end. `PollWidget.vote()` calls `setUserVoted(index)` *before* the write and then fires `void setDoc(...)` with no `.catch` — a rejected write becomes an unhandled rejection while the voter's UI already shows their vote as cast. In practice `PollWidget` has only ever written `{ count }`, so the probability is low; the cost of confirming is one read over the live collection, or one extra test seeding a legacy-shaped doc.

### Regression Risk
Low. Doesn't touch `DashboardContext.tsx`, `WidgetRegistry.ts`, `types.ts`, or `functions/src/index.ts`. The `test-count-baseline.json` bump raises a floor (`checkCountFloor` allows growth, blocks shrinkage), and the new file lands in the same PR, so the floor is never ahead of reality.

**Verified locally** (Node 22; CI on Node 24 is authoritative): `pnpm run test:rules` → **48 files / 1155 tests passed**, matching the bumped baseline exactly, and `✓ Test count guard passed`. CI is green on all 7 checks for `26336a4`.

### Missing Tests
- No test seeds a pre-existing `pollVotes` doc with an extra field or a double `count` and then attempts the real `increment()` merge — the one gap described above.
- Coverage is otherwise complete for the new rule: increment-merge control, fresh-option create, non-int, negative, extra-field, unauthenticated, read, and both delete branches.

### Scope note
The `plcMeetingMode.test.tsx` timeout bump is unrelated to the rules fix. It's one well-commented line and a reasonable flake mitigation, but the PR title doesn't cover it.

### Overall
Correct, tightly scoped, and empirically verified against the real emulator; the only open item is a pre-existing-data question that a single live-collection read would settle.

---

---

OPS-PIvers:
Good catch on the silent-failure path. I don't have access to query the live production `pollVotes` collection from here to confirm no legacy-shaped docs exist, so I didn't try to guess at that — instead:

1. Added `.catch(err => logError('PollWidget.vote', ...))` to the write in `PollWidget.vote()` (b1f77f3), so any rejected write — legacy-shaped doc, offline, or otherwise — surfaces in logs instead of an unhandled promise rejection while the UI silently shows the vote as cast. Didn't revert the optimistic `setUserVoted` UI state on failure, since that's a larger UX behavior change beyond this fix's scope (the vote-optimistically-then-reconcile pattern isn't used elsewhere in this widget either).
2. Added the suggested test seeding a legacy-shaped doc (`{count, teacherUid}`) and confirming the documented deny-on-update behavior, so this edge case has explicit coverage rather than just being described in a comment.

Full `test:rules` (48 files / 1156 tests) and `validate`/`build:all` all green on `b1f77f3`. A maintainer spot-check of the live collection for any pre-existing non-`{count}` docs before deploy is still the one thing I can't do from here, per your original note.

---

OPS-PIvers (commented):
## Automated Code Review

**Date:** 2026-08-21

### Summary
Unchanged since the 2026-08-20 review (head still `b1f77f3`). The `pollVotes` shape lock, the `PollWidget.vote()` error handling, and the scoped `plcMeetingMode` timeout bump all still read correctly.

New this run: **this branch is now a strict ancestor of #2521's** (`nightly/build-tooling-2026-08-21`), verified with `git merge-base --is-ancestor`.

### Code Quality / Widget Standards / Admin Config Standards
✓ No issues found — see the 2026-08-20 review; nothing has changed on this head since.

### Firestore & Security
One non-blocking note I raised in more detail on #2521, repeated here because it belongs to this diff: `count is int && count >= 0` still admits *any* non-negative integer, so an authenticated client can write `{count: 0}` or `{count: 999999}` and pass. What the rule genuinely prevents is shape corruption — extra keys, wrong types, negative counts — which is a real improvement over the fully-open `allow write` it replaces, but it is narrower than "can't overwrite the tally with an arbitrary value." Tightening `create` to `count == 1` and `update` to `count == resource.data.count + 1` matches exactly what `increment(1)` produces and would reduce the worst case to +1 per write. Scope call, not a defect.

### Regression Risk
✓ Low. Merges cleanly into `dev-paul`. Verified at #2521's head: `vitest run tests/components/plcMeetingMode.test.tsx components/widgets/PollWidget` passes.

### Missing Tests
✓ Adequate coverage — the rules suite exercises the new deny paths, not just the allow path.

### Overall
**Merge readiness:** Ready — but merge #2521 instead

Every commit here is contained in #2521, so merging that one closes this automatically and brings the `recount-org-members` trim fix along with it.

---

---

